PR #5, opened 2020-02-05
Fixed the problem that apiserver fails to sync pods by adding partition configs

jshaofuturewei:
1. Updated scripts by adding 
/registry/pods/,,tenant2
It enables to watch all the changes of default tenants

2. Loading config only when creating storage to reduce I/O

3. Add more options for config
key, value1, value2
key, , value2
key, value1,

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go
+++ b/staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go
@@ -21,7 +21,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io/ioutil"
 	"os"
 	"strconv"
 	"strings"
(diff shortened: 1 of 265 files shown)


staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go
```diff
@@ -244,17 +213,25 @@ func (wc *watchChan) startWatching(watchClosedCh chan struct{}) {
...
-	m := parseConfig(wc.partitionConfigFilename)
 
-	if val, ok := m[wc.key]; ok {
-		rangeStart := wc.key + val[0]
-		rangeEnd := wc.key + val[1]
-
-		wc.key = rangeStart
-		opts = append(opts, clientv3.WithRange(rangeEnd))
-		klog.V(3).Infof("Assigned watcher to partition wc.key=%v, endKey=%v", wc.key, rangeEnd)
+	if val, ok := wc.partitionConfig[wc.key]; ok {
```
Sindica: Using array for range makes the logic harder to read & maintain. Please use struct instead. When parsing the configuration file, assign range to different field such as start, end. You may use flags for open end ranges.

PR #9, opened 2020-02-07
Support multiple api server configurations

Sindica:


cmd/kubelet/app/server.go
```diff
@@ -844,35 +849,44 @@ func buildKubeletClientConfig(s *options.KubeletServer, nodeName types.NodeName)
...
-func updateDialer(clientConfig *restclient.Config) (func(), error) {
-	if clientConfig.Transport != nil || clientConfig.Dial != nil {
-		return nil, fmt.Errorf("there is already a transport or dialer configured")
+func updateDialer(clientConfigs *restclient.Config) ([]func(), error) {
+	returnFuncs := make([]func(), len(clientConfigs.GetAllConfigs()))
+	for i, clientConfig := range clientConfigs.GetAllConfigs() {
+		if clientConfig.Transport != nil || clientConfig.Dial != nil {
+			return nil, fmt.Errorf("there is already a transport or dialer configured")
+		}
+		d := connrotation.NewDialer((&net.Dialer{Timeout: 30 * time.Second, KeepAlive: 30 * time.Second}).DialContext)
```
jshaofuturewei: If we have more than 10 clients,  is that possible to wait 300 seconds to get timeout?

jshaofuturewei: I mean we can make dialers asynchronously.

cmd/kube-scheduler/app/options/options.go
```diff
@@ -302,36 +303,40 @@ func createClients(config componentbaseconfig.ClientConnectionConfiguration, mas
...
 
-	kubeConfig.AcceptContentTypes = config.AcceptContentTypes
-	kubeConfig.ContentType = config.ContentType
-	kubeConfig.QPS = config.QPS
-	//TODO make config struct use int instead of int32?
-	kubeConfig.Burst = int(config.Burst)
+	for _, kubeConfig := range kubeConfigs.GetAllConfigs() {
+		kubeConfig.AcceptContentTypes = config.AcceptContentTypes
+		kubeConfig.ContentType = config.ContentType
+		kubeConfig.QPS = config.QPS
```
jshaofuturewei: The change is fine. I spent a little more on the QPS and Burst settings.
The default value is 50/100 across different components hard coded. 
I am not sure if the stability tests driven by sonya are affected. It might be tuned based on CPU/memory/network.

cmd/kubelet/app/server.go
```diff
@@ -986,16 +1003,19 @@ func InitializeTLS(kf *options.KubeletFlags, kc *kubeletconfiginternal.KubeletCo
...
-	case runtime.ContentTypeProtobuf:
-		cfg.AcceptContentTypes = strings.Join([]string{runtime.ContentTypeProtobuf, runtime.ContentTypeJSON}, ",")
-	default:
-		// otherwise let the rest client perform defaulting
+
+	for _, cfg := range cfgs.GetAllConfigs() {
+		cfg.ContentType = contentType
+		switch contentType {
+		case runtime.ContentTypeProtobuf:
+			cfg.AcceptContentTypes = strings.Join([]string{runtime.ContentTypeProtobuf, runtime.ContentTypeJSON}, ",")
```
jshaofuturewei: The standard protobuf content type is "application/json".  There are limited type of accept content types for different components, such as "application/vnd.google.protobuf" for Prometheus. In that case, I am worried that there are duplicate entries in runtime.ContentTypeProtobuf, such as "application/json,application/json"

cmd/kubemark/hollow-node.go
```diff
@@ -93,18 +93,20 @@ func (c *hollowNodeConfig) addFlags(fs *pflag.FlagSet) {
 }
 
 func (c *hollowNodeConfig) createClientConfigFromFile() (*restclient.Config, error) {
-	clientConfig, err := clientcmd.LoadFromFile(c.KubeconfigPath)
+	clientConfigs, err := clientcmd.LoadFromFile(c.KubeconfigPath)
```
jshaofuturewei: I have made changes to kubeconfigfilepath to have multiple kubeconfig files. For example, /var/run/kubernetes/apiserver1.kubeconfig, /var/run/kubernetes/apiserver2.kubeconfig.

Just for review, no need to resolve.

cmd/kubemark/hollow-node.go
```diff
@@ -93,18 +93,20 @@ func (c *hollowNodeConfig) addFlags(fs *pflag.FlagSet) {
...
 	if err != nil {
 		return nil, fmt.Errorf("error while loading kubeconfig from file %v: %v", c.KubeconfigPath, err)
 	}
-	config, err := clientcmd.NewDefaultClientConfig(*clientConfig, &clientcmd.ConfigOverrides{}).ClientConfig()
+	configs, err := clientcmd.NewDefaultClientConfig(*clientConfigs, &clientcmd.ConfigOverrides{}).ClientConfig()
```
jshaofuturewei: It is another design issue. Originally we only have one kubeconfig and it is easy to override. Now we have multiple kubeconfig files. Do we need to identify a particular kubeconfig file to overwrite or overwrite all of them?

pkg/controller/client_builder.go
```diff
@@ -196,24 +199,27 @@ func (b SAControllerClientBuilder) getAuthenticatedConfig(sa *v1.ServiceAccount,
...
+
+	for _, config := range clientConfigCopy.GetAllConfigs() {
+		config.NegotiatedSerializer = legacyscheme.Codecs
+		client, err := restclient.UnversionedRESTClientFor(config)
+		if err != nil {
+			return nil, false, err
+		}
+		err = client.Get().AbsPath("/apis").Do().Error()
+		if apierrors.IsUnauthorized(err) {
+			klog.Warningf("Token for %s/%s did not authenticate correctly: %v", sa.Namespace, sa.Name, err)
```
jshaofuturewei: We might add config host to the warning so that we know which client causes the unauthorized error

cmd/kubelet/app/server.go
Sindica: That is the plan. I forgot to add todo note here.

Sindica: The actual dialing is not happening here so here does not need change.

cmd/kubemark/hollow-node.go
Sindica: Don't do overwrite. Just aggregate files or content.

PR #24, opened 2020-02-15
Hw multitenant network model

h-w-chen:
**What type of PR is this?**
/kind design
/kind documentation

k8s has the well-known network model. Introduction of multi-tenancy requires network resource isolation, imposing significant impact on network model.
This design describes the affected network components if multi-tenancy is in place, and proposes a network model of multi-tenancy k8s cluster.

Not yet. This PR is mainly for deign discussion.


h-w-chen:
closed - it should not have used branch of this repo directly

PR #30, opened 2020-02-17
Add design spec for partial runtime support

yb01:
*What type of PR is this?

/kind documentation

Add design spec for "partial runtime readiness feature in Arktos"

Which issue(s) this PR fixes:

None


docs/design-proposals/node/Partial-Runtime-Readiness-Handling.md
```diff
@@ -0,0 +1,121 @@
...
+
+This doc describes the proposal of the partial runtime readiness handling feature, including:
+1. Node readiness criteria with multiple runtime services on nodes
+2. Node runtime readiness state as part of the nodeStatus
+3. New scheduler predicate function for runtimeReadiness check
+
+## Requirements, principles and terminologies
+### Primary runtime of a cluster
+A primary runtime is the runtime service used for Kubernetes system daemon pods or pod with kube-system namespace. it is 
+usually a container runtime service given that the daemon runs as a process on the node.
```
XiaoningDing: What does "the daemon runs as a process" mean? DaemonSet pods also runs as containers or VMs, depending on what's specified in the daemonSet spec, right?

docs/design-proposals/node/Partial-Runtime-Readiness-Handling.md
```diff
@@ -0,0 +1,121 @@
...
+
+## Requirements, principles and terminologies
+### Primary runtime of a cluster
+A primary runtime is the runtime service used for Kubernetes system daemon pods or pod with kube-system namespace. it is 
+usually a container runtime service given that the daemon runs as a process on the node. 
+
+### Runtime service state and state changes
+1. A Runtime service can be added or removed from a node
+2. A runtime service state can be at Ready, NotReady or Unknown; where Ready means both network and compute ready for 
+the runtime; NotReady means one of them or neither of them are ready in a particular runtime service. Unknown is the intial
```
XiaoningDing: What do this mean: "network and compute ready for the runtime"?

docs/design-proposals/node/Partial-Runtime-Readiness-Handling.md
```diff
@@ -0,0 +1,121 @@
...
+1. A node Ready as long as the primary runtime service is ready, regardless of the state of the other registered runtime
+services on the node. i.e. cluster is ready to serve the same workload type its system pod workload type. 
+2. A graceful period of runtime NotReady will not trigger node from Ready to NotReady state. e.g. An interim primary 
+runtime service glitch WILL NOT change node Readiness. Graceful period can be defined as 
+N x node-status-update-period.  (Details to be defined)
+
+## Goal and no-goals
+
+### Goals
+1. A pod shall be scheduled to a node where the desired runtime service exists AND ready
```
XiaoningDing: Suggest to state explicitly that this guarantee is only for scheduling time.  After a pod is scheduled, the corresponding runtime could still fail in run time. For now we don't manage the failures in run time. We rely on the existing Kubernetes mechanism to detect failure and recreate pods.

docs/design-proposals/node/Partial-Runtime-Readiness-Handling.md
```diff
@@ -0,0 +1,121 @@
...
+2. Expand NodeCondition array with NodeContainerRuntimeReady and NodeVmRuntimeReady conditions   
+   
+       // registered runtime readiness for each workload type
+        NodeContainerRuntimeReady NodeConditionType = "ContainerRuntimeReady"
+        NodeVmRuntimeReady        NodeConditionType = "VmRuntimeReady"
+
+### Component changes
+
+### Kubelet
+1. Define isPrimary
```
XiaoningDing: Could you explain why we need to define "isPrimary"?

docs/design-proposals/node/Partial-Runtime-Readiness-Handling.md
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+
+## Discussion
+#### NodeSelectorPredicate vs. NodeConditionCheckPredicate
+Even node selector is commonly used to predicate pod-node binding for some special situations, such as current virtlet 
+Daemonset deployment uses extraRuntime=virtlet as the node selector to ensure the Virtlet runtime service is deployed 
+only on the node with the label. However, given the nature and the current code flow on node readiness with its runtime 
+service state, it is favored to expand the node condition check predicate function as the scheduler for this purpose. 
+so that the node status update routine won't need to update the node labels when runtime state changes, and the POD
```
XiaoningDing: You don't need to update the node labels, but you need to update NodeStatus, right? :)

docs/design-proposals/node/Partial-Runtime-Readiness-Handling.md
```diff
@@ -0,0 +1,121 @@
...
+1. A Runtime service can be added or removed from a node
+2. A runtime service state can be at Ready, NotReady or Unknown; where Ready means both network and compute ready for 
+the runtime; NotReady means one of them or neither of them are ready in a particular runtime service. Unknown is the intial
+state of any given runtime services, and indicates state of a given runtime is not set due to various reasons, such as 
+error querying runtime service endpoint.
+3. A runtime service state change can be Ready->NotReady OR NotReady->Ready
+
+### Node Readiness
+1. A node Ready as long as the primary runtime service is ready, regardless of the state of the other registered runtime
+services on the node. i.e. cluster is ready to serve the same workload type its system pod workload type.
```
XiaoningDing: Do we really need the concept of "primary runtime"? How about we mark a node as ready as long as any runtime is ready on this node?

docs/design-proposals/node/Partial-Runtime-Readiness-Handling.md
```diff
@@ -0,0 +1,121 @@
...
+2. A graceful period of runtime NotReady will not trigger node from Ready to NotReady state. e.g. An interim primary 
+runtime service glitch WILL NOT change node Readiness. Graceful period can be defined as 
+N x node-status-update-period.  (Details to be defined)
+
+## Goal and no-goals
+
+### Goals
+1. A pod shall be scheduled to a node where the desired runtime service exists AND ready
+
+### No-goals
```
XiaoningDing: Suggest that we clearly state this check is only for scheduling time. So any run time activities (migration, eviction) is not relevant.

docs/design-proposals/node/Partial-Runtime-Readiness-Handling.md
yb01: Cluster daemons can surely run both as container or VMs ( this is in fact one of reasons partial runtime readiness is needed ). The word "process" is used here is trying to emphasize the daemons most likely lightweight and with container workload. i will alter the wording to clarify it.

docs/design-proposals/node/Partial-Runtime-Readiness-Handling.md
yb01: A container runtime service readiness include both compute part and network part readiness.  this is to refer to the network readiness in the runtime service.

docs/design-proposals/node/Partial-Runtime-Readiness-Handling.md
yb01: updated spec to explicitly stated that and also mentioned no-goal section.
 A pod shall be scheduled to a node where the desired runtime service exists AND ready per the node status at API server 
   at the time when the POD is being scheduled.

docs/design-proposals/node/Partial-Runtime-Readiness-Handling.md
yb01: either update nodestatus( node conditions) or labels will work functionality wise. and they will need to be updated anytime the runtime service state is changed at the node, either labels or node conditions. it is currently decided to use node status simply follow the current code flow for single runtime service status. i am actually open for changes especially the node conditions probably will be "deprecated" in the future.
i will update the doc to make it clear as a least effort to clarify the reasoning here.

docs/design-proposals/node/Partial-Runtime-Readiness-Handling.md
yb01: this is changed in addressing the previous comments already. the cases for eviction etc is actually already listed in the no-goal section.

docs/design-proposals/node/Partial-Runtime-Readiness-Handling.md
yb01: as we discussed offline. this is needed. i will add more details in the doc to explain it.

docs/design-proposals/node/Partial-Runtime-Readiness-Handling.md
```diff
@@ -9,20 +9,40 @@ ready or notReady. As a result, the node status and node readiness management ne
 situations on the node, and the pod placement logic at the scheduler needs to be changed to ensure POD was scheduled to 
 a node with desired runtime service.
 
+Initially Arktos is currently implemented as "NodeReady requires ANY one of its runtime services ready" for best resource 
+utilization, i.e. the scheduler can schedule the pod to a node as long as its desired runtime service ready on the node.
+However, this implementation carries a drawback with PODs with cross workload type dependencies which can be described
+as followings:
+
```
XiaoningDing: "Initially...currently..."

docs/design-proposals/node/Partial-Runtime-Readiness-Handling.md
```diff
@@ -9,20 +9,40 @@ ready or notReady. As a result, the node status and node readiness management ne
...
+1. Container workload is a simpler way to implement the daemonset pod than VM workload. 
+2. Currently Kubernetes daemonset pods are container based, default to container workload will make the migration to Arktos
+   more straightforward. 
 
 ### Runtime service state and state changes
 1. A Runtime service can be added or removed from a node
 2. A runtime service state can be at Ready, NotReady or Unknown; where Ready means both network and compute ready for 
-the runtime; NotReady means one of them or neither of them are ready in a particular runtime service. Unknown is the intial
+the runtime; NotReady means one of them or neither of them are ready in a particular runtime service. Unknown is the initial
```
XiaoningDing: I don't suggest we use this: "where Ready means both network and compute ready for 
the runtime; NotReady means one of them or neither of them are ready in a particular runtime service.".  The runtime implementation is a black box to the node agent.  If it reports ready then it's ready, otherwise it's not ready.  If it's not ready, it could be many reasons (not only network and storage).  Node agent shouldn't care about the specific reasons.

docs/design-proposals/node/Partial-Runtime-Readiness-Handling.md
```diff
@@ -9,20 +9,40 @@ ready or notReady. As a result, the node status and node readiness management ne
...
+However, this implementation carries a drawback with PODs with cross workload type dependencies which can be described
+as followings:
+
+A cluster is setup and supports two runtime services, containerD and Arktos VM runtime,  and their corresponding workload
+types, i.e. Container workload type and VM workload type. The network agent, Alcor agent, was deployed as a cluster daemonset,
+which is running as container workload. All tenant level workload PODs, either containers or VMs will need Alcor's agent in
+order to have its network up. A pod can be scheduled to a node ONLY VM runtime ready. However, the pod cannot run because
+the network agent failed to setup the network for it.
+ 
+The following diagram illustrates the node status dependency, and cross workload type pod dependencies.
```
XiaoningDing: "cross workload type pod dependencies" is very vague and hard to detect in our system.  You can simply say we require container runtime to always be deployed because most cluster services are deployed as containers. And we don't want to force users to change them to vm-based YAML. Alcor agent is one example.

docs/design-proposals/node/Partial-Runtime-Readiness-Handling.md
yb01: fixed. thanks

docs/design-proposals/node/Partial-Runtime-Readiness-Handling.md
yb01: these condition checks in nodestatus is not added by the partial runtime feature.  i.e. kubelet's RuntimeServiceCondition function always checks for both compute and container network readiness for a given runtime service. This is an existing function. the description in the doc basically explains and emphasizes this current logic.

docs/design-proposals/node/Partial-Runtime-Readiness-Handling.md
yb01: reworded the use case for clarity. i guess your comment have been stated in the "Primary runtime of a cluster" section when it explains why the container runtime as default primary in our implmentation.

docs/design-proposals/node/Partial-Runtime-Readiness-Handling.md
XiaoningDing: I see. Thanks for the clarification.

PR #37, opened 2020-02-19
Aggregate watch channels for multiple api server supports

Sindica:


staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher_test.go
```diff
@@ -118,27 +119,27 @@ func testWatch(t *testing.T, recursive bool) {
...
 
 func TestDeleteTriggerWatch(t *testing.T) {
 	ctx, store, cluster := testSetup(t)
 	defer cluster.Terminate(t)
 	key, storedObj := testPropogateStore(ctx, t, store, &example.Pod{ObjectMeta: metav1.ObjectMeta{Name: "foo"}})
-	w, err := store.Watch(ctx, key, storedObj.ResourceVersion, storage.Everything)
-	if err != nil {
-		t.Fatalf("Watch failed: %v", err)
+	aw := store.Watch(ctx, key, storedObj.ResourceVersion, storage.Everything)
+	if aw.GetFirstError() != nil {
```
jshaofuturewei: maybe allerrors?

Sindica: For error case, we need to implement retry. It can be done in watch.go. For test, right now, there are too many places to change. Use FirstError for now.

PR #55, opened 2020-02-21
Test YAML files for VM persisted storage verification

yb01:
**What type of PR is this?**
Test yaml files for VM persisted storage verification

Test verification


test/yaml/vm_with_persistedVolumes/vmkey
```diff
@@ -0,0 +1,27 @@
+-----BEGIN RSA PRIVATE KEY-----
```
h-w-chen: what is this private key for? is it OK to expose to anyone?

yb01: this is the keypair ONLY used in the test to access to the VMs and only valid if the pubkey is specified in the test yaml. it is OK to make it public -- in fact, they are taken from virtlet project anyways.

PR #72, opened 2020-02-26
Added more partition configs to apiserver.config

jshaofuturewei:
As title


staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go
```diff
@@ -214,33 +214,36 @@ func (wc *watchChan) startWatching(watchClosedCh chan struct{}) {
...
 				klog.Errorf("watch chan error: %v", err)
 				wc.sendError(err)
 				return
 			}
-			wc.sendEvent(parsedEvent)
+			for _, e := range wres.Events {
+				parsedEvent, err := parseEvent(e)
+				if err != nil {
+					klog.Errorf("watch chan error: %v", err)
+					wc.sendError(err)
```
Sindica: When there is an error, there might be retry. wc object could be reused. But key was already updated so it will have unexpected behavior. That's why I don't like update wc.key in line 219

staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go
```diff
@@ -214,33 +214,36 @@ func (wc *watchChan) startWatching(watchClosedCh chan struct{}) {
...
-	wch := wc.watcher.client.Watch(wc.ctx, wc.key, opts...)
-	for wres := range wch {
-		if wres.Err() != nil {
-			err := wres.Err()
-			// If there is an error on server (e.g. compaction), the channel will return it before closed.
-			klog.Errorf("watch chan error: %v", err)
-			wc.sendError(err)
-			return
+	updatedKeys, updatedOptions := GetKeyAndOptFromPartitionConfig(wc.key, wc.partitionConfig)
+	for idx, updatedKey := range updatedKeys {
```
Sindica: Do you need separate go routine for each key?

staging/src/k8s.io/apiserver/pkg/storage/interfaces.go
```diff
@@ -246,5 +246,5 @@ type Interface interface {
 // If Begin is empty, it is left unbounded.
 // If End is empty, it is right unbounded.
 type Interval struct {
-	Begin, End string
+	Begin, End, Whitelist string
```
Sindica: Whilelist does need to have a copy for each interval

Sindica: There is no code reads whilelist from config file or initailze it somewhere.

jshaofuturewei: For flexibility, its tenant can be default, system or something else

Here is the code to initialize whitelist
if size >= 4 {
			m[strings.TrimSpace(slices[0])] = storage.Interval{Begin: strings.TrimSpace(slices[1]), End: strings.TrimSpace(slices[2]), Whitelist: strings.TrimSpace(slices[3])}
		}

staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go
jshaofuturewei: I removed it and keep it as the old way

jshaofuturewei: More investigation here.  All the keys go to the get operation which is stateless.

op := clientv3.OpGet(key, opts...)
	end := op.RangeBytes()
	pfxBegin, pfxEnd := prefixInterval(w.pfx, []byte(key), end)
	if pfxEnd != nil {
		opts = append(opts, clientv3.WithRange(string(pfxEnd)))
	}

	wch := w.Watcher.Watch(ctx, string(pfxBegin), opts...)

hack/apiserver.config
```diff
@@ -0,0 +1,27 @@
+/registry/actions/,tenant_begin,tenant_end,default
```
Sindica: Can we just hardcode default tenant as always watch in the code? This config seems involving more complicated code and not generic

staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher_test.go
```diff
@@ -378,134 +378,270 @@ func TestGetKeyAndOptFromPartitionConfig(t *testing.T) {
...
 		{
 			name: "opt with interval beginning and end via key",
 			key:  "registry/pods/",
 			partitionedConfig: map[string]storage.Interval{
 				"registry/pods/": {
-					Begin: "tenant1",
-					End:   "tenant2",
+					Begin:     "tenant1",
+					End:       "tenant2",
+					Whitelist: "",
```
Sindica: Not necessary to introduce a whitelist

staging/src/k8s.io/apiserver/pkg/storage/storagebackend/factory/etcd3.go
```diff
@@ -226,8 +226,8 @@ func parseConfig(configFileName string) (map[string]storage.Interval, error) {
 	for _, line := range lines {
 		slices := strings.Split(line, ",")
 		size := len(slices)
-		if size >= 3 {
-			m[strings.TrimSpace(slices[0])] = storage.Interval{Begin: strings.TrimSpace(slices[1]), End: strings.TrimSpace(slices[2])}
+		if size >= 4 {
+			m[strings.TrimSpace(slices[0])] = storage.Interval{Begin: strings.TrimSpace(slices[1]), End: strings.TrimSpace(slices[2]), Whitelist: strings.TrimSpace(slices[3])}
```
Sindica: Perhaps just use a constant "default" as the value of whitelist

staging/src/k8s.io/apiserver/pkg/storage/interfaces.go
Sindica: Does Whitelist support multiple tenants? If not, don't bother make it configurable.

hack/apiserver.config
jshaofuturewei: Removed

staging/src/k8s.io/apiserver/pkg/storage/interfaces.go
jshaofuturewei: Removed

staging/src/k8s.io/apiserver/pkg/storage/storagebackend/factory/etcd3.go
jshaofuturewei: Updated

PR #77, opened 2020-03-03
show resource tenanted info in command kubectl api-resources

chenqianfzh:
This is part of the work to make api-resource discovery support multi-tenancy. With this change, the output of "kubectl api-resources" will include the info whether the resource is "tenanted".

Test output:

qianchen@qianchen-VirtualBox:~/gopath/chenqian/arktos$ cluster/kubectl.sh api-resources
NAME                              SHORTNAMES   APIGROUP                       TENANTED   NAMESPACED   KIND
actions                           ac                                          true       true         Action
bindings                                                                      true       true         Binding
componentstatuses                 cs                                          false      false        ComponentStatus
configmaps                        cm                                          true       true         ConfigMap
controllerinstances               co                                          false      false        ControllerInstance
endpoints                         ep                                          true       true         Endpoints
events                            ev                                          true       true         Event
limitranges                       limits                                      true       true         LimitRange
namespaces                        ns                                          true       false        Namespace
nodes                             no                                          false      false        Node
persistentvolumeclaims            pvc                                         true       true         PersistentVolumeClaim
persistentvolumes                 pv                                          true       false        PersistentVolume

### DIFF
--- a/pkg/kubectl/cmd/apiresources/apiresources.go
+++ b/pkg/kubectl/cmd/apiresources/apiresources.go
@@ -48,6 +48,12 @@ var (
 		# Print the supported non-namespaced resources
 		kubectl api-resources --namespaced=false
 
+		# Print the supported tenant-scope resources
+		kubectl api-resources --tenanted=true
+
+		# Print the supported non-tenant-scope resources
+		kubectl api-resources --tenanted=false
+
 		# Print the supported API Resources with specific APIGroup
 		kubectl api-resources --api-group=extensions`)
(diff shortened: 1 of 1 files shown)


pdgetrf:
may be out of scope but do we want to show tenant name instead if it is tenanted?

chenqianfzh:
> may be out of scope but do we want to show tenant name instead if it is tenanted?

No, I don't think so. This command shows the API resource types, not any values in the resources.

PR #99, opened 2020-03-06
extract tenant from x509 cert

pdgetrf:
The format of the cert is changed to using /O (organization) for tenant and /OU (organization unit) for groups. Changed arktos-up.sh to create cert in this format.

added extraction of tenant from x509. after authentication, tenant info is available in requestAttributes.GetUser()

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go
+++ b/staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go
@@ -1,5 +1,6 @@
 /*
 Copyright 2014 The Kubernetes Authors.
+Copyright 2020 Authors of Arktos - file modified.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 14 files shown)


staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go
```diff
@@ -186,10 +187,23 @@ var CommonNameUserConversion = UserConversionFunc(func(chain []*x509.Certificate
...
+		return nil, false, nil
+	}
+
+	// tenant exists in the cert
+	if (len(chain[0].Subject.Organization)>0) {
+		return &authenticator.Response{
+			User: &user.DefaultInfo{
+				Name:   chain[0].Subject.CommonName,
+				Groups: chain[0].Subject.OrganizationalUnit,
+				Tenant: chain[0].Subject.Organization[0],
```
XiaoningDing: What will happen if there is no "/O" set in the subject?  It seems the new code didn't check this situation.

PR #103, opened 2020-03-06
set the tenant info in Object per user info

chenqianfzh:
Test result in local dev box:

(I created a context named "local" where the tenant of the user is "qian")

qianchen@qianchen-VirtualBox:~/gopath/chenqian/arktos$ cluster/kubectl.sh --context=local create ns qc-test
namespace/qc-test created

qianchen@qianchen-VirtualBox:~/gopath/chenqian/arktos$ cluster/kubectl.sh get namespaces --tenant qian
NAME      STATUS   AGE     TENANT
qc-test   Active   2m51s   qian

qianchen@qianchen-VirtualBox:~/gopath/chenqian/arktos$ cluster/kubectl.sh --context=local create ns qc-test --tenant another_tenant
Error from server (InternalError): an error on the server ("Internal Server Error: \"/api/v1/tenants/another_tenant/namespaces\": User under tenant qian is not allowed to access object under tenant another_tenant.") has prevented the request from succeeding

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/endpoints/filters/tenantInfo.go
+++ b/staging/src/k8s.io/apiserver/pkg/endpoints/filters/tenantInfo.go
@@ -17,30 +17,30 @@ limitations under the License.
 package filters
 
 import (
-	"errors"
 	"fmt"
 	"net/http"
 
 	"k8s.io/apiserver/pkg/endpoints/handlers/responsewriters"
 	"k8s.io/apiserver/pkg/endpoints/request"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// WithTenantInfo set the tenant in the requestInfo based on the authentication result.
+// WithTenantInfo set the tenant in the requestInfo based on the user info from the authentication result.
 func WithTenantInfo(handler http.Handler) http.Handler {
 	return http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 		ctx := req.Context()
 		requestInfo, exists := request.RequestInfoFrom(ctx)
-		if !exists {
-			responsewriters.InternalError(w, req, errors.New("no request info found for request"))
-			return
-		}
-
 		tenantInRequestor := ""
 		requestor, exists := request.UserFrom(ctx)
-		if !exists || requestor.GetTenant() == "" {
-			// if we hit here and found that there is no requestor info in the request,
-			// it means that the authenticator allows it in, it is mostly we are in a dev environment
-			tenantInRequestor = "system"
+		if !exists || requestor.GetTenant() == metav1.TenantNone {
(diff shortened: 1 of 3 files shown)


pdgetrf:
is the tenant of the namespace set in the context?

staging/src/k8s.io/apiserver/pkg/server/config.go
```diff
@@ -588,6 +588,7 @@ func DefaultBuildHandlerChain(apiHandler http.Handler, c *Config) http.Handler {
 	handler = genericapifilters.WithAudit(handler, c.AuditBackend, c.AuditPolicyChecker, c.LongRunningFunc)
 	failedHandler := genericapifilters.Unauthorized(c.Serializer, c.Authentication.SupportsBasicAuth)
 	failedHandler = genericapifilters.WithFailedAuthenticationAudit(failedHandler, c.AuditBackend, c.AuditPolicyChecker)
+	handler = genericapifilters.WithTenantInfo(handler)
```
XiaoningDing: This better to be after WithAudit(), so that we can keep the trace of original request.

staging/src/k8s.io/apiserver/pkg/endpoints/filters/tenantInfo.go
```diff
@@ -60,16 +60,16 @@ func WithTenantInfo(handler http.Handler) http.Handler {
...
 	}
 
-	if userTenant == "system" {
+	if userTenant == metav1.TenantSystem {
 		// for system user, we don't touch the tenant value in the objectMeta
 		return objectTenant, nil
 	}
 
-	if objectTenant != "" && objectTenant != userTenant {
+	if objectTenant != metav1.TenantNone && objectTenant != userTenant {
```
XiaoningDing: I think we'd better not do the check here.  This should be left to Authorizers.  Otherwise in future it will prevent us from supporting flexible cross-tenant access policies.

staging/src/k8s.io/apiserver/pkg/endpoints/filters/tenantInfo.go
```diff
@@ -60,16 +60,16 @@ func WithTenantInfo(handler http.Handler) http.Handler {
 
 // normalizeObjectTenant returns the tenant of the object based on what the user tenant is.
 func normalizeTenant(userTenant, objectTenant string) (string, error) {
```
XiaoningDing: We actually only change objectTenant in one situation, and keep objectTenant untouched in all other situations.  Suggest refactoring the codes of normalizeTenant() to make our intention explicit:

//the only case we change objectTenant: tenant info is not in URL and the request is issued by a non-system tenant
If objectTenant==metav1.TenantNone && userTenant != TenantSystem {
  return userTenant
}

return objectTenant

PR #108, opened 2020-03-07
Multi tenancy/short path

chenqianfzh:


staging/src/k8s.io/apiserver/pkg/endpoints/filters/tenantInfo.go
```diff
@@ -60,18 +60,14 @@ func WithTenantInfo(handler http.Handler) http.Handler {
...
+	if userTenant == metav1.TenantSystem {
 		// for system user, we don't touch the tenant value in the objectMeta
 		return objectTenant, nil
 	}
 
-	if objectTenant != "" && objectTenant != userTenant {
-		return "", fmt.Errorf("User under tenant %s is not allowed to access object under tenant %s.", userTenant, objectTenant)
-	}
-
 	return userTenant, nil
```
XiaoningDing: I think it's dangerous to return userTenant here by default. It will hurt some scenarios in future, like tenanA accesses tenantB resources (this is P2 for now but it's totally doable in our multi-tenancy design). There might be other unexpected scenarios.

As a filter, we should  keep the original objectTenant by default, and only do the replacement for cases that we know. For now there is only one case that will trigger replacement:  objectTenant==metav1.TenanNode && userTenant != metav1.TenanrSystem.   (check the comments in PR https://github.com/futurewei-cloud/arktos/pull/103 as well. thanks.)

chenqianfzh: updated.

PR #113, opened 2020-03-11
the tenant in the user info cannot be empty

chenqianfzh:
This address issue https://github.com/futurewei-cloud/arktos/issues/102
and complete the work in https://github.com/futurewei-cloud/arktos/issues/50

With this change, in the API request, the user info is required and the tenant info in it cannot be empty.


staging/src/k8s.io/apiserver/pkg/endpoints/filters/tenantinfo_test.go
```diff
@@ -0,0 +1,124 @@
...
+	ctx := req.Context()
+	requestInfo, _ := request.RequestInfoFrom(ctx)
+	if testCase.ExpectedTenant != requestInfo.Tenant {
+		t.Errorf("Test Case %q: expected tenant %q, but got %q", testCase.Name, testCase.ExpectedTenant, requestInfo.Tenant)
+	}
+}
+
+func TestTenantInfoRequest(t *testing.T) {
+	testCases := []tenantInfoTestCase{
+		// uncomment the next two test case when the following issue is solved:
```
chenqianfzh: Will remove these comments.  --Qian

PR #114, opened 2020-03-12
add rbac check for multi-tenancy

pdgetrf:
this check has two parts

1. verify that user tenant from x509 cert matches the requested resource tenant (from kubectl config). 
2. verify user's tenant is authorized by the rolebindings from the same tenant. 

both check skips if user belongs to the system tenant.

local test:

when user tenant from cert does not match the requested resource tenant
```bash
root@ip-172-31-27-32:~# kubectl --context=pengdu-context get pods
Error from server (Forbidden): pods is forbidden: User "pengdu" cannot list resource "pods" in API group "" in the namespace "pengdunamespace"
```

when user tenant matches cert and rolebinding tenant and is authorized
```bash
root@ip-172-31-27-32:~# kubectl --context=pengdu-context get pods
No resources found.
```

when user tenant matches cert and rolebinding tenant but is not authorized
```bash
root@ip-172-31-27-32:~# kubectl --context=pengdu-context create -f vanilla.yaml
Error from server (Forbidden): unknown
```


pdgetrf:
fixing unit tests

pkg/registry/rbac/validation/rule.go
```diff
@@ -212,6 +216,14 @@ func (r *DefaultRuleResolver) VisitRulesFor(user user.Info, namespace string, vi
 		} else {
 			sourceDescriber := &roleBindingDescriber{}
 			for _, roleBinding := range roleBindings {
+
+				// verify user tenant matches rolebinding tenant
+				// "to remove check for default later when rolebinding can
```
chenqianfzh: Need to make changes to to get the rolebindings only for the given userTenant. We may have tens of thousands of userTenants, it is very inefficient if we check all the role bindings for all the tenants.

pkg/registry/rbac/validation/rule.go
```diff
@@ -176,11 +177,14 @@ func (d *roleBindingDescriber) String() string {
...
 func (r *DefaultRuleResolver) VisitRulesFor(user user.Info, namespace string, visitor func(source fmt.Stringer, rule *rbacv1.PolicyRule, err error) bool) {
+
+	userTenant := user.GetTenant()
+
 	if clusterRoleBindings, err := r.clusterRoleBindingLister.ListClusterRoleBindings(); err != nil {
 		if !visitor(nil, nil, err) {
 			return
 		}
-	} else {
+	} else if userTenant == "system" {
```
chenqianfzh: Minor: we have a const definition, metav1.TenantSystem = "system"

Also, we have metav1.TenantDefault = "default"

pkg/registry/rbac/validation/rule.go
```diff
@@ -176,11 +177,14 @@ func (d *roleBindingDescriber) String() string {
 }
 
 func (r *DefaultRuleResolver) VisitRulesFor(user user.Info, namespace string, visitor func(source fmt.Stringer, rule *rbacv1.PolicyRule, err error) bool) {
+
+	userTenant := user.GetTenant()
+
 	if clusterRoleBindings, err := r.clusterRoleBindingLister.ListClusterRoleBindings(); err != nil {
```
chenqianfzh: As we discussed before, the ClusterRole and ClusterRoleBinding will be changed to tenant-scope. I will send out a PR on that. And you can make changes to get the rolebindings for the specific tenant only.

pkg/registry/rbac/validation/rule.go
```diff
@@ -176,11 +177,14 @@ func (d *roleBindingDescriber) String() string {
 }
 
 func (r *DefaultRuleResolver) VisitRulesFor(user user.Info, namespace string, visitor func(source fmt.Stringer, rule *rbacv1.PolicyRule, err error) bool) {
+
+	userTenant := user.GetTenant()
```
chenqianfzh: Do we need to handle the case that userTenant is empty?

staging/src/k8s.io/apiserver/pkg/authentication/request/union/union.go
```diff
@@ -1,5 +1,6 @@
 /*
 Copyright 2014 The Kubernetes Authors.
```
chenqianfzh: weird. Why the copyright statement is changed but there is not change in this file?

pkg/apis/authorization/types.go
```diff
@@ -87,6 +87,8 @@ type ResourceAttributes struct {
 	Subresource string
 	// Name is the name of the resource being requested for a "get" or deleted for a "delete". "" (empty) means all.
 	Name string
+	// Tenant is the tenent of the resource being request
```
chenqianfzh: typO: tenent

pkg/registry/rbac/validation/rule.go
pdgetrf: userTenant being empty sounds to me like coming into border checkpoint without a valid passport.

pkg/registry/rbac/validation/rule.go
pdgetrf: fixed

pdgetrf:
will add more unit tests after fixing the current failing tests

pkg/registry/rbac/validation/rule.go
chenqianfzh: The change is now ready at https://github.com/futurewei-cloud/arktos/pull/117

staging/src/k8s.io/apiserver/pkg/authentication/request/union/union.go
pdgetrf: I had some changes but was later removed.

pkg/registry/rbac/validation/rule.go
pdgetrf: taken

pkg/registry/rbac/validation/rule.go
pdgetrf: your changes have been picked up for both rolebinding and clusterrolebinding. thanks!

pdgetrf:
@chenqianfzh this is ready for review now. All your related changes have been merged.

api/openapi-spec/swagger.json
```diff
@@ -19282,7 +19290,7 @@
   },
   "info": {
     "title": "Kubernetes",
-    "version": "v2.17.1"
```
chenqianfzh: revert this version change.

pdgetrf: didn't change this manually. I'll do another make update.

pdgetrf:
hitting this error now in INT

Internal error occurred: failed to allocate a serviceIP: range is full\",\"reason\":\"InternalError\",\"details\":{\"causes\":[{\"message\":\"failed to allocate a serviceIP: range is full\"}]},\"code\":500}\n"

PR #116, opened 2020-03-12
Etcd add member function

jshaofuturewei:
It changed the current way to install etcd. 
It will create a new etcd service 
Regression test is needed.

We can run hack/install-etcd.sh for etcd installation

we can also run hack/install-etcd.sh start to start an etcd cluster
or 
hack/install-etcd.sh add host2 http://host2_ip:2379 to add a member to the existing cluster


hack/install-etcd.sh
```diff
@@ -21,7 +21,40 @@ set -o errexit
...
+  case $1 in
+    add)
+      if [ -n "$2" ]  && [ -n "$3" ]; then
+        kube::etcd::add_member $2 $3
+      fi
+      ;;
+    start)
+      # install etcd if necessary
+      if ! [[ $(which etcd) ]]; then
+        if ! [ -f "${KUBE_ROOT}/third_party/etcd/etcd" ]; then
```
Sindica: Sometimes etcd is installed in different location. Perhaps need to check whether etcd is in path or not

hack/install-etcd.sh
```diff
@@ -21,7 +21,40 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+LOG_DIR=${LOG_DIR:-"/tmp"}
+
 KUBE_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 source "${KUBE_ROOT}/hack/lib/init.sh"
 
-kube::etcd::install
+if [ $# -lt 1 ] ; then
```
Sindica: Can you add manual about how to use the script?

hack/install-etcd.sh
```diff
@@ -21,7 +21,40 @@ set -o errexit
...
+    kube::etcd::install
+else
+  echo "The current operation is $1"
+  case $1 in
+    add)
```
Sindica: very fancy :)

hack/install-etcd.sh
jshaofuturewei: Added in the comments

jshaofuturewei: Added

PR #117, opened 2020-03-12
Multi tenancy/cluster role scope

chenqianfzh:


pkg/registry/rbac/clusterrole/policybased/storage.go
```diff
@@ -52,7 +52,7 @@ func (r *Storage) NamespaceScoped() bool {
 }
 
 func (r *Storage) TenantScoped() bool {
-	return false
+	return true
```
pdgetrf: what does this tenant scoped mean and why is it changed to true now?

pdgetrf: and if it's always going to be tenant-scoped, do we still need this boolean?

staging/src/k8s.io/client-go/informers/rbac/v1/clusterrole.go
```diff
@@ -51,23 +52,31 @@ func NewClusterRoleInformer(client kubernetes.Interface, resyncPeriod time.Durat
...
 
+func NewClusterRoleInformerWithMultiTenancy(client kubernetes.Interface, resyncPeriod time.Duration, indexers cache.Indexers, tenant string) cache.SharedIndexInformer {
+	return NewFilteredClusterRoleInformerWithMultiTenancy(client, resyncPeriod, indexers, nil, tenant)
+}
+
 // NewFilteredClusterRoleInformer constructs a new informer for ClusterRole type.
 // Always prefer using an informer factory to get a shared informer instead of getting an independent
 // one. This reduces memory footprint and number of connections to the server.
 func NewFilteredClusterRoleInformer(client kubernetes.Interface, resyncPeriod time.Duration, indexers cache.Indexers, tweakListOptions internalinterfaces.TweakListOptionsFunc) cache.SharedIndexInformer {
+	return NewFilteredClusterRoleInformerWithMultiTenancy(client, resyncPeriod, indexers, tweakListOptions, "default")
```
pdgetrf: metav1.TenantDefault?

and do we still want default tenant?

staging/src/k8s.io/client-go/listers/rbac/v1/clusterrole.go
```diff
@@ -63,3 +67,52 @@ func (s *clusterRoleLister) Get(name string) (*v1.ClusterRole, error) {
 	}
 	return obj.(*v1.ClusterRole), nil
 }
+
+// ClusterRoles returns an object that can list and get ClusterRoles.
+func (s *clusterRoleLister) ClusterRoles() ClusterRoleTenantLister {
+	return clusterRoleTenantLister{indexer: s.indexer, tenant: "default"}
```
pdgetrf: same question here about "default" tenant

pkg/registry/rbac/clusterrole/policybased/storage.go
chenqianfzh: This value, together with the func NamespaceScoped(), indicate the scope of a resource type:
TenantScoped(): true, NamespaceScoped: true -> NamespaceScoped
TenantScoped(): true, NamespaceScoped: false -> TenantScoped
TenantScoped(): false, NamespaceScoped: false -> ClusterScoped

staging/src/k8s.io/client-go/listers/rbac/v1/clusterrole.go
chenqianfzh: That is the backward-compatible version, where the tenant value is set to "default".

There is new version, ClusterRolesWithMultiTenancy(), which allows the "tenant" to take any value.

staging/src/k8s.io/client-go/informers/rbac/v1/clusterrole.go
chenqianfzh: That is the backward-compatible version, where the tenant value is set to "default".

There is new version, ClusterRolesWithMultiTenancy(), which allows the "tenant" to take any value.

PR #121, opened 2020-03-14
Added start apiserver and workload controller manager script

jshaofuturewei:
As title

### DIFF
--- a/hack/arktos-apiserver-partition.sh
+++ b/hack/arktos-apiserver-partition.sh
@@ -16,14 +16,185 @@
 # limitations under the License.
 
 # Kube-apiserver partitioned by tenant
-APISERVER_PARTITION_TENANT=${APISERVER_PARTITION_TENANT:-"tenant1"}
+PARTITION_CONFIG_DIR=${PARTITION_CONFIG_DIR:-"/var/run/kubernetes/"}
 
+VIRTLET_METADATA_DIR=${VIRTLET_METADATA_DIR:-"/var/lib/virtlet"}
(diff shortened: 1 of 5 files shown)


hack/lib/common.sh
```diff
@@ -0,0 +1,297 @@
+#!/usr/bin/env bash
```
Sindica: How is this common.sh? It has a lot of code similar to arktos-up but I don't see any change in arktos-up. What is the relationship? How is the script used? If you plan to refactor arktos-up, please do it in same commit. Also please include detailed user manual.

hack/arktos-apiserver-partition.sh
```diff
@@ -1,45 +1,45 @@
...
-}
-
-function start_apiserver {
-  echo "Running arktos-up.sh..."
-  source $(dirname "$0")/arktos-up.sh
-}
-
-"$@"
-
+#!/usr/bin/env bash
```
Sindica: Don't know whether there will be anything else in this script. If not, suggest rename to arktos-apiserver-up-with-partition.sh

hack/lib/common.sh
jshaofuturewei: I removed some common functions in arktos-up.sh
The script is for reusing the existing function in arktos-up.sh.

hack/arktos-apiserver-partition.sh
jshaofuturewei: I suppose to run the script like
bash hack/arktos-up.sh start_apiserver begin_tenant end_tenant

Sindica: Please make sure the manual is checked into our repo

PR #124, opened 2020-03-17
Add tenant controller

### LINKED ISSUE #43: Create a tenant controller to watch tenant objects, and perform initialization and cleanup work

**Why is this needed**:

XiaoningDing:
**What type of PR is this?**

 /kind feature

 This PR adds tenant controller. It watches newly created tenants object in system space, and then initialize the tenant space.
  For now the initialization only includes three initial namespaces, but later on more objects could be included like default networks.

Fixes #43

This PR only handles tenant initialization. Tenant deletion will be handled in a separate PR later on.

```release-note
Two command line parameters are added in kube-controller-manager, following the conventions of existing controllers:

--concurrent-tenant-syncs int32     Default: 10
 The number of tenant objects that are allowed to sync concurrently. Larger number means more responsive tenant initialization and termination, but more CPU (and network) load.
--tenant-sync-period duration     Default: 5m0s
  The period for syncing tenant life-cycle updates

```


staging/src/k8s.io/client-go/informers/rbac/v1alpha1/interface.go
```diff
@@ -59,7 +59,7 @@ func (v *version) ClusterRoles() ClusterRoleInformer {
 
 // ClusterRoleBindings returns a ClusterRoleBindingInformer.
```
chenqianfzh: I found I forgot to check in some auto-upated changes in https://github.com/futurewei-cloud/arktos/pull/117. :-(
This made this PR really messy.

I just sent out a remedy PR in https://github.com/futurewei-cloud/arktos/pull/125.

Could you rebase your PR after that remedy PR is in, so the PR will look cleaner?

XiaoningDing: Done.

api/openapi-spec/swagger.json
```diff
@@ -19195,7 +19195,7 @@
   },
   "info": {
     "title": "Kubernetes",
-    "version": "v2.17.1"
```
chenqianfzh: guess it is a bug in "make update". We need to revert this unintentional change.

cmd/kube-controller-manager/app/core.go
```diff
@@ -346,6 +347,25 @@ func startNamespaceController(ctx ControllerContext) (http.Handler, bool, error)
 	return startModifiedNamespaceController(ctx, namespaceKubeClient, nsKubeconfigs)
 }
 
+func startTenantController(ctx ControllerContext) (http.Handler, bool, error) {
+	// tenant controller will be even more chatty than namespace controller when doing cleanups.
+	tnKubeConfigs := ctx.ClientBuilder.ConfigOrDie("tenant-controller")
```
chenqianfzh: It is mentioned that a tenant controller is likely to be more chatty than a namespace-controller, do we need to enlarge the value of QPS or Burst?

If making a decision is unrealistic now, I would like some comments here for future code maintainer.

pkg/controller/tenant/tenant_controller.go
```diff
@@ -0,0 +1,198 @@
...
+	corelisters "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/kubernetes/pkg/controller"
+	"k8s.io/kubernetes/pkg/util/metrics"
+
+	"k8s.io/klog"
+)
+
+var tenant_defult_namespaces = [...]string{"default", "kube-system", "kube-public"}
```
chenqianfzh: please use const string definitions: "default" -> metav1.NamespaceDefault ...

pkg/controller/tenant/tenant_controller.go
```diff
@@ -0,0 +1,198 @@
...
+		}
+		defer tc.queue.Done(key)
+
+		err := tc.processQueue(key.(string))
+		if err == nil {
+			// no error, forget this entry and return
+			tc.queue.Forget(key)
+			return false
+		} else {
+			// rather than wait for a full resync, re-add the tenant to the queue to be processed
```
chenqianfzh: Instead of just adding this key back to queue, I prefer a max-retries strategy.

pkg/controller/tenant/tenant_controller.go
```diff
@@ -0,0 +1,198 @@
...
+
+	startTime := time.Now()
+	defer func() {
+		klog.V(4).Infof("Finished initializing tenant %q (%v)", tenantName, time.Since(startTime))
+	}()
+
+	createFailures := []error{}
+
+	// Create Default namespaces
+	for _, nsName := range tenant_defult_namespaces {
```
chenqianfzh: type: defult

pkg/controller/tenant/tenant_controller.go
```diff
@@ -0,0 +1,198 @@
...
+		ns := v1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{Tenant: tenantName, Name: nsName},
+		}
+		if _, err := tc.kubeClient.CoreV1().NamespacesWithMultiTenancy(tenantName).Create(&ns); err != nil && !errors.IsAlreadyExists(err) {
+			createFailures = append(createFailures, err)
+		}
+	}
+
+	ret := utilerrors.Flatten(utilerrors.NewAggregate(createFailures))
+	klog.Infof("Tenant initialization result of %v : %v", tenantName, ret)
```
chenqianfzh: The info here might be a bit confusing. When the initialization is successful, ret will be nil, the message will read" result: nil", which might make the log reader think sth is wrong.

pkg/controller/tenant/tenant_controller_test.go
```diff
@@ -0,0 +1,118 @@
...
+		Tenant                  *v1.Tenant
+		ExpectCreatedNamespaces sets.String
+	}{
+		"new-tenants": {
+			Tenant: &v1.Tenant{
+				ObjectMeta: metav1.ObjectMeta{
+					Name: "test-tenant-1",
+				},
+			},
+			ExpectCreatedNamespaces: sets.NewString("default", "kube-system", "kube-public"),
```
chenqianfzh: can we re-use the definition of tenant_default_namespaces created above? The namespace names should not be part of a test case config.

pkg/controller/tenant/tenant_controller_test.go
```diff
@@ -0,0 +1,118 @@
...
+		if len(tc.ExpectCreatedNamespaces) != len(actions) {
+			t.Errorf("%s: Expected to create namespaces %#v. Actual actions were: %#v", k, tc.ExpectCreatedNamespaces, actions)
+			continue
+		}
+
+		actualCreatedNamespaces := sets.NewString()
+		for i := 0; i < len(actions); i++ {
+			action := actions[i]
+			if !action.Matches("create", "namespaces") {
+				t.Errorf("%s: Unexpected action %s", k, action)
```
chenqianfzh: change the second "%s" to "%v" or "%#v", as action's type is not string.

cmd/kube-controller-manager/app/core.go
XiaoningDing: It's mostly chatty during tenant cleanup because it may need to delete lots of resources. But I'm not sure if we really need to increase QPS or burst, because deleting a tenant doesn't have to be very quickly.  Anyway, I think I can update comments for later consideration and evaluation.

pkg/controller/tenant/tenant_controller.go
XiaoningDing: Sure, will change.

pkg/controller/tenant/tenant_controller.go
XiaoningDing: :) changed

pkg/controller/tenant/tenant_controller.go
XiaoningDing: Thank you for capturing this. Changed. Will return nil if the initialization is successful.

pkg/controller/tenant/tenant_controller_test.go
XiaoningDing: updated.

pkg/controller/tenant/tenant_controller_test.go
XiaoningDing: Updated.

pkg/controller/tenant/tenant_controller.go
XiaoningDing: The current way seems a common pattern being followed by lots of controllers. I'd like to first finish the functionality following the current patterns, and then let's see how this controller and other controllers can be improved.  I created issue #131 to track this.

api/openapi-spec/swagger.json
XiaoningDing: Not sure if it's a bug in make update or something intentional( so that we need to manually change version number?) Don't know the right behavior here. I will revert this and I created an issue #132 to track.

PR #129, opened 2020-03-20
extract tenant from bearer token file

pdgetrf:
Issue https://github.com/futurewei-cloud/arktos/issues/38

This change extracts tenant from token (csv) file. 

The empty entry till the end indicates the new format that is followed by the tenant. Without the empty entry, it falls back to the old format while setting tenant to "system" internally.

Most are auto-gen files. The real change is in staging/src/k8s.io/apiserver/pkg/authentication/token/tokenfile/tokenfile.go

Old format:
```bash
[token],[user],[uid],[groups...],[other data]
```

New format:
```bash
[token],[user],[uid],[groups...],[other data],,[tenant]
```

To enable:
add the following to apiserver startup option
```bash
--token-auth-file=/path/to/tokenfile
```

For example, with the following csv file:
```bash
abc,pengdu,pengduuid,"admin",,tenant
```
Set the context with 
```bash
kubectl config set-credentials pengdu --token=abc
```

And run kubectl as:
```bash
kubectl --user pengdu get pods
```

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/authentication/token/tokenfile/tokenfile_test.go
+++ b/staging/src/k8s.io/apiserver/pkg/authentication/token/tokenfile/tokenfile_test.go
@@ -1,5 +1,6 @@
 /*
 Copyright 2014 The Kubernetes Authors.
+Copyright 2020 Authors of Arktos - file modified.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


pdgetrf:
adding the no merge tag for kubemark test as it requires manual validation. will remove once it's cleared.

staging/src/k8s.io/apiserver/pkg/authentication/token/tokenfile/tokenfile_test.go
```diff
@@ -107,6 +138,26 @@ token7,user7,uid7,"group1,group2",otherdata
 	}
 }
 
+func TestBadTokenFileMustHaveTenant(t *testing.T) {
```
chenqianfzh: the test case name is a bit weird to me. I guess you mean BadTokenFileMIssingTenant?

pdgetrf: good call. I'll use BadTokenFileMIssingRequiredTenant

PR #139, opened 2020-03-26
setup cpu count for VM

yb01:
setup VM vcpu as requested in the VM spec

issue 134

this fix assumes the VM type as resource specified in the spec as Guaranteed QoS model. this PR does NOT include the admission controller code to enforce it. this should be in other PR.

Below are the domain configured:

<domain type='qemu' xmlns:qemu='http://libvirt.org/schemas/domain/qemu/1.0'>
  <name>virtlet-944c97da-3555-vm</name>
  <uuid>944c97da-3555-5b5f-7b59-7d0096a6fe9f</uuid>
  <memory unit='KiB'>1048576</memory>
  <currentMemory unit='KiB'>1048576</currentMemory>
  <vcpu placement='static'>3</vcpu>
  <cputune>
    <shares>2</shares>
    <period>100000</period>
  </cputune>


pkg/kubelet/util/podConverter/podConverter.go
```diff
@@ -134,6 +139,18 @@ func ConvertVmPodToContainerPod(pod *v1.Pod) *v1.Pod {
...
+// Libvirt guestOS cpu resource doc:
+// https://libvirt.org/docs/libvirt-appdev-guide-python/en-US/html/libvirt_application_development_guide_using_python-Guest_Domains-Domain_Config-Memory_CPU.html
+// Libvirt VM domain vcpus specifies the limit of number of vcpus the guest OS can have
+// Arktos vm runtime doesn't support guest OS vcpu reset currently, the VM workload should be classified as Guaranteed
+// resource model.
+// API validation should enforce this when accepting the VM POD to avoid resource level misalignment since scheduler uses
+// requested resources in its Predicate function
+//
+func getVCpuCountForVm(vmSpec * v1.VirtualMachine) int64 {
+	return vmSpec.Resources.Limits.Cpu().Value()
```
XiaoningDing: Can we also handle the situation that limit is not specified but request is specified?

yb01: this should be in the admission controller for VM type as i mentioned in the PR which is not include in this PR as the original thought. i.e. the admission controller should enforce the limits is set and if request is also set, it should be equal to the limitvalue ( give only one vm is in the pod in our current design ). let me add those part as well if this is agreed to the right solution here as described below:

we can expand this PR with :

1. this current bug fix to set up the cpu count for VM --- which is in this commit.
2. admission controller to enforce the limit setting of VM, and if request is also set, it should be same as limit -- which is Guaranteed QoS class for VM type
3. fix bug in QoS to add VM type check as well.

XiaoningDing: I see. I think your code comment can simply say "for now we only support Guaranteed QoS class for VM pods".  It will clearly explain why you only check .limit.

XiaoningDing: Also: if you plan to do this way, you should check if .limit is available before read it, just in case the admission controller is disabled.

yb01: that should be OK. it just returned 0 as i tested it.

PR #145, opened 2020-03-27
add tenant to JWT bearer token authentication

pdgetrf:
Part 2 of https://github.com/futurewei-cloud/arktos/issues/38

JWT bearer token is used by service account (SA) authentication. A token is stored as part of the secret generated for a SA. Tenant was missing when a token is generated and as a result during authentication, SA and secrets were fetched only for the "default" tenant. This change does the following:

- add tenant info to the token
- change the authentication to fetch SA and secrets using tenant parsed from token

### DIFF
--- a/pkg/kubeapiserver/authenticator/config.go
+++ b/pkg/kubeapiserver/authenticator/config.go
@@ -1,5 +1,6 @@
 /*
 Copyright 2014 The Kubernetes Authors.
+Copyright 2020 Authors of Arktos - file modified.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/pkg/serviceaccount/jwt.go
+++ b/pkg/serviceaccount/jwt.go
@@ -1,5 +1,6 @@
 /*
 Copyright 2014 The Kubernetes Authors.
+Copyright 2020 Authors of Arktos - file modified.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -37,8 +38,10 @@ import (
 // ServiceAccountTokenGetter defines functions to retrieve a named service account and secret
 type ServiceAccountTokenGetter interface {
 	GetServiceAccount(namespace, name string) (*v1.ServiceAccount, error)
+	GetServiceAccountWithMultiTenancy(tenant, namespace, name string) (*v1.ServiceAccount, error)
 	GetPod(namespace, name string) (*v1.Pod, error)
 	GetSecret(namespace, name string) (*v1.Secret, error)
+	GetSecretWithMultiTenancy(tenant, namespace, name string) (*v1.Secret, error)
 }
 
 type TokenGenerator interface {
--- a/pkg/serviceaccount/legacy.go
+++ b/pkg/serviceaccount/legacy.go
@@ -1,5 +1,6 @@
 /*
 Copyright 2018 The Kubernetes Authors.
+Copyright 2020 Authors of Arktos - file modified.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -36,6 +37,7 @@ func LegacyClaims(serviceAccount v1.ServiceAccount, secret v1.Secret) (*jwt.Clai
 			ServiceAccountName: serviceAccount.Name,
 			ServiceAccountUID:  string(serviceAccount.UID),
 			SecretName:         secret.Name,
+			Tenant:             secret.Tenant,
 		}
 }
 
@@ -46,6 +48,7 @@ type legacyPrivateClaims struct {
 	ServiceAccountUID  string `json:"kubernetes.io/serviceaccount/service-account.uid"`
 	SecretName         string `json:"kubernetes.io/serviceaccount/secret.name"`
 	Namespace          string `json:"kubernetes.io/serviceaccount/namespace"`
+	Tenant             string `json:"kubernetes.io/serviceaccount/tenant"`
 }
 
 func NewLegacyValidator(lookup bool, getter ServiceAccountTokenGetter) Validator {
@@ -77,6 +80,10 @@ func (v *legacyValidator) Validate(tokenData string, public *jwt.Claims, private
 	if len(namespace) == 0 {
 		return nil, errors.New("namespace claim is missing")
 	}
+	tenant := private.Tenant
+	if len(tenant) == 0 {
+		return nil, errors.New("tenant claim is missing")
+	}
 	secretName := private.SecretName
 	if len(secretName) == 0 {
 		return nil, errors.New("secretName claim is missing")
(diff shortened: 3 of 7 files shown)


pkg/master/reconcilers/mastercount.go
```diff
@@ -212,7 +212,7 @@ func checkEndpointSubsetFormat(ss []corev1.EndpointSubset, ip string, ports []co
 		}
 	}
 	for _, addr := range sub.Addresses {
-		if addr.IP == ip { 
+		if addr.IP == ip {
```
pdgetrf: gofmt change. no related.

pkg/serviceaccount/legacy.go
```diff
@@ -97,9 +104,9 @@ func (v *legacyValidator) Validate(tokenData string, public *jwt.Claims, private
 
 	if v.lookup {
 		// Make sure token hasn't been invalidated by deletion of the secret
-		secret, err := v.getter.GetSecret(namespace, secretName)
+		secret, err := v.getter.GetSecretWithMultiTenancy(tenant, namespace, secretName)
 		if err != nil {
-			klog.V(4).Infof("Could not retrieve token %s/%s for service account %s/%s: %v", namespace, secretName, namespace, serviceAccountName, err)
+			klog.V(4).Infof("Could not retrieve token %s/%s for service account %s/%s/%s: %v", namespace, secretName, tenant, namespace, serviceAccountName, err)
```
XiaoningDing: Probably changing "token" to "secret" will make the message more accurate.

pkg/serviceaccount/legacy.go
```diff
@@ -77,6 +80,10 @@ func (v *legacyValidator) Validate(tokenData string, public *jwt.Claims, private
 	if len(namespace) == 0 {
 		return nil, errors.New("namespace claim is missing")
 	}
+	tenant := private.Tenant
```
XiaoningDing: I remember in current implementation  obj.tenant could be empty if it's in system space. Will this check break the service account functionality in system space?

pkg/serviceaccount/legacy.go
pdgetrf: fixed

pkg/serviceaccount/legacy.go
pdgetrf: what this check does is to extract tenant from the token provided "user". subsequently this tenant is used to extract user's other info for validation. if the token provided by user does not contain a tenant info, I don't think we should default it to "system", otherwise any user service account without a tenant in their token would end up being authenticated as a service account of the "system" tenant.

pkg/kubeapiserver/authenticator/config.go
```diff
@@ -1,5 +1,6 @@
 /*
 Copyright 2014 The Kubernetes Authors.
+Copyright 2020 Authors of Arktos - file modified.
```
chenqianfzh: Curious, why copyright statement is changed without any code change? Is it a make-up change for some miss is previous PRs?

pkg/serviceaccount/util.go
```diff
@@ -33,25 +33,26 @@ const (
...
 		UID:       uid,
+		Tenant:    tenant,
 	}).UserInfo()
 }
 
 type ServiceAccountInfo struct {
-	Name, Namespace, UID string
-	PodName, PodUID      string
+	Name, Namespace, UID    string
+	PodName, PodUID, Tenant string
```
chenqianfzh: Minor: Move Tenant to line of "Name, Namespace". They belong to the same semantic  group.

pkg/serviceaccount/jwt.go
```diff
@@ -37,8 +38,10 @@ import (
 // ServiceAccountTokenGetter defines functions to retrieve a named service account and secret
 type ServiceAccountTokenGetter interface {
 	GetServiceAccount(namespace, name string) (*v1.ServiceAccount, error)
+	GetServiceAccountWithMultiTenancy(tenant, namespace, name string) (*v1.ServiceAccount, error)
 	GetPod(namespace, name string) (*v1.Pod, error)
```
chenqianfzh: There is no multi-tenancy change for GetPod()? How can we get a pod in a non-default tenant?

pkg/serviceaccount/util.go
```diff
@@ -33,25 +33,26 @@ const (
 )
```
chenqianfzh: util_test.go is updated with multi-tenancy change.

plugin/pkg/admission/security/podsecuritypolicy/admission.go
```diff
@@ -211,7 +212,7 @@ func (p *Plugin) computeSecurityContext(a admission.Attributes, pod *api.Pod, sp
 	klog.V(4).Infof("getting pod security policies for pod %s (generate: %s)", pod.Name, pod.GenerateName)
 	var saInfo user.Info
 	if len(pod.Spec.ServiceAccountName) > 0 {
-		saInfo = serviceaccount.UserInfo(a.GetNamespace(), pod.Spec.ServiceAccountName, "")
+		saInfo = serviceaccount.UserInfo(a.GetNamespace(), pod.Spec.ServiceAccountName, "", pod.Tenant)
```
chenqianfzh: shall the tenant value be obtained from admission attributes or the pod? Not quite sure about the difference, yet I see that the namespace value is from the admission attributes.

pkg/serviceaccount/jwt.go
pdgetrf: GetPod() is used in claims.go and is not called in this change. this change is calling the Validate() function in legacy.go.

pdgetrf: replaced the default tenants in the tests

pkg/kubeapiserver/authenticator/config.go
pdgetrf: I probably have some change here and then removed. and the copyright sticks around.

pkg/serviceaccount/util.go
pdgetrf: fixed

pkg/serviceaccount/util.go
pdgetrf: what do you mean here?

plugin/pkg/admission/security/podsecuritypolicy/admission.go
pdgetrf: fixed

pkg/serviceaccount/util.go
chenqianfzh: Sorry, I mean, util_test.go is NOT updated.

pdgetrf: updated tests

pkg/serviceaccount/legacy.go
XiaoningDing: Thanks for clarification. At first I thought it was a resource object.

pkg/serviceaccount/legacy.go
```diff
@@ -112,7 +119,7 @@ func (v *legacyValidator) Validate(tokenData string, public *jwt.Claims, private
 		}
 
 		// Make sure service account still exists (name and UID)
-		serviceAccount, err := v.getter.GetServiceAccount(namespace, serviceAccountName)
+		serviceAccount, err := v.getter.GetServiceAccountWithMultiTenancy(tenant, namespace, serviceAccountName)
 		if err != nil {
 			klog.V(4).Infof("Could not retrieve service account %s/%s: %v", namespace, serviceAccountName, err)
```
chenqianfzh: Minor: Change to ("Could not retrieve service account %s/%s/%s: %v", tenant, namespace, serviceAccountName, err)

PR #147, opened 2020-03-27
add a public slides about Arktos current features and some planned features

XiaoningDing:
This slides talk about the current Arktos features and some features planned for next.  All information is public.

During previous discussions, some community members asked to upload to documents folder for easy sharing.


XiaoningDing:
This sounds reasonable to me.  I will close this PR and see where is the best place to put this. I don't know if we have a google drive or not.

PR #149, opened 2020-03-28
Add aggregated watch for kubectl

Sindica:


Sindica:
Will submit from another branch

PR #152, opened 2020-03-31
Get runtime deployment files from runtime repo

yb01:
**What type of PR is this?**
> /kind bug
fix issue https://github.com/futurewei-cloud/arktos/issues/151
to get runtime deployment files from runtime repo

https://github.com/futurewei-cloud/arktos/issues/151


hack/arktos-up.sh
```diff
@@ -18,6 +18,10 @@ PARTITION_CONFIG_DIR=${PARTITION_CONFIG_DIR:-"/var/run/kubernetes/"}
 
 VIRTLET_METADATA_DIR=${VIRTLET_METADATA_DIR:-"/var/lib/virtlet"}
 VIRTLET_LOG_DIR=${VIRTLET_LOG_DIR:-"/var/log/virtlet"}
+VIRTLET_DEPLOYMENT_FILES_DIR=${VIRTLET_DEPLOYMENT_FILES_DIR:-"/tmp"}
+
+# Default to vm-runtime master branch for now till we have a release and versioning setup
+VIRTLET_DEPLOYMENT_FILES_SRC=${VIRTLET_DEPLOYMENT_FILES_SRC:-"https://raw.githubusercontent.com/futurewei-cloud/arktos-vm-runtime/master/deploy"}
```
XiaoningDing: You probably want to add a version here. Otherwise in future any change in arktos-vm-runtime repo will break early versions of Arktos.

yb01: Agreed we need to get versioning support finalized to get a refined support ( as the comment mention ). current thinking for this PR is that:
1. branch implicitly carries major release versions. i.e. when we cut (V1 release, or for example, the 330 release ), both Arktos and Arktos runtime will be forked to 330 release branch and the runtime deployment source files will be persevered there.
2. for minor release or critical bug fixes. in dev branch,  they can be sync-ed with each dev branch and master branches should be always in sync.  cherry-picks can be made into released branches as bug fix as well on a needed basis.
3. once we have a release site with finalized release story, those files will be tar-ed into tarball with major and minor versions, along with the binaries. 

#3 can be done post 330 release. current PR should support #1 and #2.

yb01:
please hold off the merge until the runtime repo change were merged.

yb01:
Runtime PR https://github.com/futurewei-cloud/arktos-vm-runtime/pull/19 merged

hack/arktos-up.sh
XiaoningDing: For 1) and 2), I don't think these two repos should be coupled so tight. According to the description things will break if two repos have different branch policies.  These two repos could have two different group of developers. They should only reply on some simple public locations, such as  deployment/v1/xxx.yaml

yb01: as we discussed, a release branch release-330 was created at the vm-runtime repo. will change the default in Arktos to use this current release.

hack/arktos-up.sh
```diff
@@ -18,6 +18,10 @@ PARTITION_CONFIG_DIR=${PARTITION_CONFIG_DIR:-"/var/run/kubernetes/"}
 
 VIRTLET_METADATA_DIR=${VIRTLET_METADATA_DIR:-"/var/lib/virtlet"}
 VIRTLET_LOG_DIR=${VIRTLET_LOG_DIR:-"/var/log/virtlet"}
+VIRTLET_DEPLOYMENT_FILES_DIR=${VIRTLET_DEPLOYMENT_FILES_DIR:-"/tmp"}
+
+# Default to current arktos-vm-runtime release-330 branch
+VIRTLET_DEPLOYMENT_FILES_SRC=${VIRTLET_DEPLOYMENT_FILES_SRC:-"https://raw.githubusercontent.com/futurewei-cloud/arktos-vm-runtime/release-330/deploy"}
```
XiaoningDing, line 24: It's not suggested to use "release-330" as the version identifier. What will happen for next year's 330? :).  Consider using something like "release-0.5", "release-1.1", etc.

XiaoningDing: Also the versioning of runtime could be separate from Arktos. They don't need to be always same.  For example, in the next several months Arktos version will increase. But arktos-vm-runtime version can stay same if there are no updates on it.

yb01: it might be better we align with our docker image release number, which is currently 0.5.2.

XiaoningDing: Aligning with docker image release number sounds a good idea to me.  All these (binary, image, configuration files) should share a single release number to avoid confusion.

PR #154, opened 2020-03-31
Fixed the 2 environment issues bash head and eth0 interface

jshaofuturewei:
Fixed 2 environment issues
1. Commented /usr/bin/env bash 
2.  Add check condition statement for eth0

### DIFF
--- a/hack/arktos-up.sh
+++ b/hack/arktos-up.sh
@@ -1,4 +1,4 @@
-/s!/usr/bin/env bash
+#!/usr/bin/env bash
 
 # Copyright 2020 Authors of Arktos.
 #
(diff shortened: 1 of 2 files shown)


hack/arktos-up.sh
```diff
@@ -1,4 +1,4 @@
-/s!/usr/bin/env bash
+#/s!/usr/bin/env bash
```
pdgetrf, line 1: did you mean #!/usr/bin/env bash

jshaofuturewei: Yes.

hack/lib/etcd.sh
```diff
@@ -18,9 +18,14 @@
 # A set of helpers for starting/running etcd for tests
 
 ETCD_VERSION=${ETCD_VERSION:-3.3.10}
-INT_NAME="eth0"
+INT_NAME=${INT_NAME:-eth0}
```
vinaykul, line 21: This will still default to eth0 if not specified. Please add code to query the default interface using ip route.

jshaofuturewei: Updated it to ip route | awk '/default/ { print $5 }'

PR #159, opened 2020-04-02
Suppress extraneous updates of node status runtime readiness with no status change

h-w-chen:
**What type of PR is this?**
/kind bug
/kind performance optimization

node agent regularly checks workload runtimes' readiness, and reports to api server in node status updates - whether  the readiness has gone through material change or stays the same. Though logically benign, these excessive updates lead to massive allocations in the api server, causing its GC to work harder to clean them up, as illustrated in screenshot below (captured when there are 5K hollow nodes, 20K live pods)
beforefix
  
The solution is to skip readiness updates if it does not change at all. This leads to less loads to the api server - as shown by the below cpu usage screenshot (5K hollow nodes, 20K pods)
afterfix
  
NONE


h-w-chen:
need to rebase

PR #160, opened 2020-04-02
Suppress extraneous updates of node status runtime readiness with no status change

h-w-chen:
**What type of PR is this?**
/kind bug
/kind performance optimization

node agent regularly checks workload runtimes' readiness, and reports to api server in node status updates - whether  the readiness has gone through material change or stays the same. Though logically benign, these excessive updates lead to massive allocations in the api server, causing its GC to work harder to clean them up, as illustrated in screenshot below (captured when there are 5K hollow nodes, 20K live pods)
beforefix
  
The solution is to skip readiness updates if it does not change at all. This leads to less loads to the api server - as shown by the below cpu usage screenshot (5K hollow nodes, 20K pods)
afterfix
  
NONE


pkg/kubelet/nodestatus/setters.go
```diff
@@ -612,6 +612,10 @@ func getCurrentRuntimeReadiness(runtimeCondition *v1.NodeCondition, workloadType
 	statusSet := false
 	for _, status := range runtimeServiceStatus {
 		if status == true {
+			if runtimeCondition.Status == v1.ConditionTrue {
+				// runtime was already ready; no change
+				return runtimeCondition
+			}
 			runtimeCondition.Status = v1.ConditionTrue
 			runtimeCondition.Reason = fmt.Sprintf("At least one %s runtime is ready", workloadType)
 			recordEventFunc(v1.EventTypeNormal, fmt.Sprintf("%s is ready", runtimeCondition.Type))
```
Sindica, line 621: My understanding is that those events are used as healthy check report, they would update revision # and last updated time. Skip all the events that are the same might have other issue such as node was not refreshed in certain time period. Perhaps record those event in a longer interval.

Do you know currently how often it records such event for a healthy node?

pkg/kubelet/nodestatus/setters.go
```diff
@@ -612,6 +612,10 @@ func getCurrentRuntimeReadiness(runtimeCondition *v1.NodeCondition, workloadType
 	statusSet := false
 	for _, status := range runtimeServiceStatus {
 		if status == true {
+			if runtimeCondition.Status == v1.ConditionTrue {
```
XiaoningDing, line 615: This will work by not changing anything in runtimeCondition. But maybe it's better to do it outside of this function, and this function still returns latest status/reason/lastTransactionTime.  if we look at other similar conditions (memoryPressure, diskPressure, PIDPressure, etc), they all check and decide if to update conditions in setters, like functions MemoryPressureCondition(), DiskPressureCondition).  So I think adding comparisons there will make the comparison logic explicit and easy to understand, and also make styles consistent.

h-w-chen: it is reasonable to follow the convention and keep code consistent. will change accordingly.

yb01:
LGTM

PR #163, opened 2020-04-02
Update api server partitioning document + fix&refactor common scripts

Sindica:


docs/setup-guide/arktos-apiserver-partition.md
```diff
@@ -0,0 +1,154 @@
...
+$ cd $GOPATH/src/arktos
+$ ./hack/arktos-up.sh
+```
+Note: this will soon be changed.
+
+### Start apiserver on a worker host
+
+. Verify we can access etcd from the worker host
+```
+$ curl http://<1st master server ip>:2379
```
jshaofuturewei, line 34: It might not be a good way to verify etcd 
a simple way to check curl -L http://[ip]:2379/v2/keys

docs/setup-guide/arktos-apiserver-partition.md
```diff
@@ -0,0 +1,155 @@
...
+```
+It is ok if returns "404 page not found"
+
+. Start 2nd Api server (in a different host)
+```
+# configure api server with data partition
+$ export APISERVER_SERVICEGROUPID=2
+$ ./hack/arktos-apiserver-partition.sh start_apiserver <1st master server ip>
```
jshaofuturewei, line 42: etcd ip might be better.

docs/setup-guide/arktos-apiserver-partition.md
```diff
@@ -0,0 +1,155 @@
...
+```
+
+### Get kubeconfig from master servers to client host
+. On 1st and 2nd api server hosts, run the following command and get kubeconfig file that connects to the api server
+```
+$ cd $GOPATH/src/arktos
+$ ./hack/create-kubeconfig.sh command <path to where you want to generate kubeconfig file>
+# Copy generated print out from the above command and run it in shell
```
jshaofuturewei, line 50: in the shell of different host

docs/setup-guide/arktos-apiserver-partition.md
Sindica: Updated

docs/setup-guide/arktos-apiserver-partition.md
Sindica: If it runs in different hose, all username, clustername, etc will be using the other host's setting. So it needs to be in the same host.

PR #164, opened 2020-04-03
fixed a few issues for multi-tenancy access control

pdgetrf:
Fixed a few issues exposed through further test (for demo).

1. service accounts created by the control plane is still using the default tenant. This will cause failure when such service account try to make changes to other tenant such as the case where tenant-controller tries to create namespaces for non-default tenant. 

2. GetRole requires both tenant and namespace

### DIFF
--- a/pkg/registry/rbac/validation/rule.go
+++ b/pkg/registry/rbac/validation/rule.go
@@ -102,6 +102,7 @@ func NewDefaultRuleResolver(roleGetter RoleGetter, roleBindingLister RoleBinding
 
 type RoleGetter interface {
 	GetRole(namespace, name string) (*rbacv1.Role, error)
+	GetRoleWithMultiTenancy(tenant, namespace, name string) (*rbacv1.Role, error)
 }
 
 type RoleBindingLister interface {
@@ -250,7 +251,7 @@ func (r *DefaultRuleResolver) VisitRulesFor(user user.Info, namespace string, vi
 			if !applies {
 				continue
 			}
-			rules, err := r.GetRoleReferenceRules(roleBinding.RoleRef, namespace)
+			rules, err := r.GetRoleReferenceRulesWithMultiTenancy(roleBinding.RoleRef, roleBinding.Tenant, namespace)
 			if err != nil {
 				if !visitor(nil, nil, err) {
 					return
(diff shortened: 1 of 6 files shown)


pkg/registry/rbac/validation/rule.go
```diff
@@ -361,6 +384,18 @@ func (r *StaticRoles) GetRole(namespace, name string) (*rbacv1.Role, error) {
 	return nil, errors.New("role not found")
 }
 
+func (r *StaticRoles) GetRoleWithMultiTenancy(namespace, tenant, name string) (*rbacv1.Role, error) {
```
chenqianfzh, line 387: Non-critical: it is better to arrange the param in the order of tenant/ns/name to show the hierarchy of these concepts

pkg/registry/rbac/validation/rule.go
```diff
@@ -361,6 +384,18 @@ func (r *StaticRoles) GetRole(namespace, name string) (*rbacv1.Role, error) {
...
+	if len(namespace) == 0 {
```
chenqianfzh, line 388: We also need to check that tenant is not empty.

pdgetrf: fixed

pkg/registry/rbac/validation/rule.go
pdgetrf: fixed

pkg/registry/rbac/validation/rule_test.go
```diff
@@ -76,7 +76,7 @@ func TestDefaultRuleResolver(t *testing.T) {
 	staticRoles1 := StaticRoles{
 		roles: []*rbacv1.Role{
 			{
-				ObjectMeta: metav1.ObjectMeta{Namespace: "namespace1", Name: "readthings"},
+				ObjectMeta: metav1.ObjectMeta{Namespace: "namespace1", Name: "readthings", Tenant: metav1.TenantSystem},
```
chenqianfzh, line 79: we should add a non-system tenant case

pdgetrf: will file a PR after this one is merged.

PR #179, opened 2020-04-06
Rename data partition key to general purpose

Sindica:


docs/design-proposals/apiserver/Data-Partition-Design.md
```diff
@@ -126,17 +126,17 @@ type DataPartitionConfig struct {
...
 	metav1.ObjectMeta
 
-	// Start tenant is inclusive
-	StartTenant string
+	// Range start is inclusive
+	RangeStart string
 
 	// Whether this is an open end start
-	IsStartTenantValid bool
+	IsRangeStartValid bool
```
jshaofuturewei, line 133: Is IsRangeStartValid related to inclusive, like true = inclusive, false = exclusive?

Sindica: true/false is used as limited or unlimted

jshaofuturewei: Do you mean ∞? 
If it is false(unlimited), the RangeStart/End will be meaningless, right?
There is a partition data equals statement  in staging/src/k8s.io/apiserver/pkg/storage/datapartition/datapartitionmanager.go made me a little confusing, if 
 a  and b have  both IsRangeStartValid and IsRangeEndValid false, they will be equal to each other no matter that they have different RangeStart/End value?

Sindica: yes. If IsRangeStartValid == false, RangeStart will be ignored.

jshaofuturewei: Thanks.

PR #181, opened 2020-04-07
add tests with regular tenant or ruleResolver.RulesFor

pdgetrf:
RulesFor was missing test cases using a regular tenant. Added now as promised at a [previous PR](https://github.com/futurewei-cloud/arktos/pull/164#discussion_r403387782)

Also changed the output format of the test to show expected and actual instead of the diff. With the diff output one can't tell what ***should be*** the correct output.

### DIFF
--- a/pkg/registry/rbac/validation/rule_test.go
+++ b/pkg/registry/rbac/validation/rule_test.go
@@ -26,7 +26,6 @@ import (
 
 	rbacv1 "k8s.io/api/rbac/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/util/diff"
 	"k8s.io/apiserver/pkg/authentication/user"
 )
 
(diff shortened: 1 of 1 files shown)


pkg/registry/rbac/validation/rule_test.go
```diff
@@ -144,7 +187,7 @@ func TestDefaultRuleResolver(t *testing.T) {
 			effectiveRules: nil,
 		},
 		{
-			StaticRoles: staticRoles1,
+			StaticRoles: staticRolesWithSystemTenant,
 			// Same as above but without a namespace. Only cluster rules should apply.
```
chenqianfzh, line 191: You mean "same as above but without a tenant", right?

pdgetrf: this means same as the cases right above it.

pdgetrf: will add some clarification.

pdgetrf: this comment is actually a repeat from previous cases. doesn't add much value so it's now removed.

PR #189, opened 2020-04-09
Arktos-up: conditional skip CA file generating + allow using master i…

Sindica:
Tested arktos-up.sh on Ubuntu 16.04 AWS EC2 host

### DIFF
--- a/hack/lib/common-var-init.sh
+++ b/hack/lib/common-var-init.sh
@@ -78,6 +78,7 @@ EVICTION_PRESSURE_TRANSITION_PERIOD=${EVICTION_PRESSURE_TRANSITION_PERIOD:-"1m"}
 
 # ensures all places of cert/node naming related to hostname are lower-cased for consistency
 lohostname=$(hostname | tr '[:upper:]' '[:lower:]')
+hostip=$(hostname -i)
 
 # This script uses docker0 (or whatever container bridge docker is currently using)
 # and we don't know the IP of the DNS pod to pass in as --cluster-dns.
@@ -161,6 +162,7 @@ API_SECURE_PORT=${API_SECURE_PORT:-6443}
 # WARNING: For DNS to work on most setups you should export API_HOST as the docker0 ip address,
 API_HOST=${API_HOST:-"${lohostname}"}
 API_HOST_IP=${API_HOST_IP:-"0.0.0.0"}
+API_HOST_IP_EXTERNAL=${API_HOST_IP_EXTERNAL:-"${hostip}"}
 ADVERTISE_ADDRESS=${ADVERTISE_ADDRESS:-""}
 NODE_PORT_RANGE=${NODE_PORT_RANGE:-""}
 API_BIND_ADDR=${API_BIND_ADDR:-"0.0.0.0"}
--- a/hack/lib/common.sh
+++ b/hack/lib/common.sh
@@ -209,21 +209,28 @@ function kube::common::set_service_accounts {
 
 function kube::common::generate_certs {
     # Create CA signers
-    if [[ "${ENABLE_SINGLE_CA_SIGNER:-}" = true ]]; then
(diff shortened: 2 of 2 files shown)


pdgetrf:
This is will not be tested through gates so please add info in the description about how this has been tested on any environment.

hack/lib/common-var-init.sh
```diff
@@ -78,6 +78,7 @@ EVICTION_PRESSURE_TRANSITION_PERIOD=${EVICTION_PRESSURE_TRANSITION_PERIOD:-"1m"}
 
 # ensures all places of cert/node naming related to hostname are lower-cased for consistency
 lohostname=$(hostname | tr '[:upper:]' '[:lower:]')
+hostip=$(hostname -i)
```
jshaofuturewei, line 81: Not sure if hostname -i works on MacBook.

Sindica: I don't think Arktos-up supports MacBook

Sindica:
> This is will not be tested through gates so please add info in the description about how this has been tested on any environment.

Added above.

hack/lib/common-var-init.sh
jshaofuturewei: I remember that someone ran arktos-up using his macbook and got some network interface issues. It will be ok if we don't support macbook.

Sindica: Sonya said she was never able to use macbook to build. Vinay's test also runs on Ubuntu only. Arktos is committed to support macbook.

hack/lib/common.sh
```diff
@@ -209,21 +209,28 @@ function kube::common::set_service_accounts {
...
+          sudo cp "${CERT_DIR}/server-ca.crt" "${CERT_DIR}/client-ca.crt"
+          sudo cp "${CERT_DIR}/server-ca-config.json" "${CERT_DIR}/client-ca-config.json"
+      else
+          kube::util::create_signing_certkey "${CONTROLPLANE_SUDO}" "${CERT_DIR}" server '"server auth"'
+          kube::util::create_signing_certkey "${CONTROLPLANE_SUDO}" "${CERT_DIR}" client '"client auth"'
+      fi
     else
-        kube::util::create_signing_certkey "${CONTROLPLANE_SUDO}" "${CERT_DIR}" server '"server auth"'
-        kube::util::create_signing_certkey "${CONTROLPLANE_SUDO}" "${CERT_DIR}" client '"client auth"'
+      echo "Skip generating CA as CA files existed and REGENERATE_CA == true. To regenerate CA files, export REGENERATE_CA=true"
```
pdgetrf, line 226: "REGENERATE_CA == false"?

hack/lib/common-var-init.sh
```diff
@@ -161,6 +162,7 @@ API_SECURE_PORT=${API_SECURE_PORT:-6443}
 # WARNING: For DNS to work on most setups you should export API_HOST as the docker0 ip address,
 API_HOST=${API_HOST:-"${lohostname}"}
 API_HOST_IP=${API_HOST_IP:-"0.0.0.0"}
+API_HOST_IP_EXTERNAL=${API_HOST_IP_EXTERNAL:-"${hostip}"}
```
pdgetrf, line 165: nit: seems we can move the definition of hostip from line 81 into line 165 because the variable hostip is only used here.  same to how line 163 works.

pdgetrf: also, consider adding some info explaining API_HOST_IP_EXTERNAL now that we have API_HOST, API_HOST_IP and API_HOST_IP_EXTERNAL. It can get a bit confusing.

In fact please add this to the PR description why this variable is introduced in the cert.

Sindica: Kept where lohostname is defined

hack/lib/common.sh
Sindica: Updated to REGENERATE_CA != true

PR #191, opened 2020-04-10
Backward client go

chenqianfzh:
For the convenience of reviews, this PR is split into two commits:
1. The change to the codegen template file, named "update type generator template", where only one file is involved. 
2. The files which are auto-updated by codegen based on the template changes. 101 files are involved.

Reviewers can browse through the files changes in the second commit quickly.

### DIFF
--- a/staging/src/k8s.io/code-generator/cmd/client-gen/generators/generator_for_type.go
+++ b/staging/src/k8s.io/code-generator/cmd/client-gen/generators/generator_for_type.go
@@ -566,10 +566,22 @@ var createSubresourceTemplate = `
 // Create takes the representation of a $.inputType|private$ and creates it.  Returns the server's representation of the $.resultType|private$, and an error, if there is any.
 func (c *$.type|privatePlural$) Create($.type|private$Name string, $.inputType|private$ *$.inputType|raw$) (result *$.resultType|raw$, err error) {
 	result = &$.resultType|raw${}
+	$if .tenanted$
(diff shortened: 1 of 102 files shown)


chenqianfzh:
PR for issue https://github.com/futurewei-cloud/arktos/issues/185

staging/src/k8s.io/code-generator/cmd/client-gen/generators/generator_for_type.go
```diff
@@ -566,10 +566,22 @@ var createSubresourceTemplate = `
...
+	$if .tenanted$
+	objectTenant := $.inputType|private$.ObjectMeta.Tenant 
+	if objectTenant == "" {
+		objectTenant = c.te
+	}
+	$end$	
+	$if .namespaced$
+	objectTenant := $.inputType|private$.ObjectMeta.Tenant 
+	if objectTenant == "" {
+		objectTenant = c.te
```
pdgetrf, line 578: what's c.te in this case?

Sindica:
Suggestion: the description of this review are too vague. It is hard to tell what it does later when there are a lot of changes.

staging/src/k8s.io/code-generator/cmd/client-gen/generators/generator_for_type.go
chenqianfzh: te is the tenant name, defined in type of $.typePlurals$. Together with the field of ns, it defines the space for search for the object to perform operation.

Issue #174, opened 2020-04-04
Workload controller manager needs to die sooner


Sindica:
**What would you like to be added**:
Controller instance shall be auto deleted within one minute after the instance died.
. Controller instance health report interval reduced from 1 minute to 10s (currently api server is doing it) - ideally configurable
. Controller instance object will be deleted from registry if there is no updates in past 1 minute (ideally configurable)
. It will be good to batch update different controller instances in the same workload CM so that ETCD will have less write. (optional)

**Why is this needed**:
Currently, controller instance record in registry will be automatically deleted after 5 minutes when it could not renew its lease. This is designed for a very stable environment. However, initial testing often has stability problem and caused workload controller manager restart. Further investigation also found api server does lease renewal every 10 seconds. This change will allow problematic controller instances to be discovered much faster.

PR #193, opened 2020-04-13
Workload controller manager needs to die sooner #174

Hong-Chang:
**What type of PR is this?**
/kind feature

Currently, controller instance record in registry will be automatically deleted after 5 minutes when it could not renew its lease. This is designed for a very stable environment. However, initial testing often has stability problem and caused workload controller manager restart. Further investigation also found api server does lease renewal every 10 seconds. This change will allow problematic controller instances to be discovered much faster.

Workload controller manager needs to die sooner #174

Please suggest configured values.
I introduced type ControllerConfig, so renamed original ControllerConfig to ControllerConfigMap.

No

### DIFF
--- a/cmd/workload-controller-manager/config/controllerconfig.json
+++ b/cmd/workload-controller-manager/config/controllerconfig.json
@@ -3,17 +3,17 @@
         {
             "type":    "node",
             "workers": 5,
-            "reportHealthIntervalInSecond":  10
+            "reportHealthIntervalInSecond":  20
         },
         {
             "type":    "replicaset",
             "workers": 10,
-            "reportHealthIntervalInSecond":  10
+            "reportHealthIntervalInSecond":  20
         },
         {
             "type":     "deployment",
             "workers":  5,
-            "reportHealthIntervalInSecond":  10
+            "reportHealthIntervalInSecond":  20
         }
     ]
 }
--- a/cmd/workload-controller-manager/config/controllerconfig.json
+++ b/cmd/workload-controller-manager/config/controllerconfig.json
@@ -2,15 +2,18 @@
     "controllers": [
         {
             "type":    "node",
-            "workers": 5
+            "workers": 5,
+            "reportHealthIntervalInSecond":  10
         },
         {
             "type":    "replicaset",
-            "workers": 10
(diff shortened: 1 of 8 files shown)


cmd/workload-controller-manager/config/controllerconfig.json
```diff
@@ -2,15 +2,18 @@
     "controllers": [
         {
             "type":    "node",
-            "workers": 5
+            "workers": 5,
+            "reportHealthIntervalInSecond":  20
```
Sindica, line 6: Heath report interval shall be consistent. Please keep a single value and don't duplicate necessary

pkg/cloudfabric-controller/deployment/deployment_controller.go
```diff
@@ -155,7 +156,7 @@ func NewDeploymentController(dInformer appsinformers.DeploymentInformer, rsInfor
 }
 
 // Run begins watching and syncing.
-func (dc *DeploymentController) Run(workers int, stopCh <-chan struct{}) {
+func (dc *DeploymentController) Run(controllerConfig config.ControllerConfig, stopCh <-chan struct{}) {
```
Sindica, line 159: Run() function has single signature as the original. Try to keep it the same. We are trying to make the migration of controllers from Kube CM to workload CM as easy as possible

Sindica:
+ joe for configuration change review

pkg/registry/core/controllerinstance/storage/storage.go
```diff
@@ -33,7 +33,7 @@ type REST struct {
 	*genericregistry.Store
 }
 
-const TTL = 300 // time-to-live in seconds
+const TTL = 60 // time-to-live in seconds
```
Sindica, line 36: Can we change this to 30 seconds?

cmd/workload-controller-manager/config/controllerconfig.json
Sindica: Perhaps do 10 second per report as API server renew lease every 10 seconds.

Hong-Chang:
Created another pr to replace this one.

PR #197, opened 2020-04-14
[Workload controller manager needs to die sooner #174]

Hong-Chang:
What type of PR is this?
/kind feature

Currently, controller instance record in registry will be automatically deleted after 5 minutes when it could not renew its lease. This is designed for a very stable environment. However, initial testing often has stability problem and caused workload controller manager restart. Further investigation also found api server does lease renewal every 10 seconds. This change will allow problematic controller instances to be discovered much faster.

Which issue(s) this PR fixes:
Workload controller manager needs to die sooner #174

Please suggest configured values.
I introduced type ControllerConfig, so renamed original ControllerConfig to ControllerConfigMap.

No


cmd/workload-controller-manager/app/config/controllerconfig.go
```diff
@@ -23,7 +23,8 @@ import (
 )
 
 type controllerTypes struct {
```
Sindica, line 25: Naming is not fit.

Sindica:
Joe initially authored controller config. Adding him for review.

cmd/workload-controller-manager/config/controllerconfig.json
```diff
@@ -1,4 +1,5 @@
 {
+    "reportHealthIntervalInSecond": 10,
```
jshaofuturewei, line 2: The config here looks like the controllers of all the types share the same reportHealthIntervalInSecond but in the codes each controller type deals with reportHealthIntervalInSecond  seperately.

cmd/workload-controller-manager/app/config/controllerconfig.go
Hong-Chang: Renamed type to controllerManagerConfig

cmd/workload-controller-manager/config/controllerconfig.json
Hong-Chang: Yes, reportHealthIntervalInSecond has one value for all types. Its usage is when starting each controller so appears to be seperately. I think it's ok.

PR #204, opened 2020-04-14
fix mistakenly reversed namespace and tenant in parameter

pdgetrf:
g.Lister.RolesWithMultiTenancy(namespace string, tenant string). This was checked in last October. 

The new code added recently has tenant before namespace instead. 

This is causing rbac to wrongly deny access requests.

Tested with demo setup on aws machines.


pdgetrf:
> We need to add more tests. Let's not forget this.

tracking issue created

PR #206, opened 2020-04-16
design doc: service in multi-tenancy networking

h-w-chen:
**What type of PR is this?**
/kind design
/kind documentation

1. elaborates services in multi-tenant networkings, in particular kubernetes & kube-dns;
2. adds a few architectural views;

None


docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -120,10 +128,118 @@ The reasons that DNS service is per-network instead of per-space are:
 * VPC networks are isolated by nature. It will be difficult to have a shared DNS pod that's accessible to pods in various VPC networks.
 
 ### Service
-For VPC-isolated networking, service IP support has to be provided w/o relying on iptable/IPVS rules on host netns. In other words, traditional kube-proxy can not used in VPC-isolated networking.
 
+#### Type definition
+
+Addition of service spec is metadata.network field. By default, it is the default network of the tenant.
```
XiaoningDing, line 134: "metadata.network" in text also needs to be updated.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -120,10 +128,118 @@ The reasons that DNS service is per-network instead of per-space are:
...
+    metadata:
+      name: default
+    spec:
+      type: flat
+      service:
+      - cidr: 10.0.0.0/16
+      - cidr: 192.168.0.0/24
+    ```
+   
+1. on service creation, some mechanism to assign service the proper IP address.
```
XiaoningDing, line 197: We can still reuse the existing service IP assignment mechanism in API Server.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -120,10 +128,118 @@ The reasons that DNS service is per-network instead of per-space are:
...
+    ```yaml
+    apiVersion: v1
+    kind: Network
+    metadata:
+      name: default
+    spec:
+      type: flat
+      service:
+        ipam: external
+      - cidr: 10.0.0.0/16
```
XiaoningDing, line 213: Should this be removed?

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -120,10 +128,118 @@ The reasons that DNS service is per-network instead of per-space are:
...
+
+    ```yaml
+    apiVersion: v1
+    kind: Network
+    metadata:
+      name: default
+    spec:
+      type: flat
+      service:
+        ipam: external
```
zasherif, line 212: How can an external provider learn the associated network object? For example, how to determine that this service belongs to a specific VPC (tenant)?

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -120,10 +128,118 @@ The reasons that DNS service is per-network instead of per-space are:
...
+    ```yaml
+    apiVersion: v1
+    kind: Network
+    metadata:
+      name: default
+    spec:
+      type: flat
+      service:
+      - cidr: 10.0.0.0/16
+      - cidr: 192.168.0.0/24
```
zasherif, line 194: I am not sure I fully understand the role of multiple CIDR entries here. Would you please explain in the text?

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
zasherif, line 197: Would you please provide more details here? For example, where is the association with a tenant network defined?

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -120,10 +128,118 @@ The reasons that DNS service is per-network instead of per-space are:
...
+    ```
+   
+1. on service creation, some mechanism to assign service the proper IP address.
+
+##### Network provider managed pool
+Service IP addresses are implicitly managed by the capable network provider. Network object can have the service IP cidr (or not); at best for information - Arktos delegates IPAM to the external network provider, via network controller invoking the service it provides. 
+
+Network object specifies service ipam as *external* (the default is k8s - managed by Arktos):
```
zasherif, line 202: What are your thoughts on using annotations to define the service network provider?  So we extend the Internal loadbalancer pattern defined here https://kubernetes.io/docs/concepts/services-networking/service/#publishing-services-service-types

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: yes I'll correct it.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -120,10 +128,118 @@ The reasons that DNS service is per-network instead of per-space are:
...
+      type: flat
+      service:
+      - cidr: 10.0.0.0/16
+      - cidr: 192.168.0.0/24
+    ```
+   
+1. on service creation, some mechanism to assign service the proper IP address.
+
+##### Network provider managed pool
+Service IP addresses are implicitly managed by the capable network provider. Network object can have the service IP cidr (or not); at best for information - Arktos delegates IPAM to the external network provider, via network controller invoking the service it provides.
```
zasherif, line 200: In case of an external network provider, is there a specific mutation expected to the service object? For example, in Mizar, we would allocate an IP address and will need to mutate the service object with the IP value. Will Arktos permits mutation in this case or provides another mechanism to propagate this info?

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -120,10 +128,118 @@ The reasons that DNS service is per-network instead of per-space are:
...
+```
+
+#### Naming Problem & Solution
+
+Services, as one of the critical network related resource types, are withing the network boundary. There is no such *shared* services of tenant or whole cluster, as both the critical virtual IP address and the associated endpoints (typically leading to pods) are all of same network.
+
+Putting service type in tenant level is out of primacy as it has following flaws:
+1. service IP address has to be the same across all networks, which implies some range of service IP needs to be predefined as shared across networks of a tenant (tenant manageability burden);  
+1. associating network-agnostic service to network-specific endpoints would introduce significant inconsistency;
+1. kubectl query for service (e.g. ```kubectl describe service```) has no way to return meaningful content of endpoints.
```
zasherif, line 158: How about Kubernetes' endpoints, how the object details will change?

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: yes if only it can be decided by arktos; if the service IP address should be decided by the external network provider, it would undergo some significant change. I'll update the doc to elaborate such.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: yes better to remove to avoid confusion.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -120,10 +128,118 @@ The reasons that DNS service is per-network instead of per-space are:
...
+
+The well-known services (default/kubernetes, kube-system/kube-dns) present some problem here. Service kubernetes has tenant/namespace(default) scope in current implementation, unable to have multiple kubernets services with each of them in one network due to name conflict. The same rational applies to kube-dns.
```
zasherif, line 160: I don't fully understand the problem here.
1. What is the scope in current implementation?
1. Why placing services in namespaces is not sufficient?

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -120,10 +128,118 @@ The reasons that DNS service is per-network instead of per-space are:
...
+
+Another solution is to change their name to schema that can uniquely differentiate by combining with network name; this is awkward and breaking in general. 
+
+One plausible remedy is making stringent containment of network to namespaces - resources in one namespace belongs solely to one network; one network can accommodate multiple namespace. This cannot handle the default/kubernetes service problem, as it is of default namespace.
+
+None of the two of above seems ideal. However, the latter, fiddling with the service name, if coupled with trick applicable only to kubernetes & kube-dns, seems pragmatic and practical. The idea is suffixing service name with network name (e.g. kubernetes_mynetwork), also answering query for kubernetes with that of kubernetes_mynetwork (by alias record or other form of dynamic technique), so that pods are able to look up them without using the network-suffixed names. This approach is not flexible and depending on dns ability; it works fine with limited number of well-known aliases, though. coreDNS, one of the most commonly used dns server binaries, has such dynamic aliasing ability.
```
zasherif, line 168: If I understand this correctly the suffix issue, leaks service names across tenants?

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: In that case, a controller (could be service controller, or a supplementary service IPAM controller) is responsible to notify the external network provider of network/service. As the controller is part of arktos runtime, it knows the needed info and able to pass them to the external network provider. The fully-detailed design is not decided yet; it would be elaborated in the controller design.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: multiple CIDR entries just gives tenant admin a bit more option, in case one range has insufficient entries. This helps with network update - admin wants to extend the range after it has been created.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: The statement in doc seems vague - sorry for the confusion it has caused. 
I'll update this sentence to make it more clear. In this case of arktos managed pool, when a service is created and having its IP address assigned, arktos will reuse the assignment logic to allocate one free address form the service CIDR range of the network of that service -- not much change from current arktos except for the network range instead of whole cluster range.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: Each service object must to have associated network. This design puts it in spec.network field. I think it should serve to the similar effect of annotation-of-network. If this is not the case, we can dig a bit more.

Multiple network providers is indeed something we should consider. Thanks for bringing up this concern.  I will add a TBD item as reminder of network type consideration.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: This is the detailed interaction spec between arktos controller and network provider. It is very important for the proper integration. What you described is very critical; we need to discuss more definitely - in separate meetings.

Whether the service ip address mutable or  not in arktos, I am not very clean - I'll find it out. If we can keep the service IP address inmutable, it would be ideal. 

I'll put a note in doc about this service mutation issue.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: endpoints will be in later design - we expect the minimum change of type (with addition of spec.network), hopefully it can keep the change impact small.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: This is the very tricky situation. Taking the "kubernetes" service: in multi-tenancy arktos, each tenant would have its own kubernetes service in the default namespace; we cannot put another same-named "kubernetes" service in the same "deafult" namespace - kubernetes apiserver just won't allow it. What we need is EACH network should has one "kubernetes" service, scoped in "default" namespace, and serves for the specific network only.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: yes as the name is well-known and fixed, objects (e.g. pods) of each tenant knows of (and actually relies on) these fixed-named service. -- name leakage, in some sense - for good reason.
However, each tenant is strongly isolated; it is impossible for one in a tenant to get the content of such service of another tenant - it can only get hold of the one of its tenant.

PR #208, opened 2020-04-17
Enable multi-tenancy in job controller

pdgetrf:
Part of Issue https://github.com/futurewei-cloud/arktos/issues/186

Enable multi-tenancy in job controller. 

Job is created using the tenant info provided by request.

Tested with both added unit tests and local E2E.

### DIFF
--- a/pkg/controller/job/job_controller.go
+++ b/pkg/controller/job/job_controller.go
@@ -176,13 +176,13 @@ func (jm *JobController) getPodJobs(pod *v1.Pod) []*batch.Job {
 // resolveControllerRef returns the controller referenced by a ControllerRef,
 // or nil if the ControllerRef could not be resolved to a matching controller
 // of the correct Kind.
-func (jm *JobController) resolveControllerRef(namespace string, controllerRef *metav1.OwnerReference) *batch.Job {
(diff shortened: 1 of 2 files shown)


pkg/controller/job/job_controller.go
```diff
@@ -440,14 +440,14 @@ func (jm *JobController) syncJob(key string) (bool, error) {
 		klog.V(4).Infof("Finished syncing job %q (%v)", key, time.Since(startTime))
 	}()
 
-	ns, name, err := cache.SplitMetaNamespaceKey(key)
+	tenant, namespace, name, err := cache.SplitMetaTenantNamespaceKey(key)
 	if err != nil {
 		return false, err
 	}
-	if len(ns) == 0 || len(name) == 0 {
+	if len(tenant) == 0 || len(namespace) == 0 || len(name) == 0 {
```
XiaoningDing, line 447: I remember tenant could be empty if it's the resource is in system space.  Not sure if it's still the case. If yes, then the check len(tenant)==0 will be a problem.   (ideally obj.tenant should be equal to "system" instead of empty.)

chenqianfzh: Job is a namespace-scoped object. For a namespace-scoped object, its tenant and namespace values in the ObjectMeta cannot be empty.

PR #209, opened 2020-04-17
Fixed the controller manager crash issue due to the leader election loss

jshaofuturewei:
If we add over 5000 hollow nodes to kubemark, it will cause kube controller manager failure due to the election loss.


staging/src/k8s.io/client-go/tools/leaderelection/leaderelection.go
```diff
@@ -338,9 +338,10 @@ func (le *LeaderElector) tryAcquireOrRenew() bool {
 		le.observedRecord = *oldLeaderElectionRecord
 		le.observedTime = le.clock.Now()
 	}
+
+	// No need to check if has the release now since the statement above adds time automatically
```
Sindica, line 342: typo: release -> lease

PR #215, opened 2020-04-21
fix cluster role getter in rbac for multitenancy

pdgetrf:
fix an issue where clusterrolebinding was unable to retrieve cluster role for a tenant.

### DIFF
--- a/pkg/registry/rbac/validation/rule.go
+++ b/pkg/registry/rbac/validation/rule.go
@@ -415,7 +415,7 @@ func (r *StaticRoles) GetClusterRoleWithMultiTenancy(tenant, name string) (*rbac
 			return clusterRole, nil
 		}
 	}
-	return nil, errors.New("clusterrole not found")
+	return nil, errors.New("clusterrole " + tenant + "/" + name + " not found")
 }
 
 func (r *StaticRoles) ListRoleBindings(namespace string) ([]*rbacv1.RoleBinding, error) {
--- a/pkg/registry/rbac/validation/rule.go
+++ b/pkg/registry/rbac/validation/rule.go
@@ -112,6 +112,7 @@ type RoleBindingLister interface {
 
 type ClusterRoleGetter interface {
 	GetClusterRole(name string) (*rbacv1.ClusterRole, error)
+	GetClusterRoleWithMultiTenancy(tenant, name string) (*rbacv1.ClusterRole, error)
 }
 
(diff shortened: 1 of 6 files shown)


pkg/registry/rbac/validation/rule.go
```diff
@@ -408,6 +409,15 @@ func (r *StaticRoles) GetClusterRole(name string) (*rbacv1.ClusterRole, error) {
 	return nil, errors.New("clusterrole not found")
 }
 
+func (r *StaticRoles) GetClusterRoleWithMultiTenancy(tenant, name string) (*rbacv1.ClusterRole, error) {
+	for _, clusterRole := range r.clusterRoles {
+		if clusterRole.Name == name && clusterRole.Tenant == tenant{
+			return clusterRole, nil
+		}
+	}
+	return nil, errors.New("clusterrole not found")
```
chenqianfzh, line 418: Minor: better to include the clusterrole name and tenant name in the error message.

plugin/pkg/auth/authorizer/rbac/subject_locator_test.go
```diff
@@ -115,7 +124,7 @@ func TestSubjectLocator(t *testing.T) {
 				},
 				{
 					// verb matchies correctly
```
chenqianfzh, line 126: typo: matchies

pkg/registry/rbac/validation/rule.go
pdgetrf: fixed

plugin/pkg/auth/authorizer/rbac/subject_locator_test.go
pdgetrf: fixed

plugin/pkg/auth/authorizer/rbac/rbac_test.go
```diff
@@ -44,18 +46,18 @@ func newRole(name, namespace string, rules ...rbacv1.PolicyRule) *rbacv1.Role {
 	return &rbacv1.Role{ObjectMeta: metav1.ObjectMeta{Namespace: namespace, Name: name}, Rules: rules}
```
chenqianfzh, line 46: Role is namespace-scoped, so it "Tenant" should be set.

pdgetrf: fixed

PR #220, opened 2020-04-21
make all event sinks support multi-tenancy

chenqianfzh:
This PR fixes the issue reported in https://github.com/futurewei-cloud/arktos/issues/212.


chenqianfzh:
> LGTM.
> 
> This is a temporary fix to prevent blocking other work right?

No. It is a true fix. With this fix, the events from any tenant can be recorded.

PR #224, opened 2020-04-23
Change crd scope

chenqianfzh:
This PR changed scope of CRD from cluster-scope to tenant-scope, so each tenant can have their own CRDs.

For the convenience of reviewers, the PR is split into three commits:
1. the code change to change the scope
2. the update in test files.
3. the auto updated files due to the scope change. This commit includes a big number of line changes, but the reviewer can just skip it.

### DIFF
--- a/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/validation/validation.go
+++ b/staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/validation/validation.go
@@ -59,7 +59,7 @@ func ValidateCustomResourceDefinition(obj *apiextensions.CustomResourceDefinitio
 		return ret
 	}
 
-	allErrs := genericvalidation.ValidateObjectMeta(&obj.ObjectMeta, false, false, nameValidationFn, field.NewPath("metadata"))
+	allErrs := genericvalidation.ValidateObjectMeta(&obj.ObjectMeta, true, false, nameValidationFn, field.NewPath("metadata"))
 	allErrs = append(allErrs, ValidateCustomResourceDefinitionSpec(&obj.Spec, field.NewPath("spec"))...)
 	allErrs = append(allErrs, ValidateCustomResourceDefinitionStatus(&obj.Status, field.NewPath("status"))...)
 	allErrs = append(allErrs, ValidateCustomResourceDefinitionStoredVersions(obj.Status.StoredVersions, obj.Spec.Versions, field.NewPath("status").Child("storedVersions"))...)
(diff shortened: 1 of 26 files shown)


staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/validation/validation.go
```diff
@@ -59,7 +59,7 @@ func ValidateCustomResourceDefinition(obj *apiextensions.CustomResourceDefinitio
 		return ret
 	}
 
-	allErrs := genericvalidation.ValidateObjectMeta(&obj.ObjectMeta, false, false, nameValidationFn, field.NewPath("metadata"))
+	allErrs := genericvalidation.ValidateObjectMeta(&obj.ObjectMeta, true, false, nameValidationFn, field.NewPath("metadata"))
```
pdgetrf, line 62: what does this true/false do?

staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/validation/validation_test.go
```diff
@@ -85,7 +85,7 @@ func TestValidateCustomResourceDefinition(t *testing.T) {
 		{
 			name: "webhookconfig: invalid port 0",
 			resource: &apiextensions.CustomResourceDefinition{
-				ObjectMeta: metav1.ObjectMeta{Name: "plural.group.com"},
+				ObjectMeta: metav1.ObjectMeta{Name: "plural.group.com", Tenant: "test-te"},
```
pdgetrf, line 88: nit: extract "test-te" into a variable

pdgetrf: and do we need a case for system tenant

pdgetrf:
thanks for splitting up the PR into 3 commits. It's made it a lot easier to review.

staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/validation/validation.go
chenqianfzh: This validation validates the object metadata. For a tenant-scoped object, its tenant field cannot be empty, and vice versa. Since we are changing the scope of CRD, we are changing the logic to validate it too.

staging/src/k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/validation/validation_test.go
chenqianfzh: As suggested, I extracted "test-te" into a string variable.

I don't think we need test cases for system tenant here. These tests are validating the  CRD-related structures and schemas, which are same for regular tenants and system tenants.

PR #234, opened 2020-04-25
Unit test for APIServerConfigManager & updater

Sindica:


docs/setup-guide/arktos-apiserver-partition.md
```diff
@@ -36,89 +44,36 @@ $ curl http://<1st master server ip>:2379/v2/keys
 # {"action":"get","node":{"dir":true}}
 ```
 
-. Start 2nd Api server (in a different host)
+. Copy all files from 1st API Server /var/run/kubernetes to host where you planed to run api server
```
jshaofuturewei, line 47: do we need to copy binary files as well?

staging/src/k8s.io/client-go/datapartition/apiserverconfigmanager_test.go
```diff
@@ -0,0 +1,214 @@
...
+		newInstance := GetAPIServerConfigManagerMock()
+		assert.NotNil(t, newInstance)
+
+		instance2 := GetAPIServerConfigManagerMock()
+		assert.NotNil(t, instance2)
+
+		assert.Equal(t, newInstance, instance2)
+
+		assert.Nil(t, instance.kubeClient)
+		instance = nil
```
jshaofuturewei, line 90: instance is already nil here. Do we miss something?

PR #242, opened 2020-04-28
TenantDefault to TenantSystem

chenqianfzh:
This PR move everything under TenantDefault to TenantSystem, to align with the design that everything before Arktos multi-tenancy belong to the world of the system tenant.


staging/src/k8s.io/apiextensions-apiserver/examples/client-go/pkg/client/clientset/versioned/typed/cr/v1/cr_client.go
```diff
@@ -43,7 +43,7 @@ type CrV1Client struct {
 }
 
 func (c *CrV1Client) Examples(namespace string) ExampleInterface {
-	return newExamplesWithMultiTenancy(c, namespace, "default")
+	return newExamplesWithMultiTenancy(c, namespace, "system")
```
Sindica, line 46: Can you use constant instead of string?

chenqianfzh: This is a file generated by codegen. It is hard to use constant like metav1.TenantSystem in the generated files, as codegen does not support adding dependency package for a constant.  codegen will add the dependent package for  a func/variable/type, but not for a const.

plugin/pkg/auth/authorizer/rbac/rbac.go
```diff
@@ -77,7 +77,7 @@ func (r *RBACAuthorizer) Authorize(requestAttributes authorizer.Attributes) (aut
 	ruleCheckingVisitor := &authorizingVisitor{requestAttributes: requestAttributes}
 
 	userTenant := requestAttributes.GetUser().GetTenant()
-	if userTenant != metav1.TenantSystem && userTenant != metav1.TenantDefault && userTenant != requestAttributes.GetTenant() {
+	if userTenant != metav1.TenantSystem && userTenant != metav1.TenantSystem && userTenant != requestAttributes.GetTenant() {
```
pdgetrf, line 80: duplicated check for TenantSystem

staging/src/k8s.io/api/core/v1/types.go
```diff
@@ -25,12 +25,10 @@ import (
 )
 
 const (
-	// TenantDefault means the object is in the default tenant which is applied when not specified by clients
-	TenantDefault string = "default"
 	// TenantAll is the default argument to specify on a context when you want to list or filter resources across all tenants
 	TenantAll string = ""
-	// TenantNodeLease is the tenant where we place node lease objects (used for node heartbeats)
-	TenantNodeLease string = "system"
+	// TenantSystem is the tenant where the cluster operators belong to
```
pdgetrf, line 30: double check. is this the correct definition for cluster operators?

Sindica:
Please don't merge now. I will need to test api server partition on this change

Sindica:
> Tested in my local environment. Currently, the second api server cannot be properly started with this change. It shows node not ready in second api server.
> Please use the following doc as set up guideline for data partitioned cluster.
> https://github.com/futurewei-cloud/arktos/blob/master/docs/setup-guide/arktos-apiserver-partition.md

I ran my test again twice. It seems working now. Can you also performance the test on your host? Thank you

plugin/pkg/auth/authorizer/rbac/rbac.go
chenqianfzh: fixed.

staging/src/k8s.io/api/core/v1/types.go
chenqianfzh: updated

PR #246, opened 2020-04-29
update title/version in swagger doc

chenqianfzh:
Update the title/version in swagger.json to reflect the fact that we are now at Arktos v.0.2, instead of Kubernetes v1.15.0.

This change will get us out of the pain to manually update the swagger.json every time after running "make update".

### DIFF
--- a/cmd/kube-apiserver/app/server.go
+++ b/cmd/kube-apiserver/app/server.go
@@ -24,7 +24,6 @@ import (
 	"crypto/tls"
 	"fmt"
 	"io/ioutil"
-	"k8s.io/apiserver/pkg/storage/datapartition"
 	"net"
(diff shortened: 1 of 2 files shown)


jshaofuturewei:
Discussed with Qian that There are lots of kubernetes titles and version in our current repo remained unchanged. The versions are used to see if they support other components and plugins or build images. It might be good to create an issue to track the title and version changes.

chenqianfzh:
> There are lots of kubernetes titles and version in our current repo remained unchanged. The versions are used to see if they support other components and plugins or build images. It might be good to create an issue to track the title and version changes.

As discussed, I created issue: https://github.com/futurewei-cloud/arktos/issues/274

Issue #111, opened 2020-03-10
Decide the right API group and version for network objects


XiaoningDing:
**What would you like to be added**:
Decide the right API group and version for the newly proposed network object. Or define it as CRDs for easier adopted by other projects.

**Why is this needed**:

h-w-chen:
as #247 is coded, 'arktos.futurewei.com/v1' is the GV for this network CR.

h-w-chen:
network is defined as CRD, in arktos.futureeoi.com/v1

PR #247, opened 2020-04-30
multi-tenancy networking: defines CRD of network

h-w-chen:
**What type of PR is this?**
/kind api-change
/kind feature

**What this PR does / why we need it:**
Network API type is the cornerstone of multi-tenant networking. See [multi-tenancy networking design](https://github.com/futurewei-cloud/arktos/blob/master/docs/design-proposals/multi-tenancy/multi-tenancy-network.md) for more design details.

In order to harden the network isolation in multi-tenancy cluster, a new custom API type, network, is needed to provide the isolation boundary. Pods cannot access network resources across networks; each tenant typically has different networks (exception is "flat" network, which is provided for backward compatibilities to leverage existing kubernetes network offerings), and one tenant may have multiple networks.

Defining network type in CRD imposes minimum changes to system by leveraging the built-in extension ability.

**Fixes or implements**
#111 

**Does this PR introduce a user-facing change?:**
YES

```release-note
Introduced a new custom API type: networks.arktos.futurewei.com/v1 Network.
$ kubectl apply -f pkg/controller/artifacts/crd-network.yaml
$ kubectl get crd
will return networks.arktos.futurewei.com, the new type
$  kubectl api-resources --api-group arktos.futurewei.com
NAME       SHORTNAMES   APIGROUP               TENANTED   NAMESPACED   KIND
networks   ne,net       arktos.futurewei.com   true       false        Network
```


pkg/controller/artifacts/crd-network.yaml
```diff
@@ -0,0 +1,50 @@
...
+  validation:
+    openAPIV3Schema:
+      type: object
+      properties:
+        spec:
+          type: object
+          properties:
+            type:
+              type: string
+              enum: [flat, vpc]
```
XiaoningDing, line 30: Should this be any string?

pkg/controller/artifacts/crd-network.yaml
```diff
@@ -0,0 +1,50 @@
...
+            vpcID:
+              type: string
+        status:
+          type: object
+          properties:
+            phase:
+              type: string
+              enum: [Pending, OK, Failed, Unknown, Terminating]
```
XiaoningDing, line 38: "OK" to "Ready"?

pkg/controller/artifacts/README.md
```diff
@@ -0,0 +1,22 @@
+Network CRD & related controllers
+=================================
+
+## Network CRD
+
+Network CRD is the custom API type to provide hard network isolation. Please see [multi-tenancy network design proposal](../../../docs/design-proposals/multi-tenancy/multi-tenancy-network.md). 
+
+The CRD yaml file, crd-network.yaml, __MUST__ be applied __BEFORE__ any controller runs, in cluster provision process.
```
XiaoningDing, line 8: This is a little unusual for a CRD object.  Besides this, there are some other potential problems of the CRD approach are: 1) Native objects like pod/service will reference this CRD object. 2) this CRD needs to be shared among all tenants, but our multi-tenancy doesn't support sharing CRD cross tenants yet.

In general, I still don't know which approach is better -- CRD or a native object.  Let's compare these the pros and cons and choose one PR :)

pkg/controller/artifacts/crd-network.yaml
h-w-chen: dropped the limitation.

pkg/controller/artifacts/crd-network.yaml
h-w-chen: corrected as suggested.

pkg/controller/artifacts/README.md
XiaoningDing: We had a meeting to discuss this. The decision is that: 1)  we will go with the CRD approach. 2) Multi-tenancy will provide CRD sharing feature. 3) Our network operator needs to be tenancy-aware and it will only be deployed in system space.

h-w-chen: per offline discussion, CRD approach was agreed upon.

PR #249, opened 2020-04-30
fix/enable apiextension integration tests

chenqianfzh:
When working on multi-tenancy CRD, I found that the apiextension integration tests are not part of CI/CD. 

This PR does the following:
1. fix the apiextension integration tests caused by:
       a. the API changes in Arktos code change.
       b. the mistake in existing test cases that etcd server is not started.
2. Add the apiextension integration tests to CI/CD

This PR does not introduce any new test cases, yet it will pave the way to create more CRD-related integration test cases in Arktos.


staging/src/k8s.io/apiextensions-apiserver/test/integration/conversion/main_test.go
```diff
@@ -0,0 +1,27 @@
+/*
+Copyright 2017 The Kubernetes Authors.
```
chenqianfzh, line 2: Will change it.

staging/src/k8s.io/apiextensions-apiserver/test/integration/main_test.go
```diff
@@ -0,0 +1,30 @@
+/*
+Copyright 2017 The Kubernetes Authors.
```
chenqianfzh, line 2: Will change it.

PR #258, opened 2020-05-04
Use standard error checking in List and Watch

Sindica:


staging/src/k8s.io/client-go/kubernetes/typed/core/v1/secret.go
```diff
@@ -157,7 +157,7 @@ func (c *secrets) Watch(opts metav1.ListOptions) watch.AggregatedWatchInterface
 			VersionedParams(&opts, scheme.ParameterCodec).
 			Timeout(timeout).
 			Watch()
-		if err != nil && opts.AllowPartialWatch {
+		if err != nil && opts.AllowPartialWatch && errors.IsForbidden(err) {
 			// watch error was not returned properly in error message. Skip when partial watch is allowed
 			klog.V(6).Infof("Watch error for partial watch %v. options [%+v]", err, opts)
```
chenqianfzh, line 162: Is it the expected behavior? An error is logged as Info with a very low level of 6.

staging/src/k8s.io/code-generator/cmd/client-gen/generators/generator_for_type.go
```diff
@@ -492,7 +492,7 @@ func (c *$.type|privatePlural$) List(opts $.ListOptions|raw$) (result *$.resultT
 			return
 		}
 
-		if err != nil && $.stringsContains|raw$(err.Error(), "forbidden") &&
+		if err != nil && $.errorForbidden|raw$(err) &&
 			$.stringsContains|raw$(err.Error(), "no relationship found between node") {
 			$.klogInfof|raw$(6).Infof("Skip error %v in list", err)
```
chenqianfzh, line 497: Add more info about the resources to list in the log message.

staging/src/k8s.io/code-generator/cmd/client-gen/generators/generator_for_type.go
```diff
@@ -771,7 +771,7 @@ func (c *$.type|privatePlural$) Watch(opts $.ListOptions|raw$) $.aggregatedWatch
 			VersionedParams(&opts, $.schemeParameterCodec|raw$).
 			Timeout(timeout).
 			Watch()
-		if err != nil && opts.AllowPartialWatch {
+		if err != nil && opts.AllowPartialWatch && $.errorForbidden|raw$(err) {
 			// watch error was not returned properly in error message. Skip when partial watch is allowed
 			$.klogInfof|raw$(6).Infof("Watch error for partial watch %v. options [%+v]", err, opts)
```
chenqianfzh, line 776: Is it the expected behavior? An error is logged as Info with a very low level of 6.

Besides, the log info is not clear. Add more info about the resources,  tenant/space/type,  in the log message.

staging/src/k8s.io/code-generator/cmd/client-gen/generators/generator_for_type.go
```diff
@@ -142,6 +142,7 @@ func (g *genClientForType) GenerateType(c *generator.Context, t *types.Type, w i
 		"GroupGoName":              g.groupGoName,
 		"Version":                  namer.IC(g.version),
 		"klogInfof":                c.Universe.Type(types.Name{Package: "k8s.io/klog", Name: "V"}),
+		"errorForbidden":           c.Universe.Function(types.Name{Package: "k8s.io/apimachinery/pkg/api/errors", Name: "IsForbidden"}),
```
chenqianfzh, line 145: change errorForbidden -> errorsIsForbidden, more readable and consistent with the naming style of other keys in this map.

staging/src/k8s.io/code-generator/cmd/client-gen/generators/generator_for_type.go
```diff
@@ -467,8 +468,7 @@ func (c *$.type|privatePlural$) List(opts $.ListOptions|raw$) (result *$.resultT
 		return
 	}
 
-	if !$.stringsContains|raw$(err.Error(), "forbidden") ||
-		!$.stringsContains|raw$(err.Error(), "no relationship found between node") {
+	if !($.errorForbidden|raw$(err) && $.stringsContains|raw$(err.Error(), "no relationship found between node")) {
```
chenqianfzh, line 471: It would help code reading to add a comment that node authorizer will issue an error like "no relationship found between node ...". At first I thought there was a typo as the sentence "no relationship found between node" is incorrect in grammar. ( I had thought it was meant to be "no relationship found between nodeS")

Sindica: The error message is defined in authorization plug in, used by both api server and kubelet. This update is just using part of the error message.

staging/src/k8s.io/code-generator/cmd/client-gen/generators/generator_for_type.go
Sindica: This is generic list 

> I like this change.
> 
> Would like to see some similar changes to have a clean way to check whether the error contains the string ""no relationship found between node".

The error message is in authorization plug in. You might want to ask Peng Du to make the changes.

staging/src/k8s.io/code-generator/cmd/client-gen/generators/generator_for_type.go
Sindica: This log is intentionally used in level 6, which cannot be used in level 6. This is a valid case and thus not an error. Also, the code is generated by code-gen, it is used in cases that might not have tenant and namespace, thus it is complicated to add log for tenant/namespace. I don't think it is worth to add due to code complexity and usage.

staging/src/k8s.io/code-generator/cmd/client-gen/generators/generator_for_type.go
Sindica: Updated

PR #261, opened 2020-05-05
Updated single bracket to double brackets

jshaofuturewei:
Double brackets implicitly quote all arguments. The APISERVERS_EXTRA is not quoted with single bracket. Updated to double brackets

### DIFF
--- a/hack/lib/common-var-init.sh
+++ b/hack/lib/common-var-init.sh
@@ -16,7 +16,7 @@
 # Start of Common environment variables used in arktos-up.sh& arktos-apiserver-partition.sh
 # --------------------------------------------------------------------------------------------
 APISERVERS_EXTRA=${APISERVERS_EXTRA:-""}
-if [ ${APISERVERS_EXTRA} != "" ]; then
+if [[ ${APISERVERS_EXTRA} != "" ]]; then
     APISERVERS_EXTRA=${APISERVERS_EXTRA/:/\",\"}
 fi
 


Sindica:
> Double brackets implicitly quote all arguments. The APISERVERS_EXTRA is not quoted with single bracket. Updated to double brackets

Not sure what you are trying to accomplish here.

https://stackoverflow.com/questions/2188199/how-to-use-double-or-single-brackets-parentheses-curly-braces
http://mywiki.wooledge.org/BashFAQ/031

Based on my test, [[]] runs faster than []. I am ok to replace [] with [[]]. But does it break anything?

jshaofuturewei:
Every time when I create a new repo and run hack/arktos-up.sh, I got errors telling APISERVERS_EXTRA is not a string

[[ ${APISERVERS_EXTRA} != "" ]] equals [ "${APISERVERS_EXTRA}" != "" ] which solve the issue.

> > Double brackets implicitly quote all arguments. The APISERVERS_EXTRA is not quoted with single bracket. Updated to double brackets
> 
> Not sure what you are trying to accomplish here.
> 
> https://stackoverflow.com/questions/2188199/how-to-use-double-or-single-brackets-parentheses-curly-braces
> http://mywiki.wooledge.org/BashFAQ/031
> 
> Based on my test, [[]] runs faster than []. I am ok to replace [] with [[]]. But does it break anything?

Issue #186, opened 2020-04-08
Enable multi-tenancy in controllers


XiaoningDing:
<To add details>

pdgetrf:
![image](https://user-images.githubusercontent.com/252020/79301417-6bfe9380-7e9e-11ea-8952-a48286077151.png)

deployment controller is working with tenant.

pdgetrf:
job controller is not working with multi-tenancy. need code change. 

![image](https://user-images.githubusercontent.com/252020/79301515-aa944e00-7e9e-11ea-8171-9f2d8beb4e0c.png)

pdgetrf:
daemon set needs multi-tenancy change

pdgetrf:
replicaset is multi-tenant

chenqianfzh:
FYI, the controllers that I have done the multi-tenancy changes:
1. Deployment
2. ReplicaSet
3. endpoints
4. ServiceAccount
5. Token
6. Namespace

pdgetrf:
@chenqianfzh do you have additional multi-tenancy test cases for all these?

chenqianfzh:
> @chenqianfzh did you add additional multi-tenancy test cases for all these?

Yes, each 

> @chenqianfzh did you add additional multi-tenancy test cases for all these?

Yes. Each test case before multi-tenancy is expanded into two cases, one for "default" tenant to ensure backward-compatibility and one for "test-te" tenant to ensure multi-tenancy is working.

pdgetrf:
awesome.

pdgetrf:
these are among the first batch to be finished with multi-tenancy

job controller
replication (not top priority)
daemonset
statefulset
resource quota

pdgetrf:
What's the tenant story for PV? Current behavior (from [this post](https://stackoverflow.com/questions/35364367/share-persistent-volume-claims-amongst-containers-in-kubernetes-openshift/35366775#35366775)):

the PV is global in the sense that it can be seen/accessed by any project/namespace, HOWEVER once it is bound to a project, it can then only be accessed by containers from the same project/namespace

the PVC is project/namespace specific (so if you have multple projects you would need to have a new PV and PVC for each project to connect to the shared NFS volume - can not reuse the PV from first project)

pdgetrf:
moved to 530 as controller work continues. 

pv and pvc PR is out at https://github.com/futurewei-cloud/arktos/pull/236. this is a huge change. at the moment fixing one last integration test in scheduler.

[...]

PR #271, opened 2020-05-06
add multi-tenancy to statefulset

pdgetrf:
Part of https://github.com/futurewei-cloud/arktos/issues/186

**Notes for reviewers**
- Commit 1  and 3 are the actual change. commit 2 is "make update" autogenerated files for changes in the pv and pvc PR and therefore deserves little time in reviewing.
- All unit tests for statefulset are now using a non-system & non-default test tenant. Before this, and in other unit tests too, we've been using system tenant for backward compatible test. They make changing tests very labor intensive and is actually violating how unit test should work. In this PR I'm arguing that the backward case with system tenant shouldn't be in the unit test for other part of the code. We should either make sure the unit test cases with system tenant (specificially with the lister functions) are sufficient, and/or add integration/e2e tests to verify it. 

#### create pv
```bash
TENANT            NAME                       CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS      CLAIM   STORAGECLASS   REASON   AGE
futureweitenant   persistentvolume/my-pv-0   1Gi        RWO            Retain           Available           manual                  7s
futureweitenant   persistentvolume/my-pv-1   1Gi        RWO            Retain           Available           manual                  7s
```

#### create statefulset with 2 pods consuming pv
```bash
TENANT            NAMESPACE             NAME                            HASHKEY               READY   STATUS    RESTARTS   AGE
futureweitenant   futurewei-namespace   pod/goose-ss-0                  1118939078841033003   1/1     Running   0          9s
futureweitenant   futurewei-namespace   pod/goose-ss-1                  1176681893112659923   1/1     Running   0          6s
```

#### PVCs were created automatically:
```bash
TENANT            NAMESPACE             NAME                                   STATUS   VOLUME    CAPACITY   ACCESS MODES   STORAGECLASS   AGE
futureweitenant   futurewei-namespace   persistentvolumeclaim/www-goose-ss-0   Bound    my-pv-1   1Gi        RWO            manual         28s
futureweitenant   futurewei-namespace   persistentvolumeclaim/www-goose-ss-1   Bound    my-pv-0   1Gi        RWO            manual         25s
```

#### Deletion was tested by deleting the statefulset and related service first, then manually delete pvc and pv, following the process detailed by [this](https://kubernetes.io/docs/tasks/run-application/delete-stateful-set/)

### DIFF
--- a/pkg/api/v1/pod/util.go
+++ b/pkg/api/v1/pod/util.go
@@ -1,5 +1,6 @@
 /*
 Copyright 2015 The Kubernetes Authors.
+Copyright 2020 Authors of Arktos - file modified.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -270,6 +271,7 @@ func GetPodConditionFromList(conditions []v1.PodCondition, conditionType v1.PodC
 	if conditions == nil {
 		return -1, nil
 	}
+
 	for i := range conditions {
 		if conditions[i].Type == conditionType {
 			return i, &conditions[i]
--- a/pkg/controller/history/controller_history.go
+++ b/pkg/controller/history/controller_history.go
@@ -75,6 +75,7 @@ func NewControllerRevision(parent metav1.Object,
 		ObjectMeta: metav1.ObjectMeta{
 			Labels:          labelMap,
 			OwnerReferences: []metav1.OwnerReference{*metav1.NewControllerRef(parent, parentKind)},
+			Tenant:          parent.GetTenant(),
 		},
 		Data:     data,
 		Revision: revision,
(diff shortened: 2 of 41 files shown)


api/openapi-spec/swagger.json
```diff
@@ -19323,7 +19323,7 @@
   },
   "info": {
     "title": "Kubernetes",
-    "version": "v1.15.0"
+    "version": "v0.0.0"
```
chenqianfzh, line 19326: Need to revert this. BTW, why the version is updated to "0.0"? the Arktos is at 0.2.

I will try to get the PR to fix the update on swagger.json merged next week.

chenqianfzh:
> Part of #186
> 
> **Notes for reviewers**

> * All unit tests for statefulset are now using a non-system & non-default test tenant. Before this, and in other unit tests too, we've been using system tenant for backward compatible test. They make changing tests very labor intensive and is actually violating how unit test should work. In this PR I'm arguing that the backward case with system tenant shouldn't be in the unit test for other part of the code. We should either make sure the unit test cases with system tenant (specificially with the lister functions) are sufficient, and/or add integration/e2e tests to verify it.
> 

Agreed. Good point.

api/openapi-spec/swagger.json
pdgetrf: this is auto-generated. what's the best way to revert this other than manually fixing it?

chenqianfzh: once https://github.com/futurewei-cloud/arktos/pull/246 is in, we will be free of this nuisance.

Issue #223, opened 2020-04-22
Vertical Scaling for Arktos


vinaykul:
**What would you like to be added**: Vertical Scaling Feature for Arktos

**Why is this needed**: This feature supports optimal resource management and better usage of cloud resources.

The following tasks have been identified in delivery of this feature.

- [ ] Design Document for Container Pods Vertical Scaling
- [ ] Design Document for Virtual Machine Pods Vertical Scaling
- [ ] API Changes for Containers
- [ ] API Changes to support Virtual Machine
- [ ] Scheduler Changes
- [ ] Kubelet Changes for Container Pods
- [ ] CRI changes for Container Pods
- [ ] Kubelet Changes for VM Pods
- [ ] CRI/Virtlet Changes to support VMs
- [ ] E2E Tests for Containers
- [ ] E2E Tests for VMs

PR #275, opened 2020-05-10
In-place Pod Vertical Scaling for container pods

vinaykul:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
 /kind feature
> /kind flake

**What this PR does / why we need it**: This is the initial implementation of the new community-agreed design for In-Place Pod Vertical Scaling feature.

Fixes #

Design Document: https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/20181106-in-place-update-of-pod-resources.md
Kubelet-CRI Design Document: https://github.com/kubernetes/enhancements/blob/master/keps/sig-node/20191025-kubelet-container-resources-cri-api-changes.md

```release-note

```


vinaykul:
reopen to retest

pdgetrf:
is this ready for review?

vinaykul:
> is this ready for review?

yes it is. please review.

pkg/apis/core/validation/validation.go
```diff
@@ -3745,15 +3744,6 @@ func ValidatePodUpdate(newPod, oldPod *core.Pod) field.ErrorList {
 		allErrs = append(allErrs, field.Invalid(specPath.Child("activeDeadlineSeconds"), newPod.Spec.ActiveDeadlineSeconds, "must not update from a positive integer to nil value"))
 	}
 
-	if utilfeature.DefaultFeatureGate.Enabled(features.InPlacePodVerticalScaling) {
```
pdgetrf, line 3748: why was this removed?

pkg/apis/core/validation/validation_test.go
```diff
@@ -8928,223 +8928,6 @@ func TestValidatePodUpdate(t *testing.T) {
 			"Forbidden: pod updates may not change fields other than",
 			"storage resource allocation change",
 		},
-		{
```
pdgetrf, line 8931: why were these cases removed?

pkg/kubelet/cm/cgroup_manager_linux.go
```diff
@@ -591,3 +594,102 @@ func (m *cgroupManagerImpl) GetResourceStats(name CgroupName) (*ResourceStats, e
...
+}
+
+// Set the memory limit in bytes applied to the cgroup
+func (m *cgroupManagerImpl) SetCgroupMemoryLimit(name CgroupName, memoryLimit int64) error {
+	cgroupPaths := m.buildCgroupPaths(name)
+	cgroupMemoryPath, found := cgroupPaths["memory"]
+	if !found {
+		return fmt.Errorf("failed to build memory cgroup fs path for cgroup %v", name)
+	}
+	memLimit := strconv.FormatInt(memoryLimit, 10)
```
pdgetrf, line 661: why 10? please replace 10 with a constant variable with a self-explanatory name

pkg/kubelet/cm/cgroup_manager_linux.go
```diff
@@ -591,3 +594,102 @@ func (m *cgroupManagerImpl) GetResourceStats(name CgroupName) (*ResourceStats, e
...
+// Set the cpu quota, cpu period, and cpu shares applied to the cgroup
+func (m *cgroupManagerImpl) SetCgroupCpuLimit(name CgroupName, cpuQuota *int64, cpuPeriod, cpuShares *uint64) error {
+	var cpuQuotaStr, cpuPeriodStr, cpuSharesStr string
+	cgroupPaths := m.buildCgroupPaths(name)
+	cgroupCpuPath, found := cgroupPaths["cpu"]
+	if !found {
+		return fmt.Errorf("failed to build cpu cgroup fs path for cgroup %v", name)
+	}
+	if cpuQuota != nil {
+		cpuQuotaStr = strconv.FormatInt(*cpuQuota, 10)
```
pdgetrf, line 677: same with 10 here and below

pkg/kubelet/cm/pod_container_manager_linux.go
```diff
@@ -324,3 +345,19 @@ func (m *podContainerManagerNoop) GetAllPodsFromCgroups() (map[types.UID]CgroupN
 func (m *podContainerManagerNoop) IsPodCgroup(cgroupfs string) (bool, types.UID) {
 	return false, types.UID("")
 }
+
+func (m *podContainerManagerNoop) GetPodCgroupMemoryLimit(_ *v1.Pod) (uint64, error) {
+	return 0, nil
```
pdgetrf, line 350: why are these functions returning contants?

pkg/kubelet/kubelet.go
```diff
@@ -1525,6 +1529,56 @@ func (kl *Kubelet) syncPod(o syncPodOptions) error {
...
+		if updateType == kubetypes.SyncPodCreate && !ok {
+			for idx := range apiPodStatus.ContainerStatuses {
+				if pod.Spec.Containers[idx].Resources.Limits == nil {
+					apiPodStatus.ContainerStatuses[idx].Resources.Limits = nil
+					continue
+				}
+				if apiPodStatus.ContainerStatuses[idx].Resources.Limits == nil {
+					apiPodStatus.ContainerStatuses[idx].Resources.Limits = make(v1.ResourceList)
+				}
+				//if podStatus.ContainerStatuses[idx].ResourceLimits == nil {
```
pdgetrf, line 1546: remove commented out code

pkg/kubelet/kubelet.go
```diff
@@ -1525,6 +1529,56 @@ func (kl *Kubelet) syncPod(o syncPodOptions) error {
...
+				//	podStatus.ContainerStatuses[idx].ResourceLimits = make(v1.ResourceList)
+				//}
+				if cpuLimit, found := pod.Spec.Containers[idx].Resources.Limits[v1.ResourceCPU]; found {
+					apiPodStatus.ContainerStatuses[idx].Resources.Limits[v1.ResourceCPU] = *cpuLimit.Copy()
+					//	podStatus.ContainerStatuses[idx].ResourceLimits[v1.ResourceCPU] = cpuLimit.Copy()
+				}
+				if memLimit, found := pod.Spec.Containers[idx].Resources.Limits[v1.ResourceMemory]; found {
+					apiPodStatus.ContainerStatuses[idx].Resources.Limits[v1.ResourceMemory] = *memLimit.Copy()
+					//	podStatus.ContainerStatuses[idx].ResourceLimits[v1.ResourceMemory] = memLimit.Copy()
```
pdgetrf, line 1555: here too

pkg/kubelet/kubelet.go
```diff
@@ -1525,6 +1529,56 @@ func (kl *Kubelet) syncPod(o syncPodOptions) error {
...
+				if _, found := apiPodStatus.ContainerStatuses[idx].Resources.Limits[v1.ResourceMemory]; !found {
+					if _, existing := existingStatus.ContainerStatuses[idx].Resources.Limits[v1.ResourceMemory]; existing {
+						apiPodStatus.ContainerStatuses[idx].Resources.Limits[v1.ResourceMemory] = existingStatus.ContainerStatuses[idx].Resources.Limits[v1.ResourceMemory]
+					}
+				}
+			}
+		}
+	}
+
 	// debug logging
```
pdgetrf, line 1582: this comment may not be needed.

pkg/kubelet/kubelet.go
```diff
@@ -2124,6 +2178,43 @@ func (kl *Kubelet) HandlePodAdditions(pods []*v1.Pod) {
...
+
+	var otherActivePods []*v1.Pod
+	activePods := kl.GetActivePods()
+	for _, p := range activePods {
+		if p.UID != pod.UID {
+			otherActivePods = append(otherActivePods, p)
+		}
+	}
+	if ok, failReason, failMessage := kl.canAdmitPod(otherActivePods, pod); !ok {
+		klog.Warningf("Pod '%s' resize cannot be accommodated. Reason: '%s' Message: '%s'", pod.Name, failReason, failMessage) //TODO: logV(4) not warn
```
pdgetrf, line 2201: so do we want log.V(4) or warn?

pkg/kubelet/kubelet_test.go
```diff
@@ -619,12 +619,20 @@ func TestHandleMemExceeded(t *testing.T) {
...
-		Containers: []v1.Container{{Resources: v1.ResourceRequirements{
-			Requests: v1.ResourceList{
-				v1.ResourceMemory: resource.MustParse("90"),
+	spec := v1.PodSpec{
+		NodeName: string(kl.nodeName),
+		Containers: []v1.Container{
+			{
+				Resources: v1.ResourceRequirements{
+					Requests: v1.ResourceList{
+						v1.ResourceMemory: resource.MustParse("90"),
```
pdgetrf, line 628: use integer constants with good name instead of 90

pkg/kubelet/kubelet_test.go
```diff
@@ -716,53 +724,83 @@ func TestHandlePluginResources(t *testing.T) {
...
+				Resources: v1.ResourceRequirements{
+					Limits: v1.ResourceList{
+						emptyResource: resourceQuantity2,
+					},
+					Requests: v1.ResourceList{
+						emptyResource: resourceQuantity2,
+					},
+				},
+				ResourcesAllocated: v1.ResourceList{
+					emptyResource: resourceQuantity2,
```
pdgetrf, line 760: rename resourceQuantity2 and resourceQuantity1 to more explanatory names. I have no idea what resourceQuantity2 is by looking at this part of the code

pkg/kubelet/kuberuntime/kuberuntime_container.go
```diff
@@ -459,18 +477,42 @@ func (m *kubeGenericRuntimeManager) getPodContainerStatuses(uid kubetypes.UID, n
 func toKubeContainerStatus(status *runtimeapi.ContainerStatus, runtimeName string) *kubecontainer.ContainerStatus {
 	annotatedInfo := getContainerInfoFromAnnotations(status.Annotations)
 	labeledInfo := getContainerInfoFromLabels(status.Labels)
+	var rLimits v1.ResourceList
+	if utilfeature.DefaultFeatureGate.Enabled(features.InPlacePodVerticalScaling) {
+		var cpuLimit, memLimit *resource.Quantity
+		if status.Resources.CpuShares > 0 {
+			//BUGBUG: TODO: Correctly compute cpuLimit from CpuShares, CpuQuota, CpuPeriod
```
pdgetrf, line 484: is this still a bug?

pdgetrf:
please add manual e2e test info to the PR description

pkg/kubelet/kubelet.go
```diff
@@ -1525,6 +1529,56 @@ func (kl *Kubelet) syncPod(o syncPodOptions) error {
 		metrics.DeprecatedPodStartLatency.Observe(metrics.SinceInMicroseconds(firstSeenTime))
 	}
 
+	if utilfeature.DefaultFeatureGate.Enabled(features.InPlacePodVerticalScaling) {
+		//NOTE: Currently, CRI does not return CPU limit info with dockershim (docker API limitation)
+		//      If there's no existing API pod status upon create, then set pod status resources limits from pod spec.
+		//      If existing API pod status is found, then set limits from there if not reported by kube pod stattus
+		//TODO: DOES THIS HANDLE THE CASE WHERE CRI DOES NOT SUPPORT RETURNING CPU AND MEMORY LIMIT INFO?
```
pdgetrf, line 1536: can we keep the comments to describe the intent of the code instead of asking question and listing TODOs? Maybe move TODOs to the tracking issue?

pkg/scheduler/core/generic_scheduler_test.go
```diff
@@ -643,6 +649,12 @@ func TestZeroRequest(t *testing.T) {
 							strconv.FormatInt(priorityutil.DefaultMemoryRequest*3, 10)),
 					},
 				},
+				ResourcesAllocated: v1.ResourceList{
+					v1.ResourceCPU: resource.MustParse(
+						strconv.FormatInt(priorityutil.DefaultMilliCPURequest*3, 10) + "m"),
```
pdgetrf, line 654: why 3 and 10?

pkg/scheduler/core/generic_scheduler_test.go
```diff
@@ -807,6 +825,12 @@ var mediumContainers = []v1.Container{
 					strconv.FormatInt(priorityutil.DefaultMemoryRequest*2, 10)),
 			},
 		},
+		ResourcesAllocated: v1.ResourceList{
+			"cpu": resource.MustParse(
+				strconv.FormatInt(priorityutil.DefaultMilliCPURequest*2, 10) + "m"),
```
pdgetrf, line 830: and why 2 and 10 here?

pkg/apis/core/validation/validation.go
vinaykul: moved to podresourceallocation admission plugin. this check happens earlier now.

pkg/apis/core/validation/validation_test.go
vinaykul: i'm planning to add them to podresourceallocation admission plugin.

pkg/kubelet/cm/cgroup_manager_linux.go
vinaykul: It is base value. This is standard across all ok k8s codebase. I prefer to stick with it for consistency.

pkg/kubelet/cm/cgroup_manager_linux.go
vinaykul: Yes. I prefer to stay with it for consistency with rest of codebase

pkg/kubelet/kubelet_test.go
vinaykul: this is unit test. it is pretty standard for using actual numbers since this is not expected to be varied. i'd prefer to stay with it for consistency with rest of the tests.

pkg/scheduler/core/generic_scheduler_test.go
vinaykul: this is set from the above resources line. we need to set it because the podresourceallocation plugin defaults the ResourcesAllocated to Resources.Requests if not set per the new KEP design.

pkg/scheduler/core/generic_scheduler_test.go
vinaykul: set from Resources.Requests above.

pkg/kubelet/kubelet_test.go
pdgetrf: constants with good name are for readability.

pkg/kubelet/cm/pod_container_manager_linux.go
vinaykul: this is no-op implementation.

pkg/kubelet/kubelet.go
vinaykul: thanks. i'll fix it.

pkg/kubelet/kubelet.go
vinaykul: will do.

pkg/kubelet/kubelet.go
vinaykul: this was not a change i made.

pkg/kubelet/kubelet.go
vinaykul: V4. i'm fixing it.

pkg/kubelet/kubelet_test.go
vinaykul: I agree that those names are not intuitive. (in this case they seem to refer to zero/failed/adjusting resources. The upstream code has it this way - I guess unit tests have this pattern.  See: https://github.com/kubernetes/kubernetes/blob/release-1.15/pkg/kubelet/kubelet_test.go#L640

I'm just extending it for the new feature.

pkg/kubelet/kuberuntime/kuberuntime_container.go
vinaykul: i don't think so, but i need to add thorough unit tests around it to verify all cases are covered. i have it there to get my attention when i come back to this after implementing resize policy.

pkg/kubelet/kubelet.go
vinaykul: i'll create a tracking issue for this. but this is on top of my mind since i need to discuss with sig-node - may need to update the CRI kep design based on their input.

pkg/kubelet/kubelet_test.go
vinaykul: it is not generally done in unit tests. see the rest of the unit test code that came from upstream.

pkg/kubelet/kuberuntime/kuberuntime_container.go
pdgetrf: please remove if it's done with.

yb01:
A generic question: how would vertical scaling works for windows containers ?

pkg/kubelet/kubelet.go
```diff
@@ -2124,6 +2178,43 @@ func (kl *Kubelet) HandlePodAdditions(pods []*v1.Pod) {
 	}
 }
 
+func (kl *Kubelet) checkPodResourcesResize(pod *v1.Pod) {
+	podResized := false
+	for _, container := range pod.Spec.Containers {
+		if len(diff.ObjectDiff(container.ResourcesAllocated, container.Resources.Requests)) > 0 {
```
yb01, line 2180: should the comparison with container.Resouces.limits instead of the requested values ?

pkg/kubelet/kuberuntime/kuberuntime_container.go
yb01: +1. we should remove it.

pkg/kubelet/kuberuntime/kuberuntime_manager.go
```diff
@@ -1286,6 +1381,55 @@ func (m *kubeGenericRuntimeManager) SyncPodContainer(pod *v1.Pod, podStatus *kub
 	return
 }
 
+func (m *kubeGenericRuntimeManager) updateContainerLimits(pod *v1.Pod, podStatus *kubecontainer.PodStatus, containersToUpdate []int) error {
```
yb01, line 1384: this seems update all resource limits and for all containers. should it be a atomic operation ? for example, if update failed in the middle for some reason, it will end up some are updated and some are not. what would be the consequence of this case ? retry can help but not guaranteed.

pkg/scheduler/algorithm/predicates/predicates.go
```diff
@@ -746,7 +746,7 @@ func (c *VolumeZoneChecker) predicate(pod *v1.Pod, meta PredicateMetadata, nodeI
 func GetResourceRequest(pod *v1.Pod) *schedulernodeinfo.Resource {
 	result := &schedulernodeinfo.Resource{}
 	for _, workload := range pod.Spec.Workloads() {
-		result.Add(workload.Resources.Requests)
+		result.Add(workload.ResourcesAllocated)
```
yb01: this change should be guarded by the featureGate()

pkg/scheduler/algorithm/priorities/resource_allocation.go
```diff
@@ -96,9 +96,15 @@ func getNonZeroRequests(pod *v1.Pod) *schedulernodeinfo.Resource {
 	result := &schedulernodeinfo.Resource{}
 	for i := range pod.Spec.Workloads() {
 		workload := &pod.Spec.Workloads()[i]
-		cpu, memory := priorityutil.GetNonzeroRequests(&workload.ResourcesAllocated)
-		result.MilliCPU += cpu
-		result.Memory += memory
+		if utilfeature.DefaultFeatureGate.Enabled(features.InPlacePodVerticalScaling) {
```
yb01, line 99: nit: only the GetNonzeroRequests() function is needed under the featureGate()

pkg/kubelet/kubelet.go
vinaykul: i'm going to update the code. i got a suggestion from the sig-node discussion this tuesday which simplifies the code. i tried a few different variations of QoS in pods and seems to work. this code is not necessary. i'll send an updated review by eod tomorrow

pkg/kubelet/kubelet.go
```diff
@@ -2124,6 +2178,43 @@ func (kl *Kubelet) HandlePodAdditions(pods []*v1.Pod) {
...
+		// TODO: Setup retry-backoff timer
+		return
+	}
+
+	for i, container := range pod.Spec.Containers {
+		for rName, rQuantity := range container.Resources.Requests {
+			pod.Spec.Containers[i].ResourcesAllocated[rName] = rQuantity
+		}
+	}
+	_, updateError := kl.kubeClient.CoreV1().Pods(pod.Namespace).Update(pod)
```
XiaoningDing: It looks like this only works with pods in system space. Does this work with pods in tenant spaces?

staging/src/k8s.io/cri-api/pkg/apis/runtime/v1alpha2/api.proto
```diff
@@ -949,6 +949,8 @@ message ContainerStatus {
     repeated Mount mounts = 14;
     // Log path of container.
     string log_path = 15;
+    // Resource configuration of the container.
+    LinuxContainerResources resources = 16;
```
XiaoningDing, line 953: Does this require changes from the VM runtime server(virtlet)?

pkg/kubelet/container/helpers.go
```diff
@@ -93,6 +96,16 @@ func ShouldContainerBeRestarted(container *v1.Container, pod *v1.Pod, podStatus
 // the running container with its desired spec.
 func HashContainer(container *v1.Container) uint64 {
 	hash := fnv.New32a()
+	if utilfeature.DefaultFeatureGate.Enabled(features.InPlacePodVerticalScaling) {
+		cResources := container.Resources
```
XiaoningDing, line 100: Do we need to exclude ResizePolicy as well?

vinaykul: No, we don't. At this time, we don't allow user to change resize policy once a pod has been started. Only Resources and ResourcesAllocated fields becoming mutable shouldn't be allowed to change the hash that is used to determine container needs restart (e.g. image change must result in restart, but limit change should only trigger update)

pkg/kubelet/kubelet.go
vinaykul: Thanks for catching this, I'll fix it.

staging/src/k8s.io/cri-api/pkg/apis/runtime/v1alpha2/api.proto
vinaykul: It is not required, but would be ideal if runtime reported this. Currently, I'm handling the case where runtime does not support returning this info by assuming initially configured resource limits, or the limits from last successful UpdateContainerResources API call. If runtime reports this info, the runtime report gets precedence.

pkg/kubelet/kuberuntime/kuberuntime_container.go
vinaykul: i found a problem when adding unit tests yesterday. i'm replacing this.

pkg/kubelet/kubelet.go
vinaykul: not here. this is at the entry point of handling pod update. if user changes the desired resources.requests and resources.limits, first check is to verify that new requests (resource reservation on the node) still fits the node available resources. if it does, then node patches the resourcesAllocated in apiserver pod-spec confirming that node  accepts the desired change. limits is checked in SyncPod, we always drive container towards desired limits (check that limit change is valid, does not change QoS etc happens in validation)

pkg/kubelet/kuberuntime/kuberuntime_manager.go
vinaykul: this is one of the things we discussed in sig-node earlier this week. short answer: yes it will, and that's ok.

the runtime will report error on partial update, and we don't set our status values to the updated values, and we will keep retrying in order to reconcile. as long as the ordering of pod cgroup sizing update is done with failure cases into account, we will be fine. this would be the most robust approach.

when i add multi-container support, i'll also be splitting up cpu & memory so we won't be updating both at the same time. but we are not telling runtimes to rely on that.

pkg/scheduler/algorithm/predicates/predicates.go
vinaykul: fixed.

pkg/scheduler/algorithm/priorities/resource_allocation.go
vinaykul: fixed.

pkg/kubelet/kuberuntime/kuberuntime_container.go
vinaykul: fixed in the latest commit.

vinaykul:
rebuild

PR #276, opened 2020-05-11
multi-tenancy networking: create default network on new tenants

### LINKED ISSUE #255: multi-tenancy networking: tenant controller to create default network CR
extend tenant controller to create default network CR object for each tenant been added.
 
**Why is this needed**:
as required by [design proposal](https://github.com/futurewei-cloud/arktos/blob/master/docs/design-proposals/multi-tenancy/multi-tenancy-network.md).

h-w-chen:
**What type of PR is this?**
/kind feature

Tenant admins are able to create network objects in the tenant scope. In order to make network easier to use, the default network object of every tenant, we have an option to let system to create it automatically - see the [networking design](https://github.com/futurewei-cloud/arktos/blob/master/docs/design-proposals/multi-tenancy/multi-tenancy-network.md) for more detail. This default network creation is optional:  if default-network-template-path arg not provided, it is disabled.

**Below is how a default network is created when a tenant is being created:**
Assuming kube-apiserver has the feature of sharing CRD (_in separate PR on the way_) and the Network CR has been defined beforehand. 
Given kube-apiserver is running properly and listening on http 8080, start kube-controller-manager with the default network creation enabled:
```bash
$ cat /tmp/dn.tmpl
{
    "metadata": {
        "name": "default",
	"finalizers": ["arktos.futurewei.com/network"]
    },
    "spec": {
        "type": "vpc",
        "vpcID": "{{.}}-new-default"
    }
}
$ ./kube-controller-manager --master=http://127.0.0.1:8080 --default-network-template-path=/tmp/dn.tmpl 
```
Run kubectl to create a tenant (e.g. tenant v123v), and the default network object is seen as created:
```
$ ./kubectl get ne --tenant v123v default -o yaml
apiVersion: arktos.futurewei.com/v1
kind: Network
metadata:
  creationTimestamp: "2020-05-09T17:18:34Z"
  finalizers:
  - arktos.futurewei.com/network
  generation: 1
  hashKey: 239005215023981727
  name: default
  resourceVersion: "18624"
  selfLink: /apis/arktos.futurewei.com/v1/tenants/v123v/networks/default
  tenant: v123v
  uid: 9f631529-3479-4d62-8e74-081c42d74a40
spec:
  type: vpc
  vpcID: v123v-new-default
```
Fixes #255  

40 files in commit "auto generated code - network CR", mostly under generated/ subfolder,  are generated automatically by staging/src/k8s.io/arktos-ext/hack/update-codegen.sh and ```make update```

YES
```release-note
If kube-controller-manager has arg default-network-template-path with a valid template file path value, it shall automatically create the default network object when new tenant is created.
```


docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -40,6 +40,21 @@ The "type" field is the only mandatory field in a network object. For now there
 
 A command-line parameter named "default-network-template-path" of tenant controller will decide which default network will be created for a new space.
 
+The content of default network template file should reflect the Network object in json format, with ```{{.}}``` the replacement of tenant name, like
+```json
+{
+    "metadata": {
+        "name": "default",
+        "tenant": "{{.}}"
```
XiaoningDing, line 48: is this a valid example?  I didn't see some required fields like "apiVersion" or "kind".  Also "tenant" and "finalizers" probably are not needed here.

cmd/kube-controller-manager/app/options/tenantcontroller.go
```diff
@@ -35,6 +35,7 @@ func (o *TenantControllerOptions) AddFlags(fs *pflag.FlagSet) {
 
 	fs.DurationVar(&o.TenantSyncPeriod.Duration, "tenant-sync-period", o.TenantSyncPeriod.Duration, "The period for syncing tenant life-cycle updates")
 	fs.Int32Var(&o.ConcurrentTenantSyncs, "concurrent-tenant-syncs", o.ConcurrentTenantSyncs, "The number of tenant objects that are allowed to sync concurrently. Larger number = more responsive tenant termination, but more CPU (and network) load")
+	fs.StringVar(&o.DefaultNetworkTemplatePath, "default-network-template-path", o.DefaultNetworkTemplatePath, "The path to the template file for the default network spec in JSON of tenants. Empty means system won't create  the default network for new tenants")
```
XiaoningDing, line 38: Will it be a problem if there is no default network object?

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: this is a valid example. the apiVersion & version will be the only option, which is arktos.futurewei.com and v1, respectively. In the future if multi versions are introduced, template content can have explicit GV to specify the desired one.

tenant template seems extraneous - it will always be filled with the right value by system. I'll make code change accordingly.
finalizers is put intentionally to remind that termination of network objects requires cleanup of other related resources.

cmd/kube-controller-manager/app/options/tenantcontroller.go
h-w-chen: If a tenant has no default network object, it will have some problems with DNS & kubernetes service. We don't want to enforce every tenant __MUST__ have default network rule, as tenant admin can have choose to work on non-default networks.

In flat network, if there exists multi tenancy, we still need different kube-dns for each tenant. In current design, dns service is per network, not per tenant.  We need explicit default network object (to provide dns service info). For the question of whether there is other options in flat network, I'll open an issue to track it.

PR #280, opened 2020-05-13
tenant controller bootstraps cluster role and binding

pdgetrf:
tracking issue https://github.com/futurewei-cloud/arktos/issues/176

### Changes:
A cluster role and rolebinding are created by the tenant controller when a new tenant is created. This role acts as the admin role for the tenant. 

### Note to reviewers:
The SA cluster role for tenant-controller has been bumped up to all verb, groups and resource because this role will be assumed to create the default tenant admin role. As a tenant admin role, it would allow "full" access to the tenant's own world. A lower capability role (the tenant-controller role) cannot create a higher capability role (tenant admin role). There are a few such controller role setup in addition to the tenant-controller (full list [here](plugin/pkg/auth/authorizer/rbac/bootstrappolicy/controller_policy_test.go)).

### Manual e2e test:
Create a tenant **futureweitenant**, and then cluster role and binding are created automatically:
```bash
kubectl get role,rolebinding,clusterrole,clusterrolebinding --tenant=futureweitenant --all-namespaces                                                      ip-172-31-27-32: Wed May 13 00:23:54 2020

NAME                                                               AGE
clusterrole.rbac.authorization.k8s.io/futureweitenant-admin-role   10s
NAME                                                                              AGE
clusterrolebinding.rbac.authorization.k8s.io/futureweitenant-admin-role-binding   10s
```

The user name for this role is "admin". And this user immediately has access:
```bash
kubectl --context=futurewei-admin-context get pods
No resources found.
```
This show access is permitted. 

Try to access as another user with the correct cert, and it failed authorization:
```bash
kubectl --context=futurewei-johndoe-context get pods
Error from server (Forbidden): pods is forbidden: User "johndoe" cannot list resource "pods" in API group "" in the namespace "futurewei-namespace"
```

Now let tenant admin create role and binding for johndoe, and johndoe now passed authorization:
```bash
root@ip-172-31-27-32:~/arktos_testing# kubectl --context=futurewei-admin-context create -f futureweirole.yaml
role.rbac.authorization.k8s.io/futureweirole created
root@ip-172-31-27-32:~/arktos_testing# kubectl --context=futurewei-admin-context create -f futureweirolebinding.yaml
rolebinding.rbac.authorization.k8s.io/futureweirolebinding created
kubectl --context=futurewei-johndoe-context get pods
No resources found.
```

### DIFF
--- a/pkg/controller/tenant/tenant_controller.go
+++ b/pkg/controller/tenant/tenant_controller.go
@@ -25,6 +25,7 @@ import (
 	"time"
 
 	"k8s.io/api/core/v1"
+	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
@@ -44,7 +45,13 @@ import (
 	"k8s.io/klog"
 )
 
-var tenant_default_namespaces = [...]string{core.NamespaceDefault, core.NamespaceSystem, core.NamespacePublic}
+var tenantDefaultNamespaces = [...]string{core.NamespaceDefault, core.NamespaceSystem, core.NamespacePublic}
+
+const (
+	initialClusterRoleUser        = "admin"
+	initialClusterRoleName        = "admin-role"
+	initialClusterRoleBindingName = "admin-role-binding"
+)
 
 // tenantController is responsible for performing actions dependent upon a tenant phase
 type tenantController struct {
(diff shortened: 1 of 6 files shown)


pkg/controller/tenant/tenant_controller.go
```diff
@@ -180,24 +181,84 @@ func (tc *tenantController) syncTenant(tenantName string) (err error) {
...
+			Tenant: tenant,
+		},
+		Subjects: []rbacv1.Subject{
+			{Kind: rbacv1.UserKind, Name: "admin"},
+		},
+		RoleRef: rbacv1.RoleRef{APIGroup: rbacv1.GroupName, Kind: "ClusterRole", Name: initialClusterRoleName(tenant)},
+	}
+
+	var failures []error
+	if _, err := tc.kubeClient.RbacV1().ClusterRolesWithMultiTenancy(tenant).Create(role); err != nil && !errors.IsAlreadyExists(err) {
```
XiaoningDing, line 236: Is there any problem if we proceed to create roleBinding without role being created?  I guess it should be fine, just want to double check.

pkg/controller/tenant/tenant_controller.go
```diff
@@ -180,24 +181,84 @@ func (tc *tenantController) syncTenant(tenantName string) (err error) {
...
+	return nil, true
+}
+
+func initialClusterRoleName(tenant string) string {
+	return tenant + "-admin-role"
+}
+
+func initialClusterRoleBindingName(tenant string) string {
+	return initialClusterRoleName(tenant) + "-binding"
+}
```
XiaoningDing, line 256: In fact it's not necessary to add tenant name prefix.  These objects are under each individual space, so there is no naming conflict even they are all called "admin-role" and "admin-role-binding".  It's probably more nature and transparent to users.

pkg/controller/tenant/tenant_controller.go
```diff
@@ -180,24 +181,84 @@ func (tc *tenantController) syncTenant(tenantName string) (err error) {
...
+
+func initialClusterRoleRules() rbacv1.PolicyRule {
+	return rbacv1.PolicyRule{
+		Verbs:     []string{"*"},
+		APIGroups: []string{"*"},
+		Resources: []string{"*"},
+	}
```
chenqianfzh, line 263: We should exclude the resources, such as daemonset/nodes, which are accessible for cluster admin only.

chenqianfzh: Maybe we can leave it as it for now, as it would need to another PR to discover all those resources and put them here.

But we should log an issue and put some comment here.

pkg/controller/tenant/tenant_controller.go
chenqianfzh: +1.

It gives the user the feel that he owns a world. :-)

pdgetrf: good point. fixed

pkg/controller/tenant/tenant_controller.go
pdgetrf: yeah I've been back and forth about this. the idea is for resource like daemonset/node, the attempt to access will be blocked by other admission mechanism. we have [an issue](https://github.com/futurewei-cloud/arktos/issues/39) to track this.

pkg/controller/tenant/tenant_controller.go
pdgetrf: fixed

pkg/controller/tenant/tenant_controller.go
pdgetrf: rolebinding can be created before or without the referred role. manually tested this.

XiaoningDing: Great.

pkg/controller/tenant/tenant_controller.go
XiaoningDing: +1 on this. Other mechanisms would block that.  Also in some cases we might even allow  creating nodes in a tenant space.  So it's better not to assume it in the default admin role settings.

PR #284, opened 2020-05-13
Option for not downloading runtime deployment files in Arktos onebox setup

yb01:
**What type of PR is this?**
> /kind bug fix

Fix issue Arktos issue 166. provides an option for not downloading runtime deployment files from runtime repo if a local copy is prefered to use

Fixes 166

Tested locally:

1. Verify normal start of onebox will download runtime deployment files
  ./hack/arktos-up.sh 
2. with files in local directory, verify flag=false won't download them
  OVERWRITE_DEPLOYMENT_FILES=false ./hack/arktos-up.sh 
3. remove the local files and retry to ensure files are downloaded, even flag=false
  rm /tmp/* -f -r
  OVERWRITE_DEPLOYMENT_FILES=false ./hack/arktos-up.sh

### DIFF
--- a/hack/lib/common.sh
+++ b/hack/lib/common.sh
@@ -55,14 +55,28 @@ if [[ ! -e "${CONTAINERD_SOCK_PATH}" ]]; then
   exit 1
 fi
 
+# local function to download runtime deployment file
+copyRuntimeDeploymentFile() {
+  if [[ $# != 2 ]]; then
+    echo "Invalid args in copyRuntimeDeploymentFile"
+    exit 1
+  fi
+
+  fileName=$1
+  pathInSrc=$2
+  if [[ (${OVERWRITE_DEPLOYMENT_FILES} == "true") || (! -f ${VIRTLET_DEPLOYMENT_FILES_DIR}/${fileName}) ]]; then 
+    echo "Getting runtime deployment file " ${fileName}
+    wget -O ${VIRTLET_DEPLOYMENT_FILES_DIR}/${fileName}  ${VIRTLET_DEPLOYMENT_FILES_SRC}/${pathInSrc}
+  fi
+}
+
 # Get runtime deployment files
-echo "Getting runtime deployment files"
-wget -O ${VIRTLET_DEPLOYMENT_FILES_DIR}/libvirt-qemu  ${VIRTLET_DEPLOYMENT_FILES_SRC}/apparmor/libvirt-qemu
-wget -O ${VIRTLET_DEPLOYMENT_FILES_DIR}/libvirtd  ${VIRTLET_DEPLOYMENT_FILES_SRC}/apparmor/libvirtd
-wget -O ${VIRTLET_DEPLOYMENT_FILES_DIR}/virtlet  ${VIRTLET_DEPLOYMENT_FILES_SRC}/apparmor/virtlet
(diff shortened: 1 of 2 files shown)


hack/lib/common.sh
```diff
@@ -57,12 +57,29 @@ fi
...
+  wget -O ${VIRTLET_DEPLOYMENT_FILES_DIR}/vms  ${VIRTLET_DEPLOYMENT_FILES_SRC}/apparmor/vms
+fi
+
+if [[ (${OVERWRITE_DEPLOYMENT_FILES} == "true") || (! -f ${VIRTLET_DEPLOYMENT_FILES_DIR}/vmruntime.yaml) ]]; then
+  wget -O ${VIRTLET_DEPLOYMENT_FILES_DIR}/vmruntime.yaml  ${VIRTLET_DEPLOYMENT_FILES_SRC}/data/virtlet-ds.yaml
+fi
+
+if [[ (${OVERWRITE_DEPLOYMENT_FILES} == "true") || (! -f ${VIRTLET_DEPLOYMENT_FILES_DIR}/images.yaml) ]]; then
+  wget -O ${VIRTLET_DEPLOYMENT_FILES_DIR}/images.yaml  ${VIRTLET_DEPLOYMENT_FILES_SRC}/images.yaml
+fi
```
pdgetrf, line 82: maybe put these into a function to prevent repeating it

pdgetrf:
Need a bit more info in the description on how this works and how this was tested. Something like "setting the env variable xxx will prevent downloading from internet".

hack/lib/common.sh
```diff
@@ -57,12 +57,29 @@ fi
...
+if [[ (${OVERWRITE_DEPLOYMENT_FILES} == "true") || (! -f ${VIRTLET_DEPLOYMENT_FILES_DIR}/vms) ]]; then
+  wget -O ${VIRTLET_DEPLOYMENT_FILES_DIR}/vms  ${VIRTLET_DEPLOYMENT_FILES_SRC}/apparmor/vms
+fi
+
+if [[ (${OVERWRITE_DEPLOYMENT_FILES} == "true") || (! -f ${VIRTLET_DEPLOYMENT_FILES_DIR}/vmruntime.yaml) ]]; then
+  wget -O ${VIRTLET_DEPLOYMENT_FILES_DIR}/vmruntime.yaml  ${VIRTLET_DEPLOYMENT_FILES_SRC}/data/virtlet-ds.yaml
+fi
+
+if [[ (${OVERWRITE_DEPLOYMENT_FILES} == "true") || (! -f ${VIRTLET_DEPLOYMENT_FILES_DIR}/images.yaml) ]]; then
+  wget -O ${VIRTLET_DEPLOYMENT_FILES_DIR}/images.yaml  ${VIRTLET_DEPLOYMENT_FILES_SRC}/images.yaml
```
pdgetrf, line 81: will "make clean" remove the previously downloaded local copy?

yb01: no, make clean won't delete them. they can be put into user specified folder to keep them for future uses.

hack/lib/common.sh
yb01: good point. will do so.

yb01:
update the PR description with local test cases

PR #288, opened 2020-05-15
improve tenant extraction from x509 cert

pdgetrf:
#### Issues fixed:
- Missing checks for edge cases
- A bug that could cause index out of bound with invalid input
- No unit testing

#### How fixed:
Added stricter checks for edge cases, fixed the bug and add full-fledged unit tests

#### Tests
Also tested e2e in local environment with tenant creation, role assignment, and rbac accesses.

#### Notes
- Make update was also run but didn't generate additional changes.
- TDD is fun :) 

Thanks @chenqianfzh for reporting this issue.

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go
+++ b/staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go
@@ -20,6 +20,7 @@ package x509
 import (
 	"crypto/x509"
 	"crypto/x509/pkix"
+	"errors"
 	"fmt"
 	"net/http"
 	"strings"
@@ -32,6 +33,10 @@ import (
 	"k8s.io/apiserver/pkg/authentication/authenticator"
 	"k8s.io/apiserver/pkg/authentication/user"
 )
+
+const ErrTenantEmptyCN = "CN name cannot be empty"
+const ErrMultipleOrganizationsWithTenant = "more than one organization with tenants specified"
+const ErrMultipleTenants = "more than one tenants specified in an organization "
 
 var clientCertificateExpirationHistogram = prometheus.NewHistogram(
 	prometheus.HistogramOpts{
@@ -185,48 +190,51 @@ func DefaultVerifyOptions() x509.VerifyOptions {
 
 // CommonNameUserConversion builds user info from a certificate chain using the subject's CommonName
 var CommonNameUserConversion = UserConversionFunc(func(chain []*x509.Certificate) (*authenticator.Response, bool, error) {
-	if len(chain[0].Subject.CommonName) == 0 {
-		return nil, false, nil
+	userInfo, succeeded, err := convertToUserInfo(chain[0].Subject)
+	if !succeeded {
+		return nil, false, err
 	}
-
-	isNewformat := false
-	tenant := ""
-	if len(chain[0].Subject.Organization) > 0 {
-		parts := strings.Split(chain[0].Subject.Organization[0], ":")
-		if len(parts) > 0 && parts[0] == "tenant" {
-			isNewformat = true
-			tenant = parts[1]
-		}
-	}
-
-	// backward compatible
-	// old format only uses /O
-	//if len(chain[0].Subject.OrganizationalUnit)==0 {
-	if !isNewformat {
-		parts := strings.Split(chain[0].Subject.CommonName, ":")
-		if len(parts) > 1 {
-			tenant = parts[0]
-		}
-		return &authenticator.Response{
-			User: &user.DefaultInfo{
-				Name:   chain[0].Subject.CommonName,
-				Groups: chain[0].Subject.Organization,
-				Tenant: tenant,
-			},
-		}, true, nil
-	}
-
-	// new format uses both /O and /OU
-	// one and only tenant per user
-	if len(chain[0].Subject.Organization) != 1 {
-		return nil, false, nil
-	}
-
 	return &authenticator.Response{
-		User: &user.DefaultInfo{
-			Name:   chain[0].Subject.CommonName,
-			Groups: chain[0].Subject.OrganizationalUnit,
-			Tenant: tenant,
-		},
-	}, true, nil
+		User: userInfo,
+	}, true, err
 })
+
+func convertToUserInfo(subject pkix.Name) (*user.DefaultInfo, bool, error) {
+	if len(subject.CommonName) == 0 {
+		return nil, false, errors.New(ErrTenantEmptyCN)
+	}
+	tenant := ""
+	if len(subject.Organization) > 0 {
+		parts := strings.Split(subject.Organization[0], ":")
+		if parts[0] == "tenant" {
+			// new format (tenant string from Organization)
+			if len(subject.Organization) > 1 {
+				// with new format, tenant comes from
+				// Organization and only one is allowed
+				return nil, false, errors.New(ErrMultipleOrganizationsWithTenant)
+			}
+			numOfParts := len(parts)
+			if numOfParts == 2 {
+				tenant = parts[1]
+			} else if numOfParts > 2 {
+				return nil, false, errors.New(ErrMultipleTenants)
+			} // when numOfParts==1, tenant is empty
+			return &user.DefaultInfo{
+				Name:   subject.CommonName,
+				Groups: subject.OrganizationalUnit,
+				Tenant: tenant,
+			}, true, nil
+		}
+	}
+
+	// backward compatible, old format (tenant string from CN)
+	parts := strings.Split(subject.CommonName, ":")
+	if len(parts) > 1 {
+		tenant = parts[0]
+	}
+	return &user.DefaultInfo{
+		Name:   subject.CommonName,
+		Groups: subject.Organization,
+		Tenant: tenant,
+	}, true, nil
+}
--- a/staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509_test.go
+++ b/staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509_test.go
@@ -20,6 +20,7 @@ package x509
 import (
 	"crypto/tls"
 	"crypto/x509"
+	"crypto/x509/pkix"
 	"encoding/pem"
 	"errors"
 	"io/ioutil"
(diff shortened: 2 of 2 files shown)


staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go
```diff
@@ -185,48 +185,51 @@ func DefaultVerifyOptions() x509.VerifyOptions {
 
 // CommonNameUserConversion builds user info from a certificate chain using the subject's CommonName
 var CommonNameUserConversion = UserConversionFunc(func(chain []*x509.Certificate) (*authenticator.Response, bool, error) {
```
chenqianfzh, line 187: "error", the last value of return, is always set to nil.  Shall we let convertToUserInfo() return some error message and the error is popped up in the call stack?

staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go
```diff
@@ -185,48 +185,51 @@ func DefaultVerifyOptions() x509.VerifyOptions {
...
-		return &authenticator.Response{
-			User: &user.DefaultInfo{
-				Name:   chain[0].Subject.CommonName,
-				Groups: chain[0].Subject.Organization,
+	tenant := ""
+	if len(subject.Organization) > 0 {
+		parts := strings.Split(subject.Organization[0], ":")
+		if parts[0] == "tenant" {
+			// new format (tenant string from Organization)
+			if len(subject.Organization) > 1 {
```
chenqianfzh, line 206: You mean "len(parts) > 1", right?

staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go
```diff
@@ -185,48 +185,51 @@ func DefaultVerifyOptions() x509.VerifyOptions {
...
+				// with new format, tenant comes from
+				// Organization and only one is allowed
+				return nil, false
+			}
+			numOfParts := len(parts)
+			if numOfParts == 2 {
```
chenqianfzh, line 212: The cases that the numOfParts is 1 is not handled. So if subject.Organization[0]=="tenant", we will get a userInfo where the tenant is ""

pdgetrf: that's the current defined behavior. see test case "new format, no colon tenant in organization"

staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go
```diff
@@ -185,48 +185,51 @@ func DefaultVerifyOptions() x509.VerifyOptions {
...
+			}, true
+		}
 	}
 
-	// new format uses both /O and /OU
-	// one and only tenant per user
-	if len(chain[0].Subject.Organization) != 1 {
-		return nil, false, nil
+	// backward compatible, old format (tenant string from CN)
+	parts := strings.Split(subject.CommonName, ":")
```
chenqianfzh, line 226: Need to check the tenant is not empty. If CN is ":abc", the tenant might be empty.

pdgetrf: see test case "old format, nothing but a colon in CN"

staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go
pdgetrf: good call. adding.

staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go
```diff
@@ -185,48 +185,51 @@ func DefaultVerifyOptions() x509.VerifyOptions {
...
-			tenant = parts[0]
-		}
-		return &authenticator.Response{
-			User: &user.DefaultInfo{
-				Name:   chain[0].Subject.CommonName,
-				Groups: chain[0].Subject.Organization,
+	tenant := ""
+	if len(subject.Organization) > 0 {
+		parts := strings.Split(subject.Organization[0], ":")
+		if parts[0] == "tenant" {
```
chenqianfzh, line 204: if O="tenant" (no ":" in the end), it should be identified as an old format. But the current logic will treat it as a new format. 

So I think line 204 should be len(parts) >1 && parts[0] == "tenant"

staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go
chenqianfzh, line 212: the case of numOfParts == 1 is not handled. However, if line 204 is changed as suggested, no change is needed here.

staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go
```diff
@@ -185,48 +185,51 @@ func DefaultVerifyOptions() x509.VerifyOptions {
...
+				// Organization and only one is allowed
+				return nil, false
+			}
+			numOfParts := len(parts)
+			if numOfParts == 2 {
+				tenant = parts[1]
+			} else if numOfParts > 2 {
+				return nil, false
+			}
+			return &user.DefaultInfo{
```
chenqianfzh, line 217: Need to check the tenant value is not empty.

pdgetrf: tenant value being empty will fail any rbac check. this code here is only responsible to extract.

staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go
pdgetrf: I think O="tenant" should be treated as the new format.

staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go
pdgetrf: the case of numOfParts == 1 is handled implicitly as an "else" case of the "if" condition on line 212. this is on purpose.

staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509_test.go
```diff
@@ -848,6 +849,238 @@ func TestX509Verifier(t *testing.T) {
...
+				Name: "abc:def:ghi",
+				Groups: []string{
+					"xyz",
+				},
+				Tenant: "abc",
+			},
+			shouldSucceed: true,
+		},
+		"old format, nothing but a colon in CN": {
+			input: pkix.Name{
```
chenqianfzh, line 933: Do we need to support such a CN, just a ":"?

staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509_test.go
```diff
@@ -848,6 +849,238 @@ func TestX509Verifier(t *testing.T) {
...
+				OrganizationalUnit: []string{
+					"xyz",
+				},
+			},
+			expectUserInfo: &user.DefaultInfo{
+				Name: "abc",
+				Groups: []string{
+					"xyz",
+				},
+				Tenant: "",
```
chenqianfzh, line 1003: I don't think we should expect a valid cert will give a tenant of "".

staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509_test.go
pdgetrf: I prefer to not add any rule as to what should or shouldn't be in a CN.

staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509_test.go
pdgetrf: this code is not to decide whether a cert is valid or not. it only tries to extract tenant string.

pdgetrf:
> Guess we should have some offline discussion on what is a valid Arktos cert. The code allows a the tenant to be empty, which looks insecure to me.

it's not this code's job to perform authentication. an empty tenant will be denied by rbac when tenants do not match.

pdgetrf:
> Guess we should have some offline discussion on what is a valid Arktos cert. The code allows a the tenant to be empty, which looks insecure to me.

another way to look at this is this code is in the authentication part. this part is to verify the user is what he/she claims he/she is. the tenant being empty or not should be enforced by the authorization step which determines what the user from his/her tenant can or cannot do. if we really want, a check can be added to authorization to deny all request from an empty tenant.

staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go
pdgetrf: fixed. added error and changed tests accordingly

staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go
chenqianfzh: Make sense. Agree.

staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go
chenqianfzh: Got it. It is better to add a comment here to explain the case that O="tenant" is a valid case. It will help the maintenance of the code.

staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go
chenqianfzh: I see your point.

staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go
pdgetrf: will do

staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go
```diff
@@ -185,48 +188,52 @@ func DefaultVerifyOptions() x509.VerifyOptions {
...
-	if len(chain[0].Subject.Organization) > 0 {
-		parts := strings.Split(chain[0].Subject.Organization[0], ":")
-		if len(parts) > 0 && parts[0] == "tenant" {
-			isNewformat = true
-			tenant = parts[1]
-		}
+func convertToUserInfo(subject pkix.Name) (*user.DefaultInfo, bool, error) {
+	errTenantExtraction := errors.New(ErrTenantExtraction)
+	if len(subject.CommonName) == 0 {
+		return nil, false, errTenantExtraction
```
chenqianfzh, line 203: Is errors.New("CommonName can not be empty") more descriptive?

chenqianfzh:
> > Guess we should have some offline discussion on what is a valid Arktos cert. The code allows a the tenant to be empty, which looks insecure to me.
> 
> another way to look at this is this code is in the authentication part. this part is to verify the user is what he/she claims he/she is. the tenant being empty or not should be enforced by the authorization step which determines what the user from his/her tenant can or cannot do. we could enforce that user from an empty tenant cannot do anything and deny all such request but that falls into the authority of the authorization.

Good call. Thanks for the clarification.

staging/src/k8s.io/apiserver/pkg/authentication/request/x509/x509.go
pdgetrf: I'll split the error message for each specific cases.

pdgetrf: fixed

PR #293, opened 2020-05-18
ETCD Data partition design - revision # changes

Sindica:


deepak-vij:
LGTM at this time.

docs/design-proposals/etcd/ETCD-Data-Partition-Design.md
```diff
@@ -0,0 +1,67 @@
...
+* Sending events to a separated ETCD cluster is not covered in this design doc
+* Scaling out of API server/controller framework is not discussed in this doc
+* Reallocate objects from one ETCD cluster to another is not discussed in this doc 
+
+## Proposal
+
+### ETCD Revision Generation
+
+* We updated the existing ETCD revision number generation algorithm from "starting from
+1 and always increasing by 1" to include the factors of timestamp, cluster id, and write event.
```
XiaoningDing, line 47: A minor one:  the paragraph #2 says the original algorithm starts from 0 but here it says starts from 1.

docs/design-proposals/etcd/ETCD-Data-Partition-Design.md
```diff
@@ -0,0 +1,67 @@
...
+### ETCD Revision Generation
+
+* We updated the existing ETCD revision number generation algorithm from "starting from
+1 and always increases by 1" to include the factors of timestamp, cluster id, and write event.
+ - Event number: bit 0-12
+ - Cluster id: bit 13-18
+ - Timestamp in millisecond: bit 19-62
+It will allow maximum 8K events per millisecond in a single cluster, up to 64 clusters, till 
+approximately January 1st, 2500.
+
```
cathyhongzhang, line 53: I know this has been merged. But I have three questions on this:
1. Do you mean the ETCD resvision number will now consist of Timestamp+clusterID+EventNumber in which the EventNumber will start from 1 and increase by 1?
2. If an Arktos cluster has 3 ETCD partitions, does the CLusterID here mean the ETCD partition ID or the Arktos cluster ID?
3. What is the source that supplies the timestamp? If it is the ETCD server, then different partitions' ETCD servers may not sync up their timestamp, which may lead to mismatch between the real sequence and the recorded timestamp.

Sindica: 1. The event number will start from 0 and increase by 1
2. The clusterId is ETCD cluster id. We don't have Arktos cluster id.
3. We use our programming language (in the case golang) to get timestamp. It will get time from the host. Yes we know the time sync would be an issue and discussed in the design meeting. We have agreed to request server deploy NTP service and maintain relatively synchronized clock. We will add some time tolerance into API server. We also agreed that our goal is not to have a global sequence but "roughly sortable global unique ID".

Please check the email thread "ETCD partition revision consistency discussion" if you missed the meeting on 5/15/2020.

Thanks for your comment.

cathyhongzhang: Thanks for the clarification!

PR #299, opened 2020-05-20
Fixed the scheduler throughput issue - controller instance update

jshaofuturewei:
Running integration test to see if we can get ride of GenericInformerFactory to decrease cpu usage.


cmd/workload-controller-manager/config/controllerconfig.json
```diff
@@ -1,5 +1,5 @@
 {
-    "reportHealthIntervalInSecond": 10,
+    "reportHealthIntervalInSecond": 60,
```
Sindica, line 2: This is not an ultimate fix.
1. TTL is 100 seconds. One miss report could just remove the instance from the registry and cause hash key range updates.
2. 100 seconds to find controller instance is missing is not bearable.

pkg/cloudfabric-controller/controllerframework/controllerinstancemanager.go
```diff
@@ -125,29 +125,31 @@ func (cim *ControllerInstanceManager) addControllerInstance(obj interface{}) {
...
 	}
 	existingInstancesForType, ok := cim.currentControllers[newControllerInstance.ControllerType]
+	var existingInstance v1.ControllerInstance
+	var ok1 bool
+
 	if !ok { // real new controller instance
 		cim.currentControllers[newControllerInstance.ControllerType] = make(map[string]v1.ControllerInstance)
 	} else { // check existing controller instance
-		existingInstance, ok1 := existingInstancesForType[newControllerInstance.Name]
-		if ok1 {
```
Sindica, line 141: You can leave the logic here and just add cim.mux.Unlock() before cim.updateControllerInstance

jshaofuturewei: In that case we might need to relock it and release the lock again to update controller instance mapping

Sindica: Notify handler and then does cim.updateControllerInstance again. If ok1 == true, notifyHandler will be called twice. Once in addControllerInstance, the other in updateControllerInstance

cmd/workload-controller-manager/config/controllerconfig.json
jshaofuturewei: Updated TTL to 300

Sindica: Can this be changed back to 10s and TTL keep as 30s?

Sindica: I am asking because it looks like in production, WorkloadControllerManager will be auto restarted if it exits. In that case, it will take long time for unlock.

PR #300, opened 2020-05-21
Add support to ETCD revision format changes. See design doc https://g…

Sindica:
Add support to ETCD revision format changes. See design doc https://github.com/futurewei-cloud/arktos/blob/master/docs/design-proposals/etcd/ETCD-Data-Partition-Design.md

### DIFF
--- a/staging/src/k8s.io/apimachinery/pkg/util/diff/compare.go
+++ b/staging/src/k8s.io/apimachinery/pkg/util/diff/compare.go
@@ -1,0 +1,109 @@
+/*
+Copyright 2020 Authors of Arktos.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package diff
+
+import "time"
+
+var allowedNTPDiffInMilliSecond = uint64(5000)
+
+// For backwards compatible - can still use ETCD revision # starting from 1 and increases by 1 as backend storage
+var v2MinRevision = getRevisionNumber(time.Date(2020, 5, 10, 0, 0, 0, 0, time.UTC).UnixNano(), 0, 1)
+
+// RevisionIsNewer is used in event comparison to check whether revision 1 is newer than
+// revision 2 and should be sent/accepted/processed
+func RevisionIsNewer(revision1, revision2 uint64) bool {
+	if revision1 < v2MinRevision || revision2 < v2MinRevision {
+		return revision1 > revision2
+	}
+
+	result, isSameCluster := revisionCompare(revision1, revision2)
+	if isSameCluster {
+		return result > 0
+	}
+	// if from different cluster, allow sent result == 0
+	return result >= 0
+}
+
+// revisionCompare compares two revision #s.
+// Return -1 if revision 1 is less than revision 2
+// Return 0 if revision 1 is equal to revision 2
+// Return 1 if revision 1 is larger than revision 2
+// Note: the definition of less than, equal to, and larger than is defined as follows:
+// If they are from same cluster id, their value will be directly compared
+// If they are not from same cluster id, we use allowedNTPDiffInSecond to check whether
+//     the l
+func revisionCompare(revision1, revision2 uint64) (result int, isSameCluster bool) {
+	clusterId1 := extractClusterId(revision1)
+	clusterId2 := extractClusterId(revision2)
+	isSameCluster = clusterId1 == clusterId2
+	if isSameCluster {
+		if revision1 < revision2 {
+			result = -1
+		} else if revision1 == revision2 {
+			result = 0
+		} else {
+			result = 1
+		}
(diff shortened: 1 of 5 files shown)


staging/src/k8s.io/apimachinery/pkg/util/diff/compare.go
```diff
@@ -0,0 +1,109 @@
...
+
+var allowedNTPDiffInMilliSecond = uint64(5000)
+
+// For backwards compatible - can still use ETCD revision # starting from 1 and increases by 1 as backend storage
+var v2MinRevision = getRevisionNumber(time.Date(2020, 5, 10, 0, 0, 0, 0, time.UTC).UnixNano(), 0, 1)
+
+// RevisionIsNewer is used in event comparison to check whether revision 1 is newer than
+// revision 2 and should be sent/accepted/processed
+func RevisionIsNewer(revision1, revision2 uint64) bool {
+	if revision1 < v2MinRevision || revision2 < v2MinRevision {
```
jshaofuturewei, line 29: I suppose that reversion less than v2MinRevision  is the old one before we make the change. In that case, if revision1 < v2MinRevision  while  revision2 > v2MinRevision(revision2 is newer than revision1), the return result is revision1 newer than revision2  in case revision1 > revision2?

Sindica: yes

staging/src/k8s.io/apimachinery/pkg/util/diff/compare.go
```diff
@@ -0,0 +1,109 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package diff
+
+import "time"
+
+var allowedNTPDiffInMilliSecond = uint64(5000)
```
XiaoningDing, line 21: constant?

XiaoningDing, line 21: constant?

staging/src/k8s.io/apimachinery/pkg/util/diff/compare.go
```diff
@@ -0,0 +1,109 @@
...
+	return
+}
+
+// Based on design, int64 (ETCD revision type) has bits
+// 0-12 as event number
+// 13-18 as cluster id
+// TODO: Need to test whether it is the same bit number when etcd revision # is used in API server
+// as we are using uint64 here. Not sure how conversion happened
+func extractClusterId(rev uint64) uint64 {
+	return (rev >> 13) % 64
```
XiaoningDing, line 93: Usually "& 63" will have a better performance. But I guess Golang compiler probably have optimized for this.

staging/src/k8s.io/apimachinery/pkg/util/diff/compare.go
```diff
@@ -0,0 +1,109 @@
...
+	if revision1 < v2MinRevision || revision2 < v2MinRevision {
+		return revision1 > revision2
+	}
+
+	result, isSameCluster := revisionCompare(revision1, revision2)
+	if isSameCluster {
+		return result > 0
+	}
+	// if from different cluster, allow sent result == 0
+	return result >= 0
```
XiaoningDing, line 38: Shall we move the logic of isSameCluster into function revisionCompare()?  So this function will be very simple: if old format compare directly, otherwise call new format's comparison. Now the logics of isSameCluster are in two places and hard to understand.

staging/src/k8s.io/apimachinery/pkg/util/diff/compare.go
```diff
@@ -0,0 +1,109 @@
...
+}
+
+// revisionCompare compares two revision #s.
+// Return -1 if revision 1 is less than revision 2
+// Return 0 if revision 1 is equal to revision 2
+// Return 1 if revision 1 is larger than revision 2
+// Note: the definition of less than, equal to, and larger than is defined as follows:
+// If they are from same cluster id, their value will be directly compared
+// If they are not from same cluster id, we use allowedNTPDiffInSecond to check whether
+//     the l
```
XiaoningDing, line 48: Not completed?

staging/src/k8s.io/apimachinery/pkg/util/diff/compare.go
Sindica: RevisionIsNewer() is already very simple. I would like not to over-complicate the logic of revisionCompare().

staging/src/k8s.io/apimachinery/pkg/util/diff/compare.go
Sindica: There are unit tests that test with different value of allowedNTPDiffInMilliSecond. If it is constant, the test can not be done.

staging/src/k8s.io/apimachinery/pkg/util/diff/compare.go
Sindica: Good point. I did a performance test between % and &. The result are very interesting. When test time is 1M, % and & has almost the same performance. When test time is 100M, whichever runs earlier has better performance. Changing the sequence got roughly the same result.

Sindica: Keep original for readability and there is no performance issue.

staging/src/k8s.io/apimachinery/pkg/util/diff/compare.go
Sindica: Will keep the original as there is only one public function.

PR #310, opened 2020-05-24
Need check $PRESET_INSTANCES_ENABLED with default value

Hong-Chang:
This is hotfix that $PRESET_INSTANCES_ENABLED could be used without value

**What type of PR is this?**
/kind bug

### DIFF
--- a/cluster/kube-down.sh
+++ b/cluster/kube-down.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 
 # Copyright 2014 The Kubernetes Authors.
+# Copyright 2020 Authors of Arktos - file modified.
 #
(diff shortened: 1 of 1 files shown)


cluster/kube-down.sh
```diff
@@ -35,7 +35,7 @@ verify-prereqs
 echo "... calling verify-kube-binaries" >&2
 verify-kube-binaries
 echo "... calling kube-down" >&2
-if [[ $PRESET_INSTANCES_ENABLED == $TRUE ]]; then
+if [[ ${PRESET_INSTANCES_ENABLED:-} == $TRUE ]]; then
```
sonyafenge, line 38: I suggested to use default value = false.

Hong-Chang: Done

PR #311, opened 2020-05-24
explicit tenant all in url

chenqianfzh:
This PR address the issue to explicitly indicate whether a request targets all-tenants. So, with this change, for a request from a system user:
1. GET /api/v1/namespaces: means lists namesapces of system tenants
2.  GET /api/v1/tenants/all/namespaces: means lists namesapces of all the tenants, including the system tenants and regular tenants

TenantAll is "", similar to NamespaceAll in K8s, which is also "" . By making TenantAll as an empty string, the list operation on etcd is quite simple. Therefore, I did not change the definition of TenantAll in Arktos. When Arktos detects the tenant is "all" in the url, the tenant will be set TenantAll. If tenant is empty in the url, the tenant will be set according to short-path logic. 

Tests done in my local dev box:

# "all" is not an acceptable tenant name now.
```
qianchen@qianchen-VirtualBox:~/gopath/qianc/arktos$ kubectl create tenant all
The Tenant "all" is invalid: metadata.name: Invalid value: "all": is not an acceptable tenant name

qianchen@qianchen-VirtualBox:~/gopath/qianc/arktos$ kubectl get pods --tenant all --all-namespaces
error: "all" is not a valid tenant name
```

# verify that "tenants/all" work as expected
```
qianchen@qianchen-VirtualBox:~/gopath/qianc/arktos$ kubectl create -f ./test/yaml/multi_tenancy_examples/tenant1.yaml 
tenant/sample-tenant-1 created

qianchen@qianchen-VirtualBox:~/gopath/qianc/arktos$ kubectl create -f ./test/yaml/multi_tenancy_examples/tenant1-ns-a.yaml 
namespace/sample-namespace-a created

qianchen@qianchen-VirtualBox:~/gopath/qianc/arktos$ kubectl create -f ./test/yaml/multi_tenancy_examples/tenant1-ns-a-deployment.yaml 
deployment.apps/sample-nginx-deployment created

qianchen@qianchen-VirtualBox:~/gopath/qianc/arktos$ kubectl get pods --all-tenants --all-namespaces
TENANT            NAMESPACE            NAME                                       HASHKEY               READY   STATUS    RESTARTS   AGE
system            kube-system          kube-dns-5d69f5657d-drlpr                  5427732176834485377   3/3     Running   0          19m
system            kube-system          virtlet-pghwt                              9019363285369125196   3/3     Running   0          18m
sample-tenant-1   sample-namespace-a   sample-nginx-deployment-74cbdd5d99-fj68d   607901798557078899    1/1     Running   0          24s
sample-tenant-1   sample-namespace-a   sample-nginx-deployment-74cbdd5d99-g6rct   1950399659802567693   1/1     Running   0          24s
sample-tenant-1   sample-namespace-a   sample-nginx-deployment-74cbdd5d99-grhk8   6582451568891784181   1/1     Running   0          24s
```

In apiserver.log, I confirmed that the url has been changed to the expected
I0525 23:30:14.824513    8304 wrap.go:47] GET /api/v1/**tenants/all**/pods?limit=500: (3.105091ms) 200 [kubectl/v0.2 (linux/amd64) kubernetes/0f650bf 127.0.0.1:51676]

### DIFF
--- a/staging/src/k8s.io/apimachinery/pkg/api/validation/generic.go
+++ b/staging/src/k8s.io/apimachinery/pkg/api/validation/generic.go
@@ -18,10 +18,12 @@ limitations under the License.
 package validation
 
 import (
+	"fmt"
 	"strings"
 
 	"k8s.io/apimachinery/pkg/util/validation"
 	"k8s.io/apimachinery/pkg/util/validation/field"
+	api "k8s.io/kubernetes/pkg/apis/core"
 )
 
 const IsNegativeErrorMsg string = `must be greater than or equal to 0`
@@ -58,9 +60,16 @@ func NameIsDNS1035Label(name string, prefix bool) []string {
 }
 
 // ValidateTenantName can be used to check whether the given tenant name is valid.
-// Prefix indicates this name will be used as part of generation, in which case
-// trailing dashes are allowed.
-var ValidateTenantName = NameIsDNSLabel
+func ValidateTenantName(name string, prefix bool) []string {
+	forbiddenTenantNames := []string{api.TenantAllExplicit}
+	for _, forbiddenName := range forbiddenTenantNames {
+		if name == forbiddenName {
+			return []string{fmt.Sprintf("is not an acceptable tenant name")}
+		}
+	}
+
+	return NameIsDNSLabel(name, prefix)
+}
 
 // ValidateControllerTypeName can be used to check whether the given controller type name is valid.
 var ValidateControllerTypeName = NameIsDNSLabel
--- a/staging/src/k8s.io/client-go/tools/clientcmd/client_config.go
+++ b/staging/src/k8s.io/client-go/tools/clientcmd/client_config.go
@@ -318,6 +318,9 @@ func (config *DirectClientConfig) Tenant() (string, bool, error) {
 		// the tenant override instead of having config.ConfirmUsable() return an error. This allows
 		// things like in-cluster clients to execute `kubectl get pods --tenant=foo` and have the
 		// --tenant flag honored instead of being ignored.
+		if config.overrides.Context.Tenant == metav1.TenantAllExplicit {
+			return "", false, fmt.Errorf("%q is not a valid tenant name", metav1.TenantAllExplicit)
+		}
 		return config.overrides.Context.Tenant, true, nil
 	}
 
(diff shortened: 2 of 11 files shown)


pdgetrf:
are 1 and 2 supposed to be the same? trying to understand the reason for this change.

image

pdgetrf:
Still trying to understand the reason of this change. What would happen if we don't have this change?

staging/src/k8s.io/apimachinery/pkg/api/validation/generic.go
```diff
@@ -58,9 +60,16 @@ func NameIsDNS1035Label(name string, prefix bool) []string {
 }
 
 // ValidateTenantName can be used to check whether the given tenant name is valid.
-// Prefix indicates this name will be used as part of generation, in which case
-// trailing dashes are allowed.
-var ValidateTenantName = NameIsDNSLabel
+func ValidateTenantName(name string, prefix bool) []string {
```
pdgetrf, line 63: where is this used? 

for a function named Validatexxx, it's a bit weird to see it returns a list of string instead of boolean.

staging/src/k8s.io/client-go/tools/clientcmd/client_config.go
```diff
@@ -330,8 +333,8 @@ func (config *DirectClientConfig) Tenant() (string, bool, error) {
 		return "", false, err
 	}
 
-	if len(configContext.Tenant) == 0 {
-		return metav1.TenantSystem, false, nil
+	if configContext.Tenant == metav1.TenantAllExplicit {
+		return "", false, fmt.Errorf("%q is not a valid tenant name", metav1.TenantAllExplicit)
```
pdgetrf, line 337: nit: %q can be replaced with metav1.TenantAllExplicit as it's a constant.

chenqianfzh:
> Trying to understand the reason for this change. Are 1 and 2 supposed to be the same?
> 
> image

terrible typo. Fixed the description.
Now: GET /api/v1/tenants/all/namespaces: means lists namesapces of all the tenants, including the system tenants and regular tenants

staging/src/k8s.io/apimachinery/pkg/api/validation/generic.go
chenqianfzh: This func will be called in creating/updating an object. It will be called in ValidateObject() in k8s.io/apimachinery/pkg/api/validation/objectmeta.go, which is used to validate the object data .

As to the functio Validatexxx, the return value is an aggregation of errors. It should be better to return []error instead of []string. But as all the funcs about validation in this file and all the consumers of these funcs are working with the return value of []string, I made my code consistent with them.

staging/src/k8s.io/client-go/tools/clientcmd/client_config.go
chenqianfzh: It is better to keep it like this. If someday in the future we change to TenantAllExplicit = "somethging-else", we don't need the code here.

PR #312, opened 2020-05-25
Arktos runtime refactor: simplify the northbound imageService API

yb01:
**What type of PR is this?**
> /kind bug

This PR simplifies the north bound API impl and removed the internal map for pod-imageService cache mantained by the KuberuntimeManager. it added an extra field to the existing kubecontainer.ImageSpec to rely the info to get the desired image service in PullImage() interface.

This PR also addressed a few TODOs in kuberuntime image service related code.

TODOs in code
Partial fix for issue 84.

Fixes #


pkg/kubelet/kuberuntime/kuberuntime_image.go
```diff
@@ -111,42 +104,36 @@ func (m *kubeGenericRuntimeManager) PullImage(image kubecontainer.ImageSpec, pul
 // GetImageRef gets the ID of the image which has already been in
 // the local storage. It returns ("", nil) if the image isn't in the local storage.
 func (m *kubeGenericRuntimeManager) GetImageRef(image kubecontainer.ImageSpec) (string, error) {
+	if image.Pod == nil {
+		klog.Errorf("Pod is not set in imageSpec: %v", image)
+		return "", fmt.Errorf("POD is not set in imageSpec")
+	}
 
 	// Get the image service from the map cache for this image
```
h-w-chen, line 112: is this comment still applicable?

pkg/kubelet/container/runtime.go
```diff
@@ -52,6 +52,7 @@ type Version interface {
 // information about the different image types.
 type ImageSpec struct {
 	Image string
+	Pod *v1.Pod
```
h-w-chen, line 55: What is the driving reason to have Pod info in Image spec? Usually image is not pod specific.

pkg/kubelet/kuberuntime/kuberuntime_image.go
```diff
@@ -47,36 +51,27 @@ func (m *kubeGenericRuntimeManager) PullImage(image kubecontainer.ImageSpec, pul
...
-			} else {
-				klog.Infof("Got image ref for image %q: %v", image, imageRef)
-				return imageRef, nil
-			}
+		klog.Infof("Pulling image with image service %v", imageService)
+		imageRef, err := imageService.PullImage(imgSpec, nil, podSandboxConfig)
+		if err != nil {
+			klog.Errorf("Pull image %q failed: %v", img, err)
+		} else {
+			klog.Infof("Pull image ref for image %q: %v", image, imageRef)
```
h-w-chen, line 73: nit: for successful info, we may want to lower the logging priority (confusing; higher level like v5+) to reduce the log file size.

pkg/kubelet/kuberuntime/kuberuntime_image.go
```diff
@@ -111,42 +104,36 @@ func (m *kubeGenericRuntimeManager) PullImage(image kubecontainer.ImageSpec, pul
...
-		status, err := imageService.ImageStatus(&runtimeapi.ImageSpec{Image: image.Image})
-		if err != nil || status == nil {
-			klog.Infof("ImageStatus for image %q failed: %v", image, err)
-		} else {
-			klog.Errorf("Got ImageStatus for image %q: %v", image, status)
-			return status.Id, nil
-		}
-
+	status, err := imageService.ImageStatus(&runtimeapi.ImageSpec{Image: image.Image})
+	if err != nil || status == nil {
```
h-w-chen, line 125: err != nil,   klog.Errorf or Infof?

pkg/kubelet/kuberuntime/kuberuntime_image.go
```diff
@@ -111,42 +104,36 @@ func (m *kubeGenericRuntimeManager) PullImage(image kubecontainer.ImageSpec, pul
...
+		klog.Errorf("Got ImageStatus for image %q: %v", image, status)
+		return status.Id, nil
 	}
 
 	// if not found, return "", nil
 	return "", nil
 }
 
 // ListImages gets all images currently on the machine.
-// TODO: what if the dup image from different imageService ?
```
h-w-chen, line 148: no code change addressing the concerns of the comment to remove. Is the original comment not a real concern, in that case?

pkg/kubelet/kuberuntime/kuberuntime_image.go
h-w-chen, line 125: err != nil , Errorf; else Infof?

pkg/kubelet/kuberuntime/kuberuntime_image.go
h-w-chen, line 148: no code change addressing the concern recorded by the comment; is it not a real concern? If so, removing the comment makes sense.

yb01: yes, it should be safe to remove. the Image struct uses ImageId which, in vm runtime, generated from the full image data path from the filesystem. so even with multiple runtime services, the imageID conflict situation should be skipped regardless other runtime service's logic to generated the ID of the image.

pkg/kubelet/kuberuntime/kuberuntime_image.go
yb01: Good catch. my mistake. will fix.

pkg/kubelet/kuberuntime/kuberuntime_image.go
yb01: no. it should be removed. will delete the comment.

pkg/kubelet/kuberuntime/kuberuntime_image.go
yb01: yes, will do. thanks

pkg/kubelet/container/runtime.go
yb01: this is a fair ask. for now ( maybe forever ), we won't support mixed workload type so the info used to identify the image service in the API  chain is *pod -- this is mainly to simplify the way to get the desired image service -- using POD info is probably the most straightforward way with the current APIs.

pkg/kubelet/kuberuntime/kuberuntime_image.go
yb01: fixed.

PR #313, opened 2020-05-26
multi-tenancy: flat network controller

h-w-chen:
**What type of PR is this?**
/kind feature

networks of flat type are those sharing the same IP address space across whole cluster; they could be in different tenants. For this kind of flat networks, it is critical for each to have dedicated DNS service which provides proper service discovery in tenant/network scope; its IP address is kept in network status.dnsServiceIP.

Flat network controller is implemented as standalone controller to provision the DNS service when the new network is created.

This PR is only about the controller's core implementation. Ideally flat network controller would be packaged as operator, deployed as a deployment. The image build, deployment spec, and other related supporting files like SA/role definitions, will be addressed in separate PR. 

**How to verify**
1. start etcd / api server
2. define network CRD shared across all tenants
```kubectl apply -f crd-network.yaml```
3. start flat network controller
```./flat-network-controller --kubeconfig=/path-to/.kube/config```
4. create tenant x, and create a flat network, foo,  in tenant x
5. look at network and found its status and DN service IP values
 ```
$ kubectl get ne --tenant x
NAME       TYPE   VPC            PHASE   DNS
foo        flat   inside-of-x    Ready   10.0.0.135
```
YES
```
release notes:
1. a new custom controller, flat-network-controller, is introduced to provision networks of flat type. It accepts arguments of --kubeconfig, --master, and --concurrent-works.
2. network CR has status.dnsServiceIP to keep the DNS IP address of the network
```


cmd/flat-network-controller/app/controller.go
```diff
@@ -0,0 +1,247 @@
...
+	arktosinformer "k8s.io/arktos-ext/pkg/generated/informers/externalversions/arktosextensions/v1"
+	arktosv1 "k8s.io/arktos-ext/pkg/generated/listers/arktosextensions/v1"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
+	v1core "k8s.io/client-go/kubernetes/typed/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog"
+	"strings"
```
chenqianfzh, line 42: minor: in the import list, put the standard packages (licke "strings") before the k8s packages.

cmd/flat-network-controller/app/controller.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+	klog.V(5).Infof("%d workers started", workers)
+	<-stopCh
+	klog.Info("shutting down flat network controller")
+}
+
+func (c *Controller) runWorker() {
+	for {
+		item, shutdown := c.queue.Get()
```
chenqianfzh, line 97: minor: the variable name "shutdown" sounds scary. It just means the queue is empty if the value is "true". Consider change the variable to something like "done", "empty", etc.

cmd/flat-network-controller/app/controller.go
```diff
@@ -0,0 +1,247 @@
...
+		if shutdown {
+			break
+		}
+
+		c.process(item)
+	}
+}
+
+// processItem will read a single work item off the work queue and attempt to process it
+func (c *Controller) process(item interface{}) {
```
chenqianfzh, line 107: comment says "processItem", but func name is "process". I think processNextWorkItem is better, which is consistent with other controller/operator code in K8s.

cmd/flat-network-controller/app/controller.go
```diff
@@ -0,0 +1,247 @@
...
+	net, err := c.store.NetworksWithMultiTenancy(tenant).Get(name)
+	if err != nil {
+		klog.Warningf("failed to retrieve network in local cache by tenant %s, name %s: %v", tenant, name, err)
+		c.queue.Forget(item)
+		return
+	}
+
+	if net.Spec.Type != "flat" {
+		klog.V(5).Infof("network %s/%s is of type %q; ignored", net.Tenant, net.Name, net.Spec.Type)
+		c.queue.Forget(item)
```
chenqianfzh, line 133: I found every path in this func, either success or fail, will call c.queue.Forget(item). If we need to forget this item anyway, consider call it in the func of runWorker(), which the caller of process().

cmd/flat-network-controller/app/controller.go
```diff
@@ -0,0 +1,247 @@
...
+	}
+
+	net, err := c.store.NetworksWithMultiTenancy(tenant).Get(name)
+	if err != nil {
+		klog.Warningf("failed to retrieve network in local cache by tenant %s, name %s: %v", tenant, name, err)
+		c.queue.Forget(item)
+		return
+	}
+
+	if net.Spec.Type != "flat" {
```
chenqianfzh, line 131: We should define "flat" as a const where the network.Spec.Type is defined.

cmd/flat-network-controller/app/controller.go
```diff
@@ -0,0 +1,247 @@
...
+		klog.V(5).Infof("network %s/%s is of type %q; ignored", net.Tenant, net.Name, net.Spec.Type)
+		c.queue.Forget(item)
+		return
+	}
+
+	klog.V(5).Infof("processing network %s/%s", net.Tenant, net.Name)
+
+	if err := manageFlatNetwork(net, c.netClientset, c.svcClientset); err != nil {
+		c.queue.AddRateLimited(key)
+		c.recorder.Eventf(net, corev1.EventTypeWarning, "FailedProvision", "failed to provision network %s/%s/%s: %v", net.Tenant, net.Tenant, net.Name, err)
```
chenqianfzh, line 141: "net.Tenant, net.Tenant" -> should be  "net.Tenant, net.Namespace"

cmd/flat-network-controller/app/controller.go
```diff
@@ -0,0 +1,247 @@
...
+	netReady := net.DeepCopy()
+	netReady.Status.Phase = v1.NetworkReady
+	netReady.Status.DNSServiceIP = svc.Spec.ClusterIP
+	netReady.Status.Message = "DNS service ready; network ready"
+	_, err = netClient.ArktosV1().NetworksWithMultiTenancy(netReady.Tenant).UpdateStatus(netReady)
+	return err
+}
+
+func createOrGetDNSService(net *v1.Network, svcClient kubernetes.Interface) (*corev1.Service, error) {
+	const nsDNS = "kube-system"
```
chenqianfzh, line 171: not need to define nsDNS, we can use metav1.NamespaceSystem directly.

cmd/flat-network-controller/app/controller.go
chenqianfzh: The item is not "forget" if the network provision fails. Is it intentional? The logic here means the system will keep on trying provisioning without a limit.

cmd/flat-network-controller/app/controller.go
```diff
@@ -0,0 +1,247 @@
...
+	nameDNS := "kube-dns-" + net.Name
```
chenqianfzh, line 172: consider defining "kube-dns" as a const. It is used many times in the code.

cmd/flat-network-controller/app/controller.go
```diff
@@ -0,0 +1,247 @@
...
+	dns := corev1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      nameDNS,
+			Tenant:    net.Tenant,
+			Namespace: nsDNS,
+		},
+		Spec: corev1.ServiceSpec{
+			Ports: []corev1.ServicePort{
+				{
+					Name:       "dns",
```
chenqianfzh, line 182: "dns" -> "dns-udp"?

cmd/flat-network-controller/app/controller.go
```diff
@@ -0,0 +1,247 @@
...
+func (c *Controller) Enqueue(obj interface{}) {
+	net, ok := obj.(*v1.Network)
+	if !ok {
+		klog.Fatalf("got non-network object %v", obj)
+	}
+
+	c.queue.Add(genKey(net))
+}
+
+func genKey(net *v1.Network) string {
```
chenqianfzh, line 233: let's use cache.MetaNamespaceKeyFunc(), the systematic to generate cache key.

cmd/flat-network-controller/app/controller.go
```diff
@@ -0,0 +1,247 @@
...
+		},
+	}
+	svc, err := svcClient.CoreV1().ServicesWithMultiTenancy(nsDNS, net.Tenant).Create(&dns)
+	if err != nil {
+		if !errors.IsAlreadyExists(err) {
+			// todo: consider differing temporary errors from permanent errors
+			// todo: to fail the network provision in case of permanent errors
+			return nil, err
+		} else {
+			svc, err = svcClient.CoreV1().ServicesWithMultiTenancy(nsDNS, net.Tenant).Get(nameDNS, metav1.GetOptions{})
```
chenqianfzh, line 213: Why we still need to check the existence of the service object when the Create() reports the object already exists?

cmd/flat-network-controller/app/controller_test.go
```diff
@@ -0,0 +1,167 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	v1 "k8s.io/arktos-ext/pkg/apis/arktosextensions/v1"
+	"k8s.io/client-go/kubernetes/fake"
+	coremock "k8s.io/client-go/testing"
+	api "k8s.io/kubernetes/pkg/apis/core"
+
+	//"k8s.io/client-go/kubernetes/fake"
```
chenqianfzh, line 33: remove the commented line.

cmd/flat-network-controller/app/controller_test.go
```diff
@@ -0,0 +1,167 @@
...
+	fakearktosv1 "k8s.io/arktos-ext/pkg/generated/clientset/versioned/fake"
+
+	"testing"
```
chenqianfzh, line 36: reoder the package list to put the standard packages first.

cmd/flat-network-controller/app/controller_test.go
```diff
@@ -0,0 +1,167 @@
...
+	}{
+		{
+			desc: "happy path",
+			input: &v1.Network{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:   "test-ne",
+					Tenant: "test-te",
+				},
+				Spec: v1.NetworkSpec{
+					Type: "flat",
```
chenqianfzh, line 56: As mentioned before, we should have a const definition for "flat"

cmd/flat-network-controller/app/controller_test.go
```diff
@@ -0,0 +1,167 @@
...
+				},
+				Status: v1.NetworkStatus{},
+			},
+			svcResp: &corev1.Service{
+				ObjectMeta: metav1.ObjectMeta{},
+				Spec: corev1.ServiceSpec{
+					ClusterIP: "11.22.33.44",
+					Type:      "ClusterIP",
+				},
```
chenqianfzh, line 65: Use the const ServiceTypeClusterIP instead of "ClusterIP"

cmd/flat-network-controller/app/controller_test.go
```diff
@@ -0,0 +1,167 @@
...
+			},
+			svcResp: &corev1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Tenant:    "test-te",
+					Namespace: "kube-system",
+					Name:      "kube-dns-test-ne",
+				},
+				Spec: corev1.ServiceSpec{
+					ClusterIP: "11.22.33.44",
+					Type:      "ClusterIP",
```
chenqianfzh, line 106: All the test cases here are for "ClusterIP" type. Do we need some cases covering other types?

cmd/flat-network-controller/app/controller_test.go
```diff
@@ -0,0 +1,167 @@
...
+				Status: corev1.ServiceStatus{},
+			},
+			netRespError:   fmt.Errorf("fake test error: failed to  update network status"),
+			expectingError: true,
+		},
+	}
+
+	for _, tc := range tcs {
+		t.Run(tc.desc, func(t *testing.T) {
+			netClient := fakearktosv1.NewSimpleClientset(tc.input)
```
chenqianfzh, line 140: As we have a field called "desc" in test case, let's include it in the error log, for the convenience of debugging. 

For ex, change line 143 to :
t.Fatalf("%v: unexpected update", t.desc)

cmd/flat-network-controller/flat-network-controller.go
```diff
@@ -0,0 +1,88 @@
...
+	"k8s.io/arktos-ext/pkg/generated/informers/externalversions"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/clientcmd"
+	"k8s.io/klog"
+	"k8s.io/kubernetes/cmd/flat-network-controller/app"
+	"time"
+)
+
+const defaultWorks = 4
```
chenqianfzh, line 34: Nit: name of the variable, work or worker?

cmd/flat-network-controller/flat-network-controller.go
```diff
@@ -0,0 +1,88 @@
...
+	masterURL  string
+	kubeconfig string
+	workers	int
+)
+
+func main() {
+	klog.InitFlags(nil)
+	flag.Parse()
+	if workers <= 0 {
+		workers = defaultWorks
```
chenqianfzh, line 46: Do we need to set a cap on the number of workers?

cmd/flat-network-controller/app/controller.go
```diff
@@ -0,0 +1,247 @@
...
+
+	klog.V(5).Infof("processing network %s/%s", net.Tenant, net.Name)
+
+	if err := manageFlatNetwork(net, c.netClientset, c.svcClientset); err != nil {
+		c.queue.AddRateLimited(key)
+		c.recorder.Eventf(net, corev1.EventTypeWarning, "FailedProvision", "failed to provision network %s/%s/%s: %v", net.Tenant, net.Tenant, net.Name, err)
+		return
+	}
+
+	c.recorder.Eventf(net, corev1.EventTypeNormal, "SuccessfulProvision", "successfully provision network %s/%s/%s", net.Tenant, net.Tenant, net.Name)
```
chenqianfzh, line 145: "net.Tenant, net.Tenant" -> should be "net.Tenant, net.Namespace"

cmd/flat-network-controller/app/controller.go
```diff
@@ -0,0 +1,247 @@
...
+				},
+				{
+					Name:       "metrics",
+					Protocol:   "TCP",
+					Port:       9153,
+					TargetPort: intstr.FromInt(9153),
+				},
+			},
+			Selector: map[string]string{
+				"k8s-app": "kube-dns",
```
chenqianfzh, line 201: user const for "kube-dns" and "ClusterIP"

cmd/flat-network-controller/app/controller_test.go
```diff
@@ -0,0 +1,167 @@
...
+	coremock "k8s.io/client-go/testing"
+	api "k8s.io/kubernetes/pkg/apis/core"
+
+	//"k8s.io/client-go/kubernetes/fake"
+	fakearktosv1 "k8s.io/arktos-ext/pkg/generated/clientset/versioned/fake"
+
+	"testing"
+)
+
+func TestManageFlatNetwork(t *testing.T) {
```
chenqianfzh, line 39: consider replacing "test-ne" and "test-te" with some vars defined for test?

cmd/flat-network-controller/app/controller.go
h-w-chen: not all branches call queue.Forget()

cmd/flat-network-controller/app/controller.go
h-w-chen: 1. network object has no ns; will update the log message accordingly.
2. yes it is intentional for controller to have the loop trying to provision network eventually.

cmd/flat-network-controller/app/controller.go
h-w-chen: "dns" seems the proper name following the suit of k8s dns service.

cmd/flat-network-controller/app/controller.go
h-w-chen: better not to depend on internal mechanism. It is purposed as controller/resource specific way for custom keys, though coincidently very similar to that used in apiserver in this case..

cmd/flat-network-controller/app/controller_test.go
h-w-chen: subtest would automatically put desc (e.g. "dns service already exists") in test case summary like
```=== RUN   TestManageFlatNetwork/dns_service_already_exists```

cmd/flat-network-controller/flat-network-controller.go
h-w-chen: worker. will update.

cmd/flat-network-controller/flat-network-controller.go
h-w-chen: not sure whether we need to validate with the maximum number of workers allowed. As this controller is supposed to run as pod, we can limit its resource usage in k8s ways.

cmd/flat-network-controller/app/controller.go
h-w-chen: will arrange imports as suggested.

cmd/flat-network-controller/app/controller.go
h-w-chen: will change to queueIsEmpty

cmd/flat-network-controller/app/controller.go
h-w-chen: will correct in comment. process(item), if read method together with parameter, it is process item already.

cmd/flat-network-controller/app/controller.go
h-w-chen: flat and other types are free format strings, not a keyword by network type definition. 
flat does have significance in flat network controller, though. will declare a const in controller code.

cmd/flat-network-controller/app/controller.go
h-w-chen: will refer to such const as suggested

cmd/flat-network-controller/app/controller.go
h-w-chen: will update as suggested

cmd/flat-network-controller/app/controller.go
h-w-chen: we need to get the service IP address

cmd/flat-network-controller/app/controller_test.go
h-w-chen: will remove as suggested

cmd/flat-network-controller/app/controller_test.go
h-w-chen: will update

cmd/flat-network-controller/app/controller_test.go
h-w-chen: in test code I'd rather not refer to consts declared in external packages in test data; in this way, it is easier to understand.

cmd/flat-network-controller/app/controller_test.go
h-w-chen: test data using plain value instead of reference to external const seems easier to understand.

cmd/flat-network-controller/app/controller.go
h-w-chen: should only have tenant + name for network; will remove the extra tenant fro message.

cmd/flat-network-controller/app/controller_test.go
h-w-chen: test-ne & test-te are vars defined for test only.

cmd/flat-network-controller/app/controller.go
h-w-chen: updated as suggested

PR #325, opened 2020-05-29
revert empty tenant workaround

chenqianfzh:


staging/src/k8s.io/apiserver/pkg/endpoints/filters/tenantinfo.go
```diff
@@ -47,17 +47,12 @@ func GetTenantFromContext(req *http.Request) (string, error) {
 	ctx := req.Context()
 	requestor, exists := request.UserFrom(ctx)
 	if !exists {
-		return "", errors.New("The user info is missing.")
+		return "", fmt.Errorf("The user info is missing.")
```
Sindica, line 50: What's the difference bettwen errors.New and fmt.Errorf here? fmt.Errorf takes additional argument but it is not used here.

chenqianfzh: The golang doc said they are the same except they belong to different package. I use fmt to be consistent with other error report code in this file.

Besides, I udpated the line as suggested.

PR #341, opened 2020-06-03
design doc: pod annotations as hint & cni args

h-w-chen:
**What type of PR is this?**
/kind design

proposal to use pod annotations as scheduling/network-resource hint & cni args. The network controller of the interesting network type is able to hook in the pod life cycle to prepare network resource and pass data to cni plugin at later time on cni invocation.

| annoation | purpose |
| --- | --- |
| arktos.futurewei.com/nics | user specified subnet & ip address; optional |
| arktos.futurewei.com/network-readiness | optional; if it is there and not true, scheduler should not schedule this pod |
| arktos.futurewei.com/cni-args | optional; usually present for vpc cni plugin; its value will be passed to cni as CNI_ARGS |


h-w-chen:
premature; need to reconcile with spec.nic of pod.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -92,40 +94,41 @@ spec:
     ports:
       - containerPort: 443
 ```
-When a pod is attached to a certain network, it needs to set its "network" field:
+When a pod is attached to a certain network, it needs to set its "network" in annotations:
```
chenqianfzh, line 97: you mean label?

h-w-chen: yes; thanks for pointing it out. will update

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -92,40 +94,41 @@ spec:
...
 
 ```yaml
 apiVersion: v1
 kind: Pod
 metadata:
   name: nginx
+  labels:
+    arktos.futurewei.com/network: vpc-1
+  annotations:
+    arktos.futurewei.com/nics: [{"subnet": "subnet-1", "ip": "192.168.0.12"}]
```
zasherif, line 125: I don't understand the role of nics hint. How can we make sure that they provide a concrete contract? Can you give example.

Why don't we make it even more generic and allow a network provider user input annotation (instead of the label "arktos.futurewei.com/network". For example 

User Input:
```
arktos.futurewei.com/network_provider : "Mizar"
arktos.futurewei.com/network_provider_user_parameters: "{"vpc": vpc-1, "network":  net-1}" 
```

Result (after Mizar Pod's operator):

```
arktos.futurewei.com/network-readiness=true
arktos.futurewei.com/network_provider : "Mizar"
arktos.futurewei.com/network_provider_user_parameters: "{"vpc": vpc-1, "network":  net-1}" 
arktos.futurewei.com/cni_args: "{"IPv4": 10.0.0.3, "gw":  '10.0.0.1', 'Bouncers': [B1, B2, B3]}" 
```
And then you aggregate both cni_agrs and network_provider_user_parameters as args to CNI. The same model can go with services (actually compatible to what services currently do). Right? Can you add a similar example for services.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -268,11 +271,19 @@ kube-dns-{network} shall be managed by Endpoints controller just like a regular
...
+## CNI spec
+CNI_ARGS is leveraged for Arktos workload runtime to pass information on to the cni plugin of the network type.
+
+By spec the value of CNI_ARGS env var is alphanumeric key-value pairs separated by semicolons, e.g. "IPV4=12.34.56.78;VPCID=mizar-12345".
+
+Arktos workload runtime will put whatever is the pod annotation of "arktos.futurewei.com/cni-args" to CNI_ARGS as defined in cni spec.
+
+The network controller of the vpc network type is typically responsible for annotating pods properly with the relevant information.
+
+The alternative is
```
zasherif, line 286: What are other alternatives?

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -292,3 +303,12 @@ Components not decided yet at current phase are not included in these views.
...
+
+### Activity Flow: Pod - from creation to running
+
+![Pod-activity-diag](images/pod-activity-diagram.png)
+
+1. It is strongly recommended to annotate new pods with arktos.futurewei.com/network-readiness as hint for scheduler. It is not mandatory, though - kubelet and cni plugin will have to cope with it and gets it right eventually. This kind of initial pod annotation can be done by a custom admission control.
+2. It is the network controller's responsibility to do whatever network resource preparation for the pod; if no special interests in it, it is fine. The bottom line is the controller needs to clean the annotation arktos.futurewei.com/network-readiness=false if it is there. If pod has hint of arktos.futurewei.com/nics annoation, it should honor that.
+3. It is the network controller's responsibility to clean the annotation arktos.futurewei.com/network-readiness=false if it is there.
+4. It is the network controller's responsibility to annotate pod with proper information that the cni plugin needs in term of CNI-ARGS. Arktos system just picks up the annotation value and pass it on when calling cni plugin.
```
zasherif, line 314: I generally like this approach because it gives us a lot of decoupling. My question is that; Mizar behavior here is Arktos specific. Is there a way for us to have the same model in community Kubernetes? I'm trying to avoid having a CNI/Operator behavior for arktos that is significantly different from a regular K8s implementation. Do you have any suggestions here?

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -268,11 +271,19 @@ kube-dns-{network} shall be managed by Endpoints controller just like a regular
...
 
-
-
 ### Ingress/egress/network policies
 (**TBD: more details required**)
 
+## CNI spec
+CNI_ARGS is leveraged for Arktos workload runtime to pass information on to the cni plugin of the network type.
+
+By spec the value of CNI_ARGS env var is alphanumeric key-value pairs separated by semicolons, e.g. "IPV4=12.34.56.78;VPCID=mizar-12345".
```
zasherif, line 280: Just curious; Is there a mechanism to update CNI_ARGS (after provisioning the PODs)? If not that's fine, I just need to know.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: this incomplete line should be removed, as there is no good alternative yet.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: cni invocation a one-time call and cni process runs and exit. 
k8s does not handle annotation changes in our case and no CNI_ARGS  updates would be detected, hence no mechanism to automate the pod re-creation. If we do think such a mechanism is critical and we should support,  we need to put such info in spec fields instead of annotations.
manual stopping pod and starting pod will bring pod with the new cni-args; if such mechanism exist, it is essentially a similar restarting.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: If we community k8s able to handle following properly, then yes
1. ask scheduler temporarily put pod on hold before network controller finishes its preparation part;
2. kubelet/container runtime able to pass applicanle ani-args

for 1, either we __hack__ around with taint/toleration - which does not use network-readiness annotation, or put a custom controller which understands the annotation.
for 2, the only way I can think about is modify container runtime, as k8s has no established way to pass extra arg, though cni spec does include in its specification.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: 1. for the nics annotation, it carries the pod spec.nics field in form of annotation. More detail about spec.nics can be found [vpc-network-design](https://github.com/futurewei-cloud/arktos/blob/master/docs/design-proposals/network/NICAndVPCSupportInArktos.md). We hope to keep the essences of nics, including the ability to have multi nics.  I'll update the example which includes nic name, hopefully it is more understandable.
2. Though nics annotation able to define intent of multiple nics, Mizar can start with single nic support.
3. I am not inclined to aggregate cni-args and user-parameters, as they serves different purpose. user-parameters is user inputs, expressing user's intent of network resource; cni-args is what the controller intends to pass on to cni plugin later as _call-back_ data. Better not mix them.
4. network-provider=Mizar seems unnecessary, as pod has network label associating the proper network object which has type (mizar in this case). Such information is consumed primarily by network controller, which should have network objects in memory already, easy to look it up. However, if this is not the case (controller does not keep a lookup table of network objects), explicit putting network type and vpc id in pod could make sense for performance sake.

h-w-chen: For the case of service, it has network label on creation for sure. 
If its virtual ip address is to be decided by external network provider (not sure Mizar is this kind), then the network provider can decide which hint it would like to have (nics annotation seems improper) and what sematic (like always honor the hint or OK if unsatisfied). I'll put a recommendation based on the user-input-param idea in doc; however, network provider can choose its own format if it feels like that.
If its vip is decided by Arktos, Arktos will assign vip, likely based on hints (or not, if we have no resource to impl at this stage).

h-w-chen: doc updated with addition of arktos.futurewei.com/vip-hint annotation.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -92,40 +94,41 @@ spec:
...
 
 ```yaml
 apiVersion: v1
 kind: Pod
 metadata:
   name: nginx
+  labels:
+    arktos.futurewei.com/network: vpc-1
+  annotations:
+    arktos.futurewei.com/nics: [{"name": "eth0", "ip": "192.168.0.12"}]
```
clu2xlu, line 125: Since nics only specify the ethernet information here, how is subnet name gets passed to network provider? 
The pod is associated with Subnets, not VPC. Later in the doc, you also mentioned a CIDR range, is that the subnet's CIDR range?

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -92,40 +94,41 @@ spec:
...
 spec:
   containers:
   - name: nginx
     image: nginx
     ports:
       - containerPort: 443
-  network: vpc-1
-  nics:
-    - subnet: subnet-1
-      ip: 192.168.0.12
```
clu2xlu, line 128: Since these spec is removed, Mizar needs to have a way to know the subnet this pod belongs to.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -238,6 +241,17 @@ Network object specifies service IPAM as *external*:
       service:
         ipam: external
 
+#### User hints
+
+User can provide hints about service IP address.
+
+For Arktos managed pool, it will look at the specific annotation of service object like below
+```arktos.futurewei.com/vip-hint: {"cidr":"10.10.1.0/26"}```
```
clu2xlu, line 249: Is this the CIDR range for VPC or subnet? 
Passing in a CIDR range is not very helpful, as two VPCs or two subnets can have the same CIDR. We should track using name of the subnet or vpc.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -238,6 +241,17 @@ Network object specifies service IPAM as *external*:
...
+
+Other keys supported are "ip", "policy" (with "BestEffort" the default, or "Guaranteed").
```
clu2xlu, line 251: How is ip gets passed? ```arktos.futurewei.com/vip-hint: {"cidr":"10.10.1.0/26", "ip": "10.10.1.8", "policy": "BestEffort"}``` Like this?

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: this exemplary line only reflects subset that nics could have.  please refer to [previous network extension design](https://github.com/futurewei-cloud/arktos/blob/master/docs/design-proposals/network/NICAndVPCSupportInArktos.md), pod spec change section to see what others may be supported. One valid example is {"name":"eth0", "subnet":"my-subnet-name"}. What in this nics annotation is intended for user to specify some finer constraint beyond the vpc defined in network object; nics info is optional.
From Arktos perspective, pod associated to an IP address. For vpc typed network, of course it must be on a VPC network.

As for the cidr (in vip-hint annotation, and network.spec), it is for service objects, not pod objects. This is extended from cluster-ip-range of k8s parameter, which is to define the range of service virtual IP addresses.

Please be noted nics annotation does not mandate subnet cannot have value in cidr form; if it makes sense, it can has cidr value instead of subnet name. It depends on what external network provider understands.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen:
```annotations:
￼     arktos.futurewei.com/nics: [{"name": "eth0", "subnet": "subnet-1", "ip":"192.168.0.12"}]```
like above example, user can provide subnet in nics annotation.

File: docs/design-proposals/multi-tenancy/multi-tenancy-network.md (hunk 8 above)
Comment (h-w-chen) [reply docs/design-proposals/multi-tenancy/multi-tenancy-network.md:249]: this cidr in vip-hint is for Arktos service objects only. Using vip-hint to convey optional service intent is optional. If Mizar don't take cidr or ip hint, it is totally fine to ignore. Not to implement the Guaranteed  policy semantics is OK, too. We can put it explicit that first version only supports besteffort.

File: docs/design-proposals/multi-tenancy/multi-tenancy-network.md (hunk 9 above)
Comment (h-w-chen) [reply docs/design-proposals/multi-tenancy/multi-tenancy-network.md:251]: reasonable guess :) yes. If ip is specified, no need to have cidr; but this does not matter too much for the topic.

File: docs/design-proposals/multi-tenancy/multi-tenancy-network.md (hunk 2 above)
Comment (clu2xlu) [reply docs/design-proposals/multi-tenancy/multi-tenancy-network.md]: With given info, I came up with a solution: 

```

in common.k8s.py 

    def get_pod_annotations():
        url = ("%s/api/v1/namespaces/%s/pods/%s" % (server, namespace, pod))
        response = requests.get(url, headers=headers)
        json_response = response.json()
        annotations = json_response['metadata'].get('annotations')

        return annotations
```

```

import common.k8s

def cni_add: 
    annotations = k8s.get_pod_annotations() //Kubernetes API server
    label = k8s.get_pod_label()   //Kubernetes API server

    ip = annotations['ip_address']
    subnet = annotations['subnet']
    vpc = label['arktos.futurewei.com/network']
    ep = Endpoint(name, self.obj_api, CniService.store)
    ep.set_ip(ip)
    ep.set_vpc(vpc)
    ep.set_net(subnet)
    ...

    cni_config = {
        "cniVersion": params.cni_version,
        "interfaces": [
            {
                "name": ep.veth_name,
                "mac": ep.mac,
                "sandbox": ep.netns
            }
        ],
        "ips": [
            {
                "version": "4",
                "address": "{}/{}".format(ep.ip, ep.prefix),
                "gateway": ep.gw,
                "interface": 0
            }
        ]
    }
```

However, the major concerns with this approach is that it's depeneding on k8s API, which we really want avoid. 
We are hoping to have network info dynamically inserted into the CNI network configuration that is passed to CNI plugin binaries via stdin.
This way we can minimize dependances on k8s api.

File: docs/design-proposals/multi-tenancy/multi-tenancy-network.md [hunk 10]
```diff
@@ -70,7 +72,7 @@ spec:
 And here is a sample of a VPC network:
 
 ```yaml
-apiVersion: v1
+apiVersion: arktos.futurewei.com/v1
 kind: Network
 metadata:
   name: vpc-1
```
clu2xlu, line 78: The vpc-1 is the network name for arktos network object OR Mizar vpc object?

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
clu2xlu: Another approach would be: 
1. Pod operator kicks off and create a pod object. 
2. Retrieve values from annotation. 
3. CNI gets triggered, and create network interface. 
4. Pod operator updates pod spec. 
5. Arktos gets new ip from pod's spec.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
zasherif: I now understand the reason you have the nice. In this case, for PODS, the NICS is the one we really care about. So doesn't it make sense then that all the network specific fields goes to the NIC. For example, a POD can have multiple NICs each belong to a different VPC/Subnet? Also the IP address allocation on Mizar side, will be NIC specific not POD specific. 

Also, how the user will be able to add/remove NICs from the POD. How are you going to (re)-invoke the CNI workflow then?

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
zasherif: So you will keep the NIC arguments unspecified?

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
zasherif: I think it is necessary. Here are two scenarios in which we need to update (re)-invoke the CNI some how:
1. In Mizar we update the list of bouncers (transit switch) to the host.
1. The user add/remove NICs from a POD.

In Mizar, we can do these updates with RPC calls to the transit daemon. But that's something Mizar specific. If CNI and runtime workflow is not capable of doing these updates, then how this solution supports different network provider?

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
zasherif: @h-w-chen If you agree of the purpose of "arktos.futurewei.com/network_provider_user_parameters" annotations for users to pass network specific configuration, how can we separate them from cni_args annotations. My understanding is that annotations remain with the object!. Or you suggest to remove the user parameters annotations after the Network Controller processes the object and marks it network_ready?

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
clu2xlu: Correct me if I am wrong: Mizar needs to launch a pod with IP address that is within CIDR 10.10.1.0/26 range. 
However, the subnet field is optional. From pure network perspective, a pod needs to be launched within a subnet. So if we allow using to specify a CIDR range that does not reference to an existing subnet, we might assign an IP address that does not fall into range provided by the user here (in this case is 10.10.1.0/26) 
For example, if vpc-1 has only one subnet with CIDR 10.10.2.0/24 and user assigned a vip-hint of cidr = 10.10.1.0/26. 
Mizar only reads the subnet's CIDR range and pick an IP address from 10.10.2.0/24 pool. 
Will this cause an error?

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
clu2xlu: Interesting concept here, with possibly multiple network interfaces linked to a pod and each network interface can be in different subnet. If IP address allocation is NIC specific, we can add a new Nic type object in Mizar, and after Nic object is provisioned, it then will be associated with a pod.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: @zasherif 
network controller doesn't have to remove the user input annotation; leave it there is fine.
cni-args annotation is for cni usage only; since network controller and cni plugin are for same network type, controller knows what kind of data the cni plugin may need. The user input annotation, is for network controller to consume. 
yes, network controller does need to mark the pod as network-ready.

h-w-chen: @clu2xlu CNI_ARGS env will have whatever in pod's cni-args annotation; cni-args annotation is assigned by network controller (in your case, mizar controller). What value should be put there is totally up to mizar requirements; Arktos system only pass this data to cni plugin when cni plugin is being invoked.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: based on previous design, agreed-upon keys include name, subnet, ip, portid. I'll call it out explicitly in the updated design doc.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: it seems too much for us to implement multiple nics right now. Let's focus on single nic right now.
I'll update design to change nics to nic only to avoid confusion.
If we decide to support multiple nics in the future, we can pick up annotation nics by that time.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: it is Arktos network object name

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: @zasherif  I've change the design to allow only one nic each pod. Let's focus on single nic pod right now, defer the consideration of multiple nics support to later.

For reason 1, Mizar updates the list of bouncer internally, will the nic already allocated be affected? will its ip address change?  Can we make _encapsulate_ Mizar change internal to Mizar and not affect the nic that has already been allocated to pod?

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: in that case, mark network-readiness to error,  as it cannot be satisfied. Meanwhile, please fire an event to Arktos reporting the error.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: exchanged emails offline. It turns out a Mizar operator will do (most of the) underlying network wirings based on  network_user_input annotation of pod; its cni is mere a synchronization mechanism to wait for the network to be provisioned. Arktos won't prevent such interaction sequence, with current design. Such interactions would be primarily driven by Mizar network controller itself.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -36,6 +36,8 @@ Pods in a network __should__ be able to access API server nodes by their IP addr
 ### Network Object
 A network object contains its type, and type-specific configurations. It's a not namespace-scoped object, so that one network object can be shared by applications in multiple namespaces in a same space.
 
+It is defined as CRD by [crd-network.yaml](../../../pkg/controller/artifacts/crd-network.yaml)
+
 The "type" field is the only mandatory field in a network object. For now there are two types defined: flat and vpc.
```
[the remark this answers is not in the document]
h-w-chen: yes the type should be the network-provider, as you pointed out. will update doc shortly.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -92,42 +94,52 @@ spec:
...
+Annotation arktos.futurewei.com/nic is for user to provide optional information about pod nic. The recognized keys of element include:
+
+|key|exemplary value|
+|---:|---:|
+|name|eth0|
+|subnet|subnet-1, or 192.168.100.0/26|
+|ip|192.168.100.5|
+
+
 If these settings are set on a pod attached to a flat network, the settings will be ignored by the flat network controller and also the corresponding CNI plugins.
```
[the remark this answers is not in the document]
h-w-chen: this is to state that FLAT typed network will ignore nic hint user may have. I will update doc a bit to include explain the reason.
For vpc typed network providers like mizar & neutron, we expect them to implement the support. If not the case, please call it out explicitly.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -238,6 +250,17 @@ Network object specifies service IPAM as *external*:
...
+#### User hints
+
+User can provide hints about service IP address.
+
+For Arktos managed pool, it will look at the specific annotation of service object like below
+```arktos.futurewei.com/vip-hint: {"cidr":"10.10.1.0/26"}```
+
+Other keys supported are "ip", "policy" (with "BestEffort" the default, or "Guaranteed").
+
+For network provider managed pool, it is encouraged to stick to the above convention. However, it is totally up to the provider service controller.
```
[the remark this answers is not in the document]
h-w-chen: service associates to pods using labels, __particularly the network label will always be included__.
service objects in arktos in essence almost the same as in k8s, except for the network label implicitly being enforced.
doc will be updated to include a service definition which has network label.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -92,43 +94,52 @@ spec:
...
 ```
 
-If these settings are set on a pod attached to a flat network, the settings will be ignored by the flat network controller and also the corresponding CNI plugins. 
+Annotation arktos.futurewei.com/nic is for user to provide optional information about pod nic. The recognized keys of element include:
+
+|key|exemplary value|
+|---:|---:|
+|name|eth0|
+|subnet|subnet-1, or 192.168.100.0/26|
+|ip|192.168.100.5|
```
XiaoningDing, line 140: Do we need to support portID?

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -238,6 +250,17 @@ Network object specifies service IPAM as *external*:
       service:
         ipam: external
 
+#### User hints
+
+User can provide hints about service IP address.
+
+For Arktos managed pool, it will look at the specific annotation of service object like below
+```arktos.futurewei.com/vip-hint: {"cidr":"10.10.1.0/26"}```
```
XiaoningDing, line 258: Got lost.  I was under the impression that this was an optional field in network object, according to our previous discussion.  What's changed and why does this need to be set on every service object?

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -292,3 +322,12 @@ Components not decided yet at current phase are not included in these views.
 ### Key Scenario: tenant & DNS provisioning
 
 ![tenant & DNS provision](images/network-scenario-tenant-init.png)
+
+### Activity Flow: Pod - from creation to running
+
+![Pod-activity-diag](images/pod-activity-diagram.png)
```
XiaoningDing, line 328: A couple of questions about the digram:  1) Should the last column mean agent, instead of scheduler?  2) From the digram it seems the API Server have knowledge of flat network, it's a little strange.  The API Server shouldn't care about the network provider type, right?

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -292,3 +322,12 @@ Components not decided yet at current phase are not included in these views.
...
+
+1. It is strongly recommended to annotate new pods with arktos.futurewei.com/network-readiness as hint for scheduler. It is not mandatory, though - kubelet and cni plugin will have to cope with it and gets it right eventually. This kind of initial pod annotation can be done by a custom admission control.
```
XiaoningDing, line 330: If the network-readiness annotation is not mandatory, how does the node agent know when it's ready to call CNI agents?

XiaoningDing, line 330: Not sure if this "network-readiness" annotation get couple with scheduling.  Some network providers may need to know the scheduling decision before it can initialize pod network.  So there could be a deadlock. (I know flannel is not a network provider type here, but Flannel is an example that it allocates IP based on node selection.)

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -169,15 +168,16 @@ The reasons that DNS service is per-network instead of per-space are:
 
 #### Type definition
 
-Addition of service spec is spec.network field. By default, it is the default network of the tenant. 
+Service objects associate to network via label arktos.futurewei.com/network. By default, network is the default network of the tenant.
```
clu2xlu, line 171: In mizar, service acts like a load balancer, we call it scaled endpoint.
Will we be provisioning a default number of backends for this service/scaled ep?
Will we allow the user to specify how many backends they desire for their service/scaled ep?

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
clu2xlu, line 258: Shall we include this vip-hint when creating service instead? Mizar will assign an ip address to a service, and the ip address needs to fall within the range of vip-hint. correct?

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
clu2xlu: We need also make sure that there is at least one subnet within any of the vpc listed under vpcID field! Otherwise, if user does not specify a subnet, we cannot create the pod as the pod needs to be associated and only associated with a subnet.

h-w-chen: subnets's creation should be totally managed by network provider. user does not have to specify the subnet - it is optional as hint if present, not mandatory. Hope that makes sense.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: portID seems neutron specific cni-args. It could be annotated by neutron operator as ani-args, and neutron cni plugin will take podID to wire up nic. 
in previous design, portID and others are all in pod.spec, which serve for two purposes: user hint, extra cni-arg. In this design, it uses two annotations in attempt to separate them.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: network objects has a list of cidrs for service ip addresses.
Here is for individual service object - if user would like to provide some hints(constraints). It seems implementing such user hints is not of high priority; I'll put them in to TBD, deferring for later decision.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: 1. yes it should be agent in last column;
2. putting flat type check in diagram is improper; I'll update the diagram shortly

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
clu2xlu: Yeah, that's fine. My point is if vpc listed does not have a subnet, Mizar will throw an error saying no subnet in vpc. Does that make sense?

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: kubelet will try to start pod after the pod is scheduled to the node mostly as current process, with addition of ensuring network status is ready associated to the pod. In case pod cannot gets its nic setup successfully, kubelet will try it later.
Absence of network-readiness is allowed for backward compatibility, particularly for flat network using regular cni plugins. Ideally pod with presence of such makes the workflow more efficient, but w/o it pod is still able to start, eventually. However, if a certain network cannot work in that relaxed way, then such network-readiness annotation is required for that type of network - in that case scheduler will ensure pod won't be scheduled until network operator has made necessary preparations.

h-w-chen: this network-readiness is for scheduler to go after network controller has done necessary preparation, which unblock scheduling. If network controller needs to do some post-scheduled task, it can still go by watching pod spec.nodeName and do whatever it needs to.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: how many backends are determined by the number of endpoints in arktos(k8s). That number could be dynamic. please make sure mizar is able to handle that case.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: based on our discussion a few days ago, mizar network should exists __before__ arktos network object is created. I think mizar network should include valid subnet in place.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
h-w-chen: sorry this had caused some confusion. 
I've remove the vip-hint at this stage; so mizar can assign vip decided by itself.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
clu2xlu: Customer needs to specify the subnet CIDR range, thus customer **has to** create a subnet manually. Mizar can't decide for customer subnet's CIDR.

h-w-chen: it is Okay that user has to ensure mizar vpc + subnet created before hand.
creating mizar vpc & subnets inside vpc is out of the design proposal's scope.

deepak-vij:
@XiaoningDing, I have provided my comments/feedback for this.

deepak-vij:
@XiaoningDing, for some reason I am not able to locate the following “Naming Problem & Solution” section using “source diff” view. But do see it in the “rich diff” view. In any case, here is the reworded text for this particular section.

Naming Problem & Solution
Services, as one of the critical network related resource types, are within the network boundary. There is no cluster wide shared services for a tenant, as both the virtual IP address and the associated endpoints are all scoped within a single network.

XiaoningDing:
Thank @deepak-vij for helping improve the document. @h-w-chen

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -92,43 +94,52 @@ spec:
     ports:
       - containerPort: 443
 ```
-When a pod is attached to a certain network, it needs to set its "network" field:
+When a pod is attached to a certain network, it needs to set its "network" in labels:
```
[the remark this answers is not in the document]
h-w-chen: corrected as suggested.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -36,7 +36,9 @@ Pods in a network __should__ be able to access API server nodes by their IP addr
 ### Network Object
 A network object contains its type, and type-specific configurations. It's a not namespace-scoped object, so that one network object can be shared by applications in multiple namespaces in a same space.
 
-The "type" field is the only mandatory field in a network object. For now there are two types defined: flat and vpc.
+It is defined as CRD by [crd-network.yaml](../../../pkg/controller/artifacts/crd-network.yaml)
```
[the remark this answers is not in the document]
h-w-chen: corrected as suggested.

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -36,7 +36,9 @@ Pods in a network __should__ be able to access API server nodes by their IP addr
 ### Network Object
 A network object contains its type, and type-specific configurations. It's a not namespace-scoped object, so that one network object can be shared by applications in multiple namespaces in a same space.
 
-The "type" field is the only mandatory field in a network object. For now there are two types defined: flat and vpc.
```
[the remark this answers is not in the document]
h-w-chen: updated as suggested

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -157,15 +168,16 @@ The reasons that DNS service is per-network instead of per-space are:
 
 #### Type definition
 
-Addition of service spec is spec.network field. By default, it is the default network of the tenant. 
+Service objects associate to network via label arktos.futurewei.com/network. By default, network is the default network of the tenant.
```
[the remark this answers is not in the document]
h-w-chen: corrected as suggested

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -92,43 +94,52 @@ spec:
...
   - name: nginx
     image: nginx
     ports:
       - containerPort: 443
-  network: vpc-1
 ```
 
 
-When a pod is attached to a certain network and it wants to specify subnet or IP, it needs to be specified in pod fields:
+When a pod is attached to a certain network and it wants to specify subnet or IP, it can be expressed in annotations:
```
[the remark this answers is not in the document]
h-w-chen: corrected as suggested

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -92,43 +94,52 @@ spec:
...
     ports:
       - containerPort: 443
-  network: vpc-1
-  nics:
-    - subnet: subnet-1
-      ip: 192.168.0.12
 ```
 
-If these settings are set on a pod attached to a flat network, the settings will be ignored by the flat network controller and also the corresponding CNI plugins. 
+Annotation arktos.futurewei.com/nic is for user to provide optional information about pod nic. The recognized keys of element include:
```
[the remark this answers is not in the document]
h-w-chen: updated as suggested

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -268,11 +282,18 @@ kube-dns-{network} shall be managed by Endpoints controller just like a regular
...
 
-
-
 ### Ingress/egress/network policies
 (**TBD: more details required**)
 
+## CNI spec
+CNI_ARGS is leveraged for Arktos workload runtime to pass information on to the cni plugin of the network type.
+
+By spec the value of CNI_ARGS env var is alphanumeric key-value pairs separated by semicolons, e.g. "IPV4=12.34.56.78;VPCID=mizar-12345".
```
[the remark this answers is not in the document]
h-w-chen: reworded as suggested

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -268,11 +282,18 @@ kube-dns-{network} shall be managed by Endpoints controller just like a regular
...
+
+Arktos workload runtime will put whatever is the pod annotation of "arktos.futurewei.com/cni-args" to CNI_ARGS as defined in cni spec.
+
+The network controller of the vpc network type is typically responsible for annotating pods properly with the relevant information.
+
 ## Architectural Views
 
 Components not decided yet at current phase are not included in these views.
```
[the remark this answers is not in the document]
h-w-chen: updated as suggested

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -292,3 +313,13 @@ Components not decided yet at current phase are not included in these views.
 ### Key Scenario: tenant & DNS provisioning
 
 ![tenant & DNS provision](images/network-scenario-tenant-init.png)
+
+### Activity Flow: Pod - from creation to running
```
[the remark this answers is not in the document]
h-w-chen: updated as suggested

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -292,3 +313,13 @@ Components not decided yet at current phase are not included in these views.
...
+
+![Pod-activity-diag](images/pod-activity-diagram.png)
+
+1. On creation to annotate new pods with arktos.futurewei.com/network-readiness=false as hint for scheduler, by a custom admission control.
```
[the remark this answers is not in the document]
h-w-chen: updated as suggested

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -292,3 +313,13 @@ Components not decided yet at current phase are not included in these views.
...
+   For flat network type, such annotation is not required as its operator typically has no network preparation work to do.
+2. It is the network controller's responsibility to do whatever network resource preparation for the pod. If pod has hint of arktos.futurewei.com/nic annoation, it should honor that.
+3. It is the network controller's responsibility to clean the annotation arktos.futurewei.com/network-readiness=false if it is there, if applicable network resource preparation is finished.
```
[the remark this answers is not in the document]
h-w-chen: updated as suggested

docs/design-proposals/multi-tenancy/multi-tenancy-network.md
```diff
@@ -292,3 +313,13 @@ Components not decided yet at current phase are not included in these views.
...
+4. It is the network controller's responsibility to annotate pod with proper information that the cni plugin needs in term of CNI-ARGS. Arktos system just picks up the annotation value and pass it on when calling cni plugin.
```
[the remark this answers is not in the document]
h-w-chen: corrected as suggested

Issue #289, opened 2020-05-16
Handle CRI not returning limit info, add E2E tests for Pod Vertical Scaling


vinaykul:
**What would you like to be added**:
1. CRI may return partial or no limit information. Investigate how it can be handled by kubelet.
2. Resize policies implementation.
3. Handle multiple containers up/down resizing for cpu/memory.
4. E2E tests for Pod Vertical Scaling

**Why is this needed**:

PR #343, opened 2020-06-03
In-place Pod Vertical Scaling for Container Pods - core KEP implement…

vinaykul:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

**What this PR does / why we need it**: This is the core implementation portion of In-Place Pod Vertical Scaling for Container Pods refactored and with some failure scenario handling issues fixed.
Resource-quota, limit range, and mult-itenancy supprot is still being tested, and will send PR in a couple of days.

Fixes #

Code in kubelet handles updating resourcesAllocated field in PodSpec in response to resize request.
Code in kuberuntime_manager.go implements the container limit update. Please review for edge-cases, failure scenarios. I have already tested some corner cases.

Manual E2E tests done:
- Single container guaranteed QoS pod - memory and cpu increase & decrease
- Single container burstable Q0S pod - memory, cpu increase & decrease, set memory and cpu limits where previously there was none.
- Best effort QoS pod - verify no default resize policy is set, cannot add requests or limits
- Pod with 3 containers c1, c2, c3, and c2 memory has RestartContainer policy - verify no restart when updating multiple containers in a single shot if c2 memory is not updated. verify c2 restart if c2 memory and cpu are updated.
- Increase cpu and memory - verify that pod level cgroup limits are updated first.
- decrease cpu and memory - verify pod level cgroup limits are updated in the very end.
- error injection during multi-container pod limit update - verify that syncpod loop retries, status is not updated until success.
- error injection in resourcesAllocated patch from kubelet - verify SyncPod keeps retrying.
- add new container - verify that podresourceallocation plugin rejects it.

```release-note

```
Documentation: 
KEP: /keps/sig-node/20181106-in-place-update-of-pod-resources.md


pkg/kubelet/cm/cgroup_manager_linux.go
```diff
@@ -591,3 +594,102 @@ func (m *cgroupManagerImpl) GetResourceStats(name CgroupName) (*ResourceStats, e
...
+}
+
+// Set the memory limit in bytes applied to the cgroup
+func (m *cgroupManagerImpl) SetCgroupMemoryLimit(name CgroupName, memoryLimit int64) error {
+	cgroupPaths := m.buildCgroupPaths(name)
+	cgroupMemoryPath, found := cgroupPaths["memory"]
+	if !found {
+		return fmt.Errorf("failed to build memory cgroup fs path for cgroup %v", name)
+	}
+	memLimit := strconv.FormatInt(memoryLimit, 10)
```
pdgetrf, line 661: why 10?

pkg/kubelet/cm/cgroup_manager_linux.go
```diff
@@ -591,3 +594,102 @@ func (m *cgroupManagerImpl) GetResourceStats(name CgroupName) (*ResourceStats, e
...
+// Set the cpu quota, cpu period, and cpu shares applied to the cgroup
+func (m *cgroupManagerImpl) SetCgroupCpuLimit(name CgroupName, cpuQuota *int64, cpuPeriod, cpuShares *uint64) error {
+	var cpuQuotaStr, cpuPeriodStr, cpuSharesStr string
+	cgroupPaths := m.buildCgroupPaths(name)
+	cgroupCpuPath, found := cgroupPaths["cpu"]
+	if !found {
+		return fmt.Errorf("failed to build cpu cgroup fs path for cgroup %v", name)
+	}
+	if cpuQuota != nil {
+		cpuQuotaStr = strconv.FormatInt(*cpuQuota, 10)
```
pdgetrf, line 677: please take 10 out as a constant variable to be shared

pkg/kubelet/cm/pod_container_manager_linux.go
```diff
@@ -126,6 +127,26 @@ func (m *podContainerManagerImpl) GetPodContainerName(pod *v1.Pod) (CgroupName,
 	return cgroupName, cgroupfsName
 }
 
+func (m *podContainerManagerImpl) GetPodCgroupMemoryLimit(pod *v1.Pod) (uint64, error) {
+	podCgroupName, _ := m.GetPodContainerName(pod)
+	return m.cgroupManager.GetCgroupMemoryLimit(podCgroupName)
+}
+
+func (m *podContainerManagerImpl) GetPodCgroupCpuLimit(pod *v1.Pod) (int64, uint64, uint64, error) {
+	podCgroupName, _ := m.GetPodContainerName(pod)
```
pdgetrf, line 136: error checking?

pkg/kubelet/cm/pod_container_manager_linux.go
```diff
@@ -324,3 +345,19 @@ func (m *podContainerManagerNoop) GetAllPodsFromCgroups() (map[types.UID]CgroupN
 func (m *podContainerManagerNoop) IsPodCgroup(cgroupfs string) (bool, types.UID) {
 	return false, types.UID("")
 }
+
+func (m *podContainerManagerNoop) GetPodCgroupMemoryLimit(_ *v1.Pod) (uint64, error) {
```
pdgetrf, line 349: I probably have asked this before in another PR. why these functions returning constants?

pkg/kubelet/container/helpers.go
```diff
@@ -93,6 +96,16 @@ func ShouldContainerBeRestarted(container *v1.Container, pod *v1.Pod, podStatus
 // the running container with its desired spec.
 func HashContainer(container *v1.Container) uint64 {
 	hash := fnv.New32a()
+	if utilfeature.DefaultFeatureGate.Enabled(features.InPlacePodVerticalScaling) {
+		cResources := container.Resources
+		cResourcesAllocated := container.ResourcesAllocated
+		container.Resources = v1.ResourceRequirements{}
+		container.ResourcesAllocated = nil
+		defer func() {
```
pdgetrf, line 104: why the defer here?

pkg/kubelet/kubelet.go
```diff
@@ -1682,6 +1684,12 @@ func (kl *Kubelet) syncPod(o syncPodOptions) error {
 	// Fetch the pull secrets for the pod
 	pullSecrets := kl.getPullSecretsForPod(pod)
 
+	if utilfeature.DefaultFeatureGate.Enabled(features.InPlacePodVerticalScaling) {
+		if pod.DeletionTimestamp == nil {
```
pdgetrf, line 1688: move this condition to the same level as the if on line 1687?

pkg/kubelet/kuberuntime/kuberuntime_manager.go
```diff
@@ -920,6 +944,63 @@ func (m *kubeGenericRuntimeManager) computePodActions(pod *v1.Pod, podStatus *ku
...
+							restartContainer = true
+							return true
+						}
+						cUpdateInfo := containerToUpdateInfo{
+							specIndex:           idx,
+							apiContainerStatus:  &apiContainerStatus,
+							kubeContainerStatus: containerStatus,
+							restart:             &restartContainer,
+						}
+						switch {
```
pdgetrf, line 978: nit: seems if statement would suffice

pkg/kubelet/kuberuntime/kuberuntime_manager.go
```diff
@@ -946,6 +1027,11 @@ func (m *kubeGenericRuntimeManager) computePodActions(pod *v1.Pod, podStatus *ku
 
 	if keepCount == 0 && len(changes.ContainersToStart) == 0 {
 		changes.KillPod = true
+		if utilfeature.DefaultFeatureGate.Enabled(features.InPlacePodVerticalScaling) {
+			if len(changes.ContainersToRestart) != 0 {
```
pdgetrf, line 1031: move this condition up one level

pkg/kubelet/kuberuntime/kuberuntime_manager.go
```diff
@@ -1241,6 +1397,76 @@ func (m *kubeGenericRuntimeManager) SyncPodContainer(pod *v1.Pod, podStatus *kub
...
+	})
+
+	for _, cInfo := range containersToUpdate {
+		if *cInfo.restart == true {
+			continue
+		}
+		container := &pod.Spec.Containers[cInfo.specIndex]
+		specLimits := container.Resources.Limits
+		defer func() {
+			container.Resources.Limits = specLimits
```
pdgetrf, line 1417: this is to revert in case of failure right?  please consider add a comment here.

plugin/pkg/admission/podresourceallocation/admission_test.go
```diff
@@ -267,6 +268,7 @@ func TestValidateCreate(t *testing.T) {
 			expectError:        true,
 		},
 		//TODO: more unit tests and negative tests
+		//TODO: Add multi-container tests. Add tests for adding new container via patch
```
pdgetrf, line 271: please create issues to track such TODOs instead of in the code.

pkg/kubelet/cm/types.go
```diff
@@ -92,6 +93,14 @@ type CgroupManager interface {
 	ReduceCPULimits(cgroupName CgroupName) error
 	// GetResourceStats returns statistics of the specified cgroup as read from the cgroup fs.
 	GetResourceStats(name CgroupName) (*ResourceStats, error)
+	// GetCgroupMemoryLimit returns the memory limit of the specified cgroup as read from cgroup fs.
+	GetCgroupMemoryLimit(name CgroupName) (uint64, error)
+	// GetCgroupCpuLimit returns the cpu quota, cpu period, and cpu shares of the specified cgroup as read from cgroup fs.
```
yb01, line 98: it will be more extendable to return a pre-defined CPU struct instead of individual values of the CPU.

pkg/kubelet/kubelet.go
```diff
@@ -1701,6 +1709,50 @@ func (kl *Kubelet) syncPod(o syncPodOptions) error {
 	return nil
 }
 
+func (kl *Kubelet) handlePodResourcesResize(pod *v1.Pod) {
+	podResized := false
+	for _, container := range pod.Spec.Containers {
+		if len(diff.ObjectDiff(container.ResourcesAllocated, container.Resources.Requests)) > 0 {
```
yb01, line 1715: just of my curioucity,  should this diff only check the resources that can be resized or resize is supported, ie. CPU, Memory for the current design ? so it won't be disturbed by other resources such GPU or other cusotmer resources ?

pkg/kubelet/kubelet.go
```diff
@@ -1701,6 +1709,50 @@ func (kl *Kubelet) syncPod(o syncPodOptions) error {
...
+	}
+
+	var otherActivePods []*v1.Pod
+	activePods := kl.GetActivePods()
+	for _, p := range activePods {
+		if p.UID != pod.UID {
+			otherActivePods = append(otherActivePods, p)
+		}
+	}
+	if ok, failReason, failMessage := kl.canAdmitPod(otherActivePods, pod); !ok {
```
yb01, line 1731: is this function thread safe, and also, will this ActivePods, which was retrieved from the runtimeCache, which can be slight different with the exact current PODs ( with incomings from kubelet ). how do you plan to handle situations as such ?

pkg/kubelet/cm/cgroup_manager_linux.go
vinaykul: It is for decimal conversion. This is consistent with rest of the k8s codebase, and really with other code as well. See this from latest k8s upstream branch:
`root@fw0000358:~/MASTER_K8S# git grep "strconv\.FormatInt" 
cmd/kubeadm/app/apis/kubeadm/apiendpoint.go:    return net.JoinHostPort(endpoint.AdvertiseAddress, strconv.FormatInt(int64(endpoint.BindPort), 10))
cmd/kubeadm/app/util/config/initconfiguration.go:                       cfg.ControlPlaneEndpoint = net.JoinHostPort(host, strconv.FormatInt(int64(LocalAPIEndpoint.BindPort), 10))
pkg/api/v1/resource/helpers.go: return strconv.FormatInt(c, 10), nil
pkg/api/v1/resource/helpers.go: return strconv.FormatInt(m, 10), nil
pkg/api/v1/resource/helpers.go: return strconv.FormatInt(m, 10), nil
pkg/apis/apps/v1/conversion.go:         out.Annotations[appsv1.DeprecatedRollbackTo] = strconv.FormatInt(in.Spec.RollbackTo.Revision, 10)
pkg/apis/apps/v1/conversion.go: out.Annotations[appsv1.DeprecatedTemplateGeneration] = strconv.FormatInt(in.Spec.TemplateGeneration, 10)
pkg/apis/apps/v1beta2/conversion.go:            out.Annotations[appsv1beta2.DeprecatedRollbackTo] = strconv.FormatInt(in.Spec.RollbackTo.Revision, 10)
pkg/apis/apps/v1beta2/conversion.go:    out.Annotations[appsv1beta2.DeprecatedTemplateGeneration] = strconv.FormatInt(in.Spec.TemplateGeneration, 10)
pkg/controller/daemon/daemon_controller_test.go:        obj.SetResourceVersion(strconv.FormatInt(ver+1, 10))
pkg/controller/deployment/deployment_controller_test.go:        obj.SetResourceVersion(strconv.FormatInt(ver+1, 10))
pkg/controller/deployment/rollback.go:  d.Annotations[apps.DeprecatedRollbackTo] = strconv.FormatInt(rollbackTo.Revision, 10)
pkg/controller/deployment/sync.go:      newRevision := strconv.FormatInt(maxOldRevision+1, 10)
pkg/controller/deployment/util/deployment_util_test.go: return types.UID(strconv.FormatInt(rand.Int63(), 10))
`

-----

`root@fw0000358:~/MASTER_K8S# git grep "strconv\.FormatInt" vendor/
vendor/github.com/Azure/azure-sdk-for-go/storage/blockblob.go:          headers["Content-Length"] = strconv.FormatInt(n, 10)
vendor/github.com/Azure/azure-sdk-for-go/storage/util.go:                                       val = strconv.FormatInt(reflectedValue.Int(), 10)
vendor/github.com/BurntSushi/toml/encode.go:            enc.wf(strconv.FormatInt(rv.Int(), 10))
vendor/github.com/aws/aws-sdk-go/aws/credentials/stscreds/web_identity_provider.go:             sessionName = strconv.FormatInt(now().UnixNano(), 10)
vendor/github.com/aws/aws-sdk-go/aws/csm/metric.go:     return []byte(strconv.FormatInt(int64(ns/time.Millisecond), 10)), nil
vendor/github.com/aws/aws-sdk-go/aws/ec2metadata/api.go:        ttl := strconv.FormatInt(int64(duration/time.Second), 10)
vendor/github.com/aws/aws-sdk-go/aws/signer/v4/v4.go:           ctx.Query.Set("X-Amz-Expires", strconv.FormatInt(duration, 10))
vendor/github.com/aws/aws-sdk-go/private/protocol/query/queryutil/queryutil.go:         v.Set(name, strconv.FormatInt(value, 10))
vendor/github.com/aws/aws-sdk-go/private/protocol/rest/build.go:                str = strconv.FormatInt(value, 10)
vendor/github.com/aws/aws-sdk-go/private/protocol/timestamp.go:         return strconv.FormatInt(t.Unix(), 10)
vendor/github.com/aws/aws-sdk-go/private/protocol/xml/xmlutil/build.go:         str = strconv.FormatInt(converted, 10)
vendor/github.com/cespare/prettybench/prettybench.go:   return strconv.FormatInt(int64(iter), 10)
vendor/github.com/cilium/ebpf/asm/alu_string.go:                return "Source(" + strconv.FormatInt(int64(i), 10) + ")"
vendor/github.com/cilium/ebpf/asm/alu_string.go:                return "Endianness(" + strconv.FormatInt(int64(i), 10) + ")"
vendor/github.com/cilium/ebpf/asm/alu_string.go:        return "ALUOp(" + strconv.FormatInt(int64(i), 10) + ")"
vendor/github.com/cilium/ebpf/asm/func_string.go:               return "BuiltinFunc(" + strconv.FormatInt(int64(i), 10) + ")"
vendor/github.com/cilium/ebpf/asm/jump_string.go:       return "JumpOp(" + strconv.FormatInt(int64(i), 10) + ")"
vendor/github.com/cilium/ebpf/asm/load_store_string.go:         return "Mode(" + strconv.FormatInt(int64(i), 10) + ")"
vendor/github.com/cilium/ebpf/asm/load_store_string.go:         return "Size(" + strconv.FormatInt(int64(i), 10) + ")"
vendor/github.com/cilium/ebpf/asm/opcode_string.go:             return "Class(" + strconv.FormatInt(int64(i), 10) + ")"
vendor/github.com/cilium/ebpf/syscalls.go:      return strconv.FormatInt(fd.raw, 10)
`

pkg/kubelet/cm/cgroup_manager_linux.go
vinaykul: Ideally yes, but I'll stay with it for consistency with rest of codebase.

pkg/kubelet/kuberuntime/kuberuntime_manager.go
vinaykul: you mean move it into the same line as featuregate check? if yes, this is by intention- typically when we get rid of featuregates, it is easy to just remove the featuregate line and not have to rewrite the code. but i can move it up if you feel strongly about it.

plugin/pkg/admission/podresourceallocation/admission_test.go
vinaykul: Already have issue #289 . The TODOs help me locate the places where we need more by grepping on the git diff.

pkg/kubelet/cm/pod_container_manager_linux.go
vinaykul: It does not return an error - the consuming function does return error if there is a parameter problem. Here's the precedent from upstream master 1.18 codebase: 

`root@fw0000358:~/MASTER_K8S# git grep -A3 GetPodContainerName
pkg/kubelet/cm/pod_container_manager_linux.go:  podContainerName, _ := m.GetPodContainerName(pod)
pkg/kubelet/cm/pod_container_manager_linux.go-  return m.cgroupManager.Exists(podContainerName)
pkg/kubelet/cm/pod_container_manager_linux.go-}
pkg/kubelet/cm/pod_container_manager_linux.go-

pkg/kubelet/cm/pod_container_manager_linux.go:  podContainerName, _ := m.GetPodContainerName(pod)
pkg/kubelet/cm/pod_container_manager_linux.go-  // check if container already exist
pkg/kubelet/cm/pod_container_manager_linux.go-  alreadyExists := m.Exists(pod)
pkg/kubelet/cm/pod_container_manager_linux.go-  if !alreadyExists {
`

pkg/kubelet/cm/pod_container_manager_linux.go
vinaykul: these are no-op implementations. all other no-op interfaces are doing the same.

pkg/kubelet/container/helpers.go
vinaykul: I zero out Resources and ResourcesAllocated fields for computing the hash since these fields are mutable for Vertical Scaling, and we don't want the mutation to trigger a hash changed detection that syncpod uses to restart containers (restart is needed when image is updated for e.g)
defer restores the zeroed out fields back to their original values after the hash computation is done.

pkg/kubelet/kubelet.go
vinaykul: if we remove the featuregate at some-point, it is lot easier to just remove the above line, helps make review easier as well. i did this as i'm porting the same code for upstream - keeps things simple for me.

pkg/kubelet/kuberuntime/kuberuntime_manager.go
vinaykul: yes. i tried it out after reading that for cases like this switch is faster.. maybe golang optimizes similar if-else-if.

pkg/kubelet/kuberuntime/kuberuntime_manager.go
vinaykul: yes.

pkg/kubelet/kubelet.go
vinaykul: it checks if any of the desired resource values have changed than what's allocated. currently there's cpu, memory, and ephemeral-storage supported as valid resource-types in Resources field. someone (from alibaba i think) asked about implementing this for ephemeral storage as well but i scoped it out to keep things simple.

pkg/kubelet/kubelet.go
vinaykul: very good point.. this may be a problem. let me investigate this more. thanks for catching this!

vinaykul: this was indeed an issue. i've modified it to serialize with a mutex and i'm testing it now, will send an updated change out tomorrow.

yb01: excellent!

pkg/kubelet/kubelet.go
vinaykul: keeping this so that feature-gate flag removal is simpler.

pkg/kubelet/kuberuntime/kuberuntime_manager.go
vinaykul: upstream review is fine with this.

pkg/kubelet/kuberuntime/kuberuntime_manager.go
vinaykul: keeping this so that feature-gate flag removal is simpler.

pkg/kubelet/kuberuntime/kuberuntime_manager.go
vinaykul: this is actually to revert when exiting the function - appears to be a standard golang pattern, not just k8s pattern.

pkg/kubelet/cm/types.go
vinaykul: i updated the naming of this to reflect that it is getting and setting cgroup config based on upstream review. struct would be overkill.

pkg/kubelet/kubelet.go
vinaykul: i pushed a commit just now that fixes this for add vs. resize race, as well as multiple resize races.

vinaykul:
reopen to trigger new build

pkg/kubelet/kubelet.go
yb01: ephemeral resources are fine with resizing - the others like GPU etc, could be a valid resource for resizing. probably can have a future task for those resources.

pkg/api/v1/resource/helpers.go
```diff
@@ -96,6 +97,47 @@ func GetResourceRequest(pod *v1.Pod, resource v1.ResourceName) int64 {
...
+		addResourceList(allocations, container.ResourcesAllocated)
+	}
+	// init containers define the minimum of any resource
+	for _, container := range pod.Spec.InitContainers {
+		maxResourceList(allocations, container.Resources.Requests)
+	}
+	return
+}
+
+// GetResourceAllocation finds and returns resource allocation for a specific resource.
```
yb01, line 113: if simply the resource value is returned, the caller will need to adjust with the unit ( Mi, Gi, M, G etc ) which is error prone. return the rQuantity probably a better choice here.

pkg/kubelet/kubelet.go
```diff
@@ -2139,6 +2151,8 @@ func (kl *Kubelet) HandlePodActions(update kubetypes.PodUpdate) {
 func (kl *Kubelet) HandlePodAdditions(pods []*v1.Pod) {
 	start := kl.clock.Now()
 	sort.Sort(sliceutils.PodsByCreationTime(pods))
+	kl.podResizeMutex.Lock()
```
yb01, line 2154: for best practice, this function scope lock should be avoid as possible.

vinaykul: this is necessary. when kubelet restarts, all the pods for the node are sent by apiserver. and we need to process them in the lock so they get re-added in the order they were initially added. if we do this on per pod basis instead of the entire sequence, there is a chance that a pending resize of an earlier pod would be accepted before all the original pods could be added, resulting in a pod at the end of the list that was successfully added being rejected on kubelet restart.

pkg/kubelet/kubelet.go
vinaykul: they would require pods to be restarted by the controller - same behavior as today. the scope of this KEP is to allow restart-free resizing only for CPU and memory. ephermeral storage and other extended resources are explicitly disallowed in this kep and i have implemented validation check to block it. So you will never be able to trigger this by changing anything besides cpu or memory.

pkg/api/v1/resource/helpers.go
vinaykul: the use case for this is to get the actual value. if vertical scaling feature is on, then we would use this instead of [GetResourceRequest](https://github.com/futurewei-cloud/arktos/pull/343/files#diff-09e7de804ef61948e84b89d9dea88e51R73)

PR #352, opened 2020-06-05
added scripts and guide for user to setup tenant and access

pdgetrf:
Tracking issue: https://github.com/futurewei-cloud/arktos/issues/337

No additional tests performed as the scripts added are tested as detailed by the added doc.

### DIFF
--- a/hack/setup-multi-tenancy/setup_client.sh
+++ b/hack/setup-multi-tenancy/setup_client.sh
@@ -1,0 +1,64 @@
+#!/bin/bash
+
+# Copyright 2020 Authors of Arktos.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+set -e
+
+if [ "$#" -lt 2 ]
+then
+        echo "Usage: setup_client.sh [tenant] [person] (optional [host ip])"
+        exit 1
+fi
+
+tenant=$1
+person=$2
+namespace="default"
+tenant_person="${tenant}-${person}"
+group_name="${tenant}-group"
+context_name="${tenant_person}-context"
+ip=${3:-localhost}
+echo "setting up context $context_name for $tenant/$person in group $group_name at host $ip"
+
+CONFIG=/var/run/kubernetes/admin.kubeconfig
+#if [ -f "$CONFIG" ]; then
+#    echo "making a backup of previous config at $CONFIG.bk"
+#    mv $CONFIG $CONFIG.bk
+#fi
+
+echo "generating a client key $tenant.key..."
+openssl genrsa -out $tenant_person.key 2048 > /dev/null 2>&1
+
+echo "creating a sign request $tenant.csr..."
+openssl req -new -key $tenant_person.key -out $tenant.csr -subj "/CN=$person/O=tenant:$tenant/OU=$group_name" > /dev/null 2>&1
+
+echo "creating an tenant certificate ${tenant_person}.crt and get it signed by CA"
+openssl x509 -req -in $tenant.csr -CA /var/run/kubernetes/client-ca.crt -CAkey /var/run/kubernetes/client-ca.key -CAcreateserial -out $tenant_person.crt -days 500 > /dev/null 2>&1
+
+echo "Setting up tenant context..."
+kubectl config set-cluster ${tenant_person}-cluster --server=https://$ip:6443 --certificate-authority=/var/run/kubernetes/server-ca.crt > /dev/null 2>&1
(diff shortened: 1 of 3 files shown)


docs/user-guide/features/multi-tenancy-setup.md
```diff
@@ -0,0 +1,75 @@
...
+## Set Up Tenant For Testing
+
+[Two scripts](../../../hack/setup-multi-tenancy) are provided to quickly set up tenant on an Arktos cluster. There are two steps involved:
+
+### 1. Create A Tenant
+
+```bash
+./create_tenant.sh mycompany
+setting storage cluster to system
+tenant/futurewei created
```
chenqianfzh, line 25: the output should be "tenant/mycompany created".

docs/user-guide/features/multi-tenancy-setup.md
```diff
@@ -0,0 +1,75 @@
...
+This doc describes how to setup multi-tenancy on a testing cluster.
+
+## Tenant Bootstrap
+Tenant can be created using the following command:
+
+```bash
+kubectl create tenant $[tenant name]
+```
+
+When a new tenant is created, a cluster role and rolebinding is automatically created for this tenant for user "admin". This admin cluster role allows full access to this tenant's space resource management.
```
chenqianfzh, line 12: rephrase: to resource management in this tenant's space.

docs/user-guide/features/multi-tenancy-setup.md
pdgetrf: fixed here and other 2 places

pdgetrf:
I'm also moving the doc to setup-guide

image

hack/setup-multi-tenancy/setup_client.sh
```diff
@@ -0,0 +1,61 @@
...
+        exit 1
+fi
+
+tenant=$1
+person=$2
+namespace="default"
+tenant_person="${tenant}-${person}"
+group_name="${tenant}-group"
+context_name="${tenant_person}-context"
+ip=$3
```
chenqianfzh, line 32: can we make this parameter optional? The default value could be "localhost:6443", as a lot of engineers are running test cluster from the local dev box.

docs/user-guide/features/multi-tenancy-setup.md
pdgetrf: fixed

hack/setup-multi-tenancy/setup_client.sh
pdgetrf: good idea.

hack/setup-multi-tenancy/setup_client.sh
```diff
@@ -0,0 +1,61 @@
...
+kubectl config set-context $context_name --cluster=${tenant_person}-cluster --namespace=$namespace --user=$tenant_person --tenant=$tenant > /dev/null 2>&1
+echo "cleaned up $tenant.csr"
+rm $tenant.csr
+echo "***************************"
+echo "Context has been setup for tenant $tenant/$person in $CONFIG." 
+echo
+echo "Here's an example on how to access cluster:"
+echo
+set -x
+kubectl --context=${context_name} get pods
```
chenqianfzh, line 61: need to echo this line before running it, so user can take a look at the command.

pdgetrf: set -x does exactly that

hack/setup-multi-tenancy/setup_client.sh
```diff
@@ -0,0 +1,61 @@
...
+openssl x509 -req -in $tenant.csr -CA /var/run/kubernetes/client-ca.crt -CAkey /var/run/kubernetes/client-ca.key -CAcreateserial -out $tenant_person.crt -days 500 > /dev/null 2>&1
+
+echo "Setting up tenant context..."
+kubectl config set-cluster ${tenant_person}-cluster --server=https://$ip:6443 --certificate-authority=/var/run/kubernetes/server-ca.crt > /dev/null 2>&1
+kubectl config set-credentials ${tenant_person} --client-certificate=$tenant_person.crt --client-key=$tenant_person.key > /dev/null 2>&1
+kubectl config set-context $context_name --cluster=${tenant_person}-cluster --namespace=$namespace --user=$tenant_person --tenant=$tenant > /dev/null 2>&1
+echo "cleaned up $tenant.csr"
+rm $tenant.csr
+echo "***************************"
+echo "Context has been setup for tenant $tenant/$person in $CONFIG."
```
XiaoningDing, line 56: Probably we can add a prompt here that this can be used as a default context:  "kubectl config use-context $context_name"

chenqianfzh: I don't think we should add a prompt here. The scalabilty test team (they may be the heaviest user someday) may use this script to create thousands of tenants. They need to get the job silently done.

However, we can print some instructions on how to set current context at the end of the script.

XiaoningDing: That's what a "prompt" mean :) We just print something like "run XXX if you want to make this a default context".  So that users can simply copy and run the command if he wants.

pdgetrf: both good points. will add the print.

hack/setup-multi-tenancy/setup_client.sh
pdgetrf: added

hack/setup-multi-tenancy/setup_client.sh
pdgetrf: added the following part:

image

PR #354, opened 2020-06-09
TenantExists Admission Controller

chenqianfzh:
/kind: feature


plugin/pkg/admission/tenant/exists/admission.go
```diff
@@ -0,0 +1,119 @@
...
+	genericadmissioninitializer "k8s.io/apiserver/pkg/admission/initializer"
+	informers "k8s.io/client-go/informers"
+	"k8s.io/client-go/kubernetes"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+)
+
+// PluginName indicates name of admission plugin.
+const PluginName = "TenantExists"
+
+// Register registers a plugin
```
XiaoningDing, line 35: Just curious why this isn't implemented as a check in endpoint resolution, such as a new handler in the pipeline? It will require less configurations and also works for all actions (create, list, get....)

chenqianfzh: That is a very good question. I did not think about this way as the behavior I would like this PR to produce is very similar to  that of NamespaceExists Admission controller. So I mimicked NamespaceExists Admission controller.

I spent some time to see whether we can put the function in endpoints handler as you suggested. And the answer is probably not. Here is what I have learned:

If we would like the endpoint handler being able to check the availability of a tenant, the endpoint handler should have a tenantLister (or a tenantClient). The startup of APIServer are in phases. First it builds the basic config, where the step of building endpoint handler chain belongs to. Next, it builds the completed config, where the main job is to build shared informers and clients. This design is kind of expected, as informers/clients are external components dependent on the api server.

You may see the problem here. Our endpoint handler, which is built in phase I, is in need of components built in Phase II. So it is a mission impossible.

Please let me know you have any other thought.

PR #357, opened 2020-06-10
ETCD data partition part 2

Sindica:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note

```


staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go
```diff
@@ -159,6 +160,9 @@ func (w *watcher) run(ctx context.Context, key string, rev int64, recursive bool
 }
 
 func (w *watcher) updatePartitionConfig(dp corev1.DataPartitionConfig) {
+	// try to avoid concurrent map writes error
```
jshaofuturewei, line 163: My concerns is that do we need to add  concurrent read protection as well while the partitionConfig is being updated?

staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go
```diff
@@ -853,15 +936,103 @@ func (s *store) ttlOpts(ctx context.Context, ttl int64) ([]clientv3.OpOption, er
...
+			tenant = getTenantForKey(segs, 2)
+		}
+		if err != nil {
+			klog.Errorf("Regex match error %v. key %s", err, key)
+		}
+	}
+	if !hasPrefix {
+		tenant = getTenantForKey(segs, 1)
+	}
+	if tenant == "" {
```
jshaofuturewei, line 991: We don't think we have empty tenant now. Do we need to add warnf here?

staging/src/k8s.io/apiserver/pkg/storage/storagecluster/tenantstoragemapmanager.go
```diff
@@ -0,0 +1,273 @@
...
+
+func (ts *TenantStorageMapManager) syncTenants() error {
+	ts.mux.Lock()
+	klog.V(4).Infof("mux acquired syncTenants.")
+	defer func() {
+		ts.mux.Unlock()
+		klog.V(4).Infof("mux released syncTenants.")
+	}()
+
+	// Question: do we need to ignore deleted tenants?
```
jshaofuturewei, line 110: Can we check there? Like deleteTimeStamp?

staging/src/k8s.io/apiserver/pkg/storage/storagecluster/tenantstoragemapmanager.go
```diff
@@ -0,0 +1,273 @@
...
+	ts.mux.Lock()
+	klog.V(4).Infof("mux acquired deleteTenant.")
+	delete(ts.tenantToClusterMap, tenant.Name)
+
+	ts.mux.Unlock()
+	klog.V(4).Infof("mux released deleteTenant.")
+}
+
+func (ts *TenantStorageMapManager) GetClusterIdFromTenant(tenant string) string {
+	ts.mux.RLock()
```
jshaofuturewei, line 260: It seems that read and write have no differences at all.

staging/src/k8s.io/apiserver/pkg/storage/etcd3/watcher.go
Sindica: From the code, reads only happens after update is done. However, we still have concurrent write error. I will keep digging.

staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go
Sindica: We don't have empty tenant. But the algorithm is not intelligent enough to find tenant only. It is possible that there are an extra / in the key so that tenant name was extracted wrong. "" is safe guard.

staging/src/k8s.io/apiserver/pkg/storage/storagecluster/tenantstoragemapmanager.go
Sindica: I think list should not have objects that have deleteTimeStamp

staging/src/k8s.io/apiserver/pkg/storage/storagecluster/tenantstoragemapmanager.go
Sindica: https://golang.org/src/sync/rwmutex.go
The lock can be held by an arbitrary number of readers or a single writer.

RLock does not prevent additional reader, but it can prevent from writing.

PR #363, opened 2020-06-11
admission controller: annotates pod network-readiness=false by default

h-w-chen:
**What type of PR is this?**
/kind feature

This custom admission controller, "DefaultPodNetworkReadiness", adds annotation "arktos.futurewei.com/network-readiness": "false" to pods, except for those of host network, or already have the annotation (regardless of the value).

External network provider may need to do some network resource preparations on its custom network controller before pod is scheduled and able to start. pod annotation of "arktos.futurewei.com/network-readiness" is introduced as signal to sync up scheduler with the network preparation task before scheduling is made for the pod.

This admission controller is typically required by Arktos working with external VPC network providers. For flat typed network, as there is no controller conducting such network preparations, this admission should not be enabled.

This is part of Arktos support for multi-tenancy multi-network.

**How to verify in local dev env**
1. ensure kube-apiserver will start with this admission controller enabled (in hack/lib/common-var-init.sh)
```ENABLE_ADMISSION_PLUGINS=${ENABLE_ADMISSION_PLUGINS:-"...,DefaultPodNetworkReadiness"}```
2. run hack/arktos-up.sh
3. create a regular pod
4. look at the pod detail spec in yaml format, ensure below annotation is there
```yaml
kind: Pod
metadata:
  annotations:
    arktos.futurewei.com/network-readiness: "false"
```

YES
```release-note
For Arktos working with VPC network providers like Mizar, Neutron etc, kube-apiserver should start with DefaultPodNetworkNotReady admission controller enabled, via parameter like
    --enable-admission-plugins=...,DefaultPodNetworkReadiness
```


plugin/pkg/admission/defaultpodnetworknotready/admission.go
```diff
@@ -0,0 +1,85 @@
...
+import (
+	"io"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apiserver/pkg/admission"
+	api "k8s.io/kubernetes/pkg/apis/core"
+)
+
+const (
+	// PluginName indicates name of admission plugin.
+	PluginName = "DefaultPodNetworkNotReady"
```
chenqianfzh, line 28: Minor: I found the admission controller names in K8s usually do not have the default value appended, such as "DefaultStorgaeClass" or "DefaultTolerationSeconds".  If we follow the same convention, the name should be something like "DefaultPodNetworkStatus" or "DefaultPodNetworkReadiness".

plugin/pkg/admission/defaultpodnetworknotready/admission.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	if _, readinessExists := pod.Annotations[networkReadiness]; readinessExists {
+		return nil
+	}
+
+	if pod.ObjectMeta.Annotations == nil {
+		pod.ObjectMeta.Annotations = map[string]string{}
+	}
+	pod.ObjectMeta.Annotations[networkReadiness] = "false"
```
chenqianfzh, line 68: I believe pod.Annotations(used in lin 61) and pod.ObjectMeta.Annotations (used in line 65,66,68) refer to the same thing. So the code should work, but let's keep the code consistent.

plugin/pkg/admission/defaultpodnetworknotready/admission_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+	handler := admissiontesting.WithReinvocationTesting(t, &plugin{})
+
+	for _, tc := range tcs {
+		t.Run(tc.desc, func(t *testing.T) {
+			pod := tc.pod
+			attributes := admission.NewAttributesRecord(pod, nil, api.Kind("Pod").WithVersion("version"), pod.Tenant, pod.Namespace, pod.Name, api.Resource("pods").WithVersion("version"), "", admission.Create, &metav1.CreateOptions{}, false, nil)
+			err := handler.Admit(attributes, nil)
+			if err != nil {
+				t.Fatalf("Unexpected error returned from admission handler")
```
chenqianfzh, line 84: Let's log what the err is.

plugin/pkg/admission/defaultpodnetworknotready/admission.go
h-w-chen: being consistent makes sense. will change accordingly.

plugin/pkg/admission/defaultpodnetworknotready/admission.go
h-w-chen: they are not. here the annotation is newly added.
a comment will add to make sure it easier to understand

plugin/pkg/admission/defaultpodnetworknotready/admission_test.go
h-w-chen: good point. will log err.

h-w-chen: will log.

PR #364, opened 2020-06-11
"make update" should return 1 if there is file updated. 

Hong-Chang:
"make update" should return 1 if there is file updated. This is to hook up pr ci check to avoid miss running of "make update".

**What type of PR is this?**
/kind EE

### DIFF
--- a/hack/make-rules/update.sh
+++ b/hack/make-rules/update.sh
@@ -79,14 +79,12 @@ ${KUBE_ROOT}/hack/arktos_copyright.sh ${KUBE_ROOT} ${KUBE_ROOT}/_output
 
 echo -e "${color_green:?}Update scripts completed successfully${color_norm}"
 
-if [ "${JENKINS_HOME:-}" != "" ]; then
-	echo "Under Jenkins CI environment, check whether there is file updated."
-	echo "We define return value as 0 if there is no file updated, and 1 if there is file updated."
-	MODIFIED_FILES=$(git status | grep modified:) || true
-	if [[ "${MODIFIED_FILES:-}" != "" && "${MODIFIED_FILES:-}" != "	modified:   api/openapi-spec/swagger.json" ]]; then
-		echo "Following files are updated:"
-		echo $MODIFIED_FILES
-		exit 1
-	fi
-	echo "Check completed."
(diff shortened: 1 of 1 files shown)


hack/make-rules/update.sh
```diff
@@ -78,3 +78,12 @@ done
 ${KUBE_ROOT}/hack/arktos_copyright.sh ${KUBE_ROOT} ${KUBE_ROOT}/_output
 
 echo -e "${color_green:?}Update scripts completed successfully${color_norm}"
+
+echo "Now let's check whether there is file updated."
+echo "We define return value as 0 if there is no file updated, and 1 if there is file updated."
+MODIFIED_FILES=$(git status | grep modified:) || true
```
vinaykul, line 84: we should only do this in ci builds. normal dev environment builds should not because users can have filed checked out or staged.

Hong-Chang: Updated. Only check in ci environment.

Sindica: This should work in CI the same way as work in dev env. Otherwise, dev will be confused if CI fails but local dev make update pass

Hong-Chang: Now changed back to check in both cases.

hack/make-rules/update.sh
```diff
@@ -78,3 +78,13 @@ done
 ${KUBE_ROOT}/hack/arktos_copyright.sh ${KUBE_ROOT} ${KUBE_ROOT}/_output
 
 echo -e "${color_green:?}Update scripts completed successfully${color_norm}"
+
+echo "Now, let's check whether there is file updated."
+echo "We define return value as 0 if there is no file updated, and 1 if there is file updated."
+MODIFIED_FILES=$(git status | grep modified:) || true
+if [[ "${MODIFIED_FILES:-}" != "" && "${MODIFIED_FILES:-}" != "	modified:   api/openapi-spec/swagger.json" ]]; then
```
Sindica, line 85: Why exclude swagger.json? If swagger.json is updated, there is something wrong.

Hong-Chang: Updated

PR #369, opened 2020-06-12
delete CRD during tenant deletion

chenqianfzh:
This PR is the second PR to address issue https://github.com/futurewei-cloud/arktos/issues/338.

For the convenience of reviewers, the code is split into two commits, where the commit "auto update"  is code updated by command "make update"

### What issues does the PR address?
With PR https://github.com/futurewei-cloud/arktos/pull/356, a non-system tenant will be deleted successfully if the tenant does not have CRD or custom resources. But tenant deletion will hit the following issues:
1. If there is a CRD defined under the tenant and custom resources under the tenant, tenant deleter will delete the CRD but not the custom resources.
2. If a system-forced-sharing CRD is defined by the system tenant, the deletion of a tenant will get stuck.

### How does the PR solve the issues?

To address the first issue, changes to make to the CRD resource type, so it no longer support **"deletecollection"** verb. As a result, tenant-deleter will use verb *"delete"** instead to remove the CRD resources, which will trigger the CRD finalizer to remove the custom resources.

As to the second issue, tenant deleter will skip the system-forced-shared CRD when checks whether the resources are cleard.

### Test Done is my Dev Box
I created a tenant called atlantis. 

The created a system-forced-shared CRD and deployed a custom resources of this system-forced-shared CRD.

```
$prompt:~/gopath/qianc/arktos$ kubectl create -f ~/gopath/qianc/arktos/test/yaml/multi_tenancy_examples/crd/forced_crd.yaml
customresourcedefinition.apiextensions.k8s.io/burgers.arktos.futurewei.com created
 
$prompt:~/gopath/qianc/arktos$ kubectl apply -f ~/gopath/qianc/arktos/test/yaml/multi_tenancy_examples/crd/turkey-burger.yaml --tenant atlantis
burger.arktos.futurewei.com/turkey-burger created
```


pkg/controller/tenant/deletion/tenanted_resources_deleter.go
```diff
@@ -151,6 +152,7 @@ func (d *tenantedResourcesDeleter) Delete(tenantName string) error {
 	if finalized(tenant) {
 		return d.deleteTenant(tenant)
 	}
+
```
pdgetrf, line 155: extra newline

chenqianfzh: removed.

chenqianfzh: Funny. I removed it. Then ran "make update" and found the extra newline was back. :-D

PR #371, opened 2020-06-12
external network provider integration support: scheduler holds pods not network ready

h-w-chen:
**What type of PR is this?**
/kind feature

makes scheduler not to schedule pods that are marked as network not ready.

This is part of external network provider integration support. For more information, please refer to [PR- design proposal: pod annotations as hint & cni args](https://github.com/futurewei-cloud/arktos/pull/341)

(Partially) Yes
It has no impacts on pods to be scheduled except for those explicitly annotated as network-readiness not true
```release-note
pods explicitly having annotation "arktos.futurewei.com/network-readiness" not "true" are put on hold for scheduling by kube-scheduler
```

### DIFF
--- a/pkg/scheduler/eventhandlers.go
+++ b/pkg/scheduler/eventhandlers.go
@@ -22,10 +22,12 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog"
 	"reflect"
+	"strings"
 
 	"k8s.io/api/core/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	arktosv1 "k8s.io/arktos-ext/pkg/apis/arktosextensions/v1"
 	coreinformers "k8s.io/client-go/informers/core/v1"
 	storageinformers "k8s.io/client-go/informers/storage/v1"
 	"k8s.io/client-go/tools/cache"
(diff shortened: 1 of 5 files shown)


pkg/scheduler/eventhandlers.go
```diff
@@ -292,6 +294,20 @@ func vmPodShouldSleep(pod *v1.Pod) bool {
 	return pod.Status.Phase == v1.PodNoSchedule
 }
 
+func podIsNetworkReady(pod *v1.Pod) bool {
+	if v, ok := pod.Annotations[arktosv1.NetworkReadiness]; ok {
+		v = strings.TrimSpace(v)
```
yb01, line 299: is it a good idea to treat empty string or blanks as the "true" condition?

h-w-chen: as absence of such annotation is treated as true, empty/blank valued annotation treated as true is reasonable.

pkg/scheduler/eventhandlers.go
```diff
@@ -387,7 +403,7 @@ func AddAllEventHandlers(
 			FilterFunc: func(obj interface{}) bool {
 				switch t := obj.(type) {
 				case *v1.Pod:
-					return !assignedPod(t) && responsibleForPod(t, schedulerName) && !vmPodShouldSleep(t)
+					return !assignedPod(t) && responsibleForPod(t, schedulerName) && !vmPodShouldSleep(t) && podIsNetworkReady(t)
```
XiaoningDing, line 406: Not sure if it's necessary to couple "networkReadiness" with scheduling. This might created some undesired situation: 1) Potential deadlocks, some network controllers may need scheduling decision in order to finish network resource creation. If a controller sets such work into two phases, it's complicated and will cause problem for kubelet  2) scheduler now has a dependency on external network controllers.

h-w-chen: This annotation is purposed for holding scheduler from premature schedulings. If no controller updates the annotation, yes the pod will not be scheduled. Such synchronization mechanism is collaborative between controller and scheduler; so each one has to fulfill its responsibility.

Regarding your points of concerns,
1) in that case, yes task should be split into pre-scheduling and other (concurrent-scheduling or post-scheduling); if kubelet has problem handling this, it will have problem handling w/o such mechanism;
2) yes scheduler has a dependency (coupling) on the collaborative controller. If this depency is undesired, we can consider using a switch (like scheduler startup param or feature gate) to turn it on/off.

h-w-chen:
offline discussion concluded that such holding(synchronization) is moved to kubelet. Scheduler won't check such readiness annotation; kubelet will wait instead. This PR will be discarded, for now.

If in the future there is need to have scheduler blocking on some condition of network preparation, we'll revisit this option.

PR #372, opened 2020-06-12
Add a flag IS_DRY_RUN which will be used in Jenkins CI

Hong-Chang:
In Jenkins CI, it needs majority steps which is to generating binaries, scripts for preset machines, but don't execute it. It delays executing to be run by Jenkins clients. So in this pr it introduced a new flag that indicating the process only generate scripts but not execute it.

**What type of PR is this?**
/kind feature

### DIFF
--- a/aws_kube_params.inc
+++ b/aws_kube_params.inc
@@ -20,7 +20,7 @@ export NUM_WINDOWS_NODES=0
 export AWS_SSH_KEY=$HOME/.ssh/id_rsa
 
 export PRESET_INSTANCES_ENABLED=false
-export IS_DRY_RUN=false
+export IS_PRESET_INSTANCES_DRY_RUN=false  # In Preset instance, if it's dry run, then it only generate script but not deploying. It has no impact to preset instances.
 export PRESET_KUBE_MASTER_IP=172.31.6.63
 export PRESET_KUBE_MINION1_IP=172.31.11.206
 # Uncomment next line and set minion2 ip to enable deploy 2-nodes cluster
--- a/aws_kube_params.inc
+++ b/aws_kube_params.inc
@@ -20,6 +20,7 @@ export NUM_WINDOWS_NODES=0
 export AWS_SSH_KEY=$HOME/.ssh/id_rsa
 
 export PRESET_INSTANCES_ENABLED=false
+export IS_DRY_RUN=false
 export PRESET_KUBE_MASTER_IP=172.31.6.63
 export PRESET_KUBE_MINION1_IP=172.31.11.206
 # Uncomment next line and set minion2 ip to enable deploy 2-nodes cluster
--- a/aws_kube_params.inc
+++ b/aws_kube_params.inc
@@ -20,7 +20,7 @@ export NUM_WINDOWS_NODES=0
 export AWS_SSH_KEY=$HOME/.ssh/id_rsa
 
 export PRESET_INSTANCES_ENABLED=false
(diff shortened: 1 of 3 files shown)


aws_kube_params.inc
```diff
@@ -20,6 +20,7 @@ export NUM_WINDOWS_NODES=0
 export AWS_SSH_KEY=$HOME/.ssh/id_rsa
 
 export PRESET_INSTANCES_ENABLED=false
+export IS_PRESET_INSTANCES_DRY_RUN=false  # In Preset instance, if it's dry run, then it only generate script but not deploying. It has no impact to preset instances.
```
sonyafenge, line 23: your comments mentioned "It has no impact to preset instances"? please double check.

cluster/aws/util.sh
```diff
@@ -1384,6 +1397,8 @@ function start-minions() {
     fi
   ) > "${KUBE_TEMP}/node-user-data"
 
+  cp ${KUBE_TEMP}/node-user-data /tmp/node-user-data
```
sonyafenge, line 1400: what is the purpose for this line?

aws_kube_params.inc
Hong-Chang: Right, there is no impact.

cluster/aws/util.sh
Hong-Chang: $KUBE_TEMP is a random name that Jenkins has no idea. So this is to copy the file to tmp folder so Jenkins has access.

cluster/aws/util.sh
```diff
@@ -733,11 +736,13 @@ function upload-server-tars() {
     SERVER_BINARY_TAR_URL=$PRESET_SERVER_BINARY_TAR_URL
     BOOTSTRAP_SCRIPT_URL=$PRESET_BOOTSTRAP_SCRIPT_URL
 
-    scp -o 'StrictHostKeyChecking no' -i ${ACCESS_FILE} ${SERVER_BINARY_TAR} ${BOOTSTRAP_SCRIPT} ${SSH_USER}@${KUBE_MASTER_IP}:/tmp
-    scp -o 'StrictHostKeyChecking no' -i ${ACCESS_FILE} ${SERVER_BINARY_TAR} ${BOOTSTRAP_SCRIPT} ${SSH_USER}@${KUBE_MINION1_IP}:/tmp
-    if [[ -n ${KUBE_MINION2_IP:-} ]]; then
-      scp -o 'StrictHostKeyChecking no' -i ${ACCESS_FILE} ${SERVER_BINARY_TAR} ${BOOTSTRAP_SCRIPT} ${SSH_USER}@${KUBE_MINION2_IP}:/tmp
-    fi    
+    if [[ $IS_PRESET_INSTANCES_DRY_RUN == $FALSE ]]; then
+      scp -o 'StrictHostKeyChecking no' -i ${ACCESS_FILE} ${SERVER_BINARY_TAR} ${BOOTSTRAP_SCRIPT} ${SSH_USER}@${KUBE_MASTER_IP}:/tmp
```
Sindica, line 740: Can scp be avoided in non dry run case if we have Jenkins CI workflow set up?

Hong-Chang: yes, that's designed to be. In Jenkins pipeline, $IS_PRESET_INSTANCES_DRY_RUN will be true, by condition in the code, scp will be skipped. And since there is no access file in Jenkins, scp won't success anyway.

PR #382, opened 2020-06-17
In-place Pod Vertical Scaling - add resource-quota & limit-ranger support

vinaykul:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

**What this PR does / why we need it**: Vertical Scaling needs to work with resource quota and limit ranges. This PR adds that support.

Fixes #

Tested this by creating a resource-quota limiting CPUs to 4, and creating pod and updating (attempting) beyond the quota, as well as within the quota.
For limit range, create a limitrange with min 2 cpu, max 5 cpu, and resize within and outside the bounds. The latter fails.

```release-note

```

### DIFF
--- a/plugin/pkg/admission/limitranger/admission.go
+++ b/plugin/pkg/admission/limitranger/admission.go
@@ -1,5 +1,6 @@
 /*
 Copyright 2014 The Kubernetes Authors.
+Copyright 2020 Authors of Arktos - file modified.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


plugin/pkg/admission/limitranger/admission.go
```diff
@@ -466,6 +469,14 @@ func (d *DefaultLimitRangerActions) SupportsAttributes(a admission.Attributes) b
 	if a.GetKind().GroupKind() == api.Kind("Pod") && a.GetOperation() == admission.Update {
 		return false
 	}
+	if !utilfeature.DefaultFeatureGate.Enabled(features.InPlacePodVerticalScaling) {
```
XiaoningDing, line 472: You didn't remove the existing check. I don't see how this newly added check works.  Pod.update will return false even before hit your new check, right?

vinaykul: good catch. i rebased yesterday and that action somehow lost my commit, so i applied the saved patch to the rebase, and it was rejected - i messed up copying the check but not removing it. i'll fix it.

vinaykul: fixed and tested in latest commit.

PR #399, opened 2020-06-19
create system tenant at tenant controller bootstrap

chenqianfzh:
Create the system tenant at boostrap.

Verification Done in Dev Box:
Right after the cluster is up by arktos-up.sh, run the following commands:
```
qianchen@qianchen-VirtualBox:~$ ETCDCTL_API=3 etcdctl get "" --prefix=true --keys-only | grep tenants
/registry/tenants/system

qianchen@qianchen-VirtualBox:~$ kubectl create tenant system
setting storage cluster to system
Error from server (AlreadyExists): tenants "system" already exists
```

### DIFF
--- a/pkg/controller/tenant/tenant_controller.go
+++ b/pkg/controller/tenant/tenant_controller.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/api/core/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilerrors "k8s.io/apimachinery/pkg/util/errors"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
@@ -127,6 +128,7 @@ func (tc *TenantController) Run(workers int, stopCh <-chan struct{}) {
 	defer tc.queue.ShutDown()
 
 	klog.Infof("Starting tenant controller.")
+	tc.cerateSystemTenantIfNotExist()
 	defer klog.Infof("Shutting down tenant controller")
 
 	if !controller.WaitForCacheSync("tenant", stopCh, tc.listerSynced) {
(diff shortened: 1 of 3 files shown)


pkg/controller/tenant/tenant_controller.go
```diff
@@ -355,3 +357,20 @@ func flattenedError(failures []error, tenant string) (error, bool) {
...
+func (tc *TenantController) cerateSystemTenantIfNotExist() {
+	_, err := tc.lister.Get(metav1.TenantSystem)
+
+	if apierrors.IsNotFound(err) {
+		tc.kubeClient.CoreV1().Tenants().Create(&v1.Tenant{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: metav1.TenantSystem,
+			},
+			Spec: v1.TenantSpec{
+				StorageClusterId: "system",
```
pdgetrf, line 371: do we have some constant for the string "system" somewhere? it's used repeatedly in lots of code.

pkg/controller/tenant/tenant_controller.go
```diff
@@ -127,6 +128,7 @@ func (tc *TenantController) Run(workers int, stopCh <-chan struct{}) {
 	defer tc.queue.ShutDown()
 
 	klog.Infof("Starting tenant controller.")
+	tc.cerateSystemTenantIfNotExist()
```
pdgetrf, line 131: if you have time, please add a test in the tenant controller integration test here:

https://github.com/futurewei-cloud/arktos/blob/master/test/integration/tenantcontroller/tenantcontroller_test.go

pkg/controller/tenant/tenant_controller.go
chenqianfzh: I had a talk with Ying Huang on this a while. According to her, we can just hard code "system" for now. She will define a constant later. Issue (https://github.com/futurewei-cloud/arktos/issues/326) was created and assigned to her to track this.

pkg/controller/tenant/tenant_controller.go
chenqianfzh: Added test case TestSystemTenantCreatedAutomically as suggested

pdgetrf:
not related to this PR, buy can the system tenant be deleted?

chenqianfzh:
> not related to this PR, buy can the system tenant be deleted?

No, system tenant can never be deleted. The Namespace LifeCycle admission controller (this is a default admission controller) make sure the following resource can never be deleted:
1. tenant "system"
2. namespace system/default
3. namespace system/kube-system
4. namespace system/kube-public

PR #412, opened 2020-06-24
using new tenanted user info in integration tests

chenqianfzh:
These tests were using the old-format user info when the multi-tenancy cert and token work was not done. It is time to use the new tenant-aware user info now.

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/authentication/group/group_adder_test.go
+++ b/staging/src/k8s.io/apiserver/pkg/authentication/group/group_adder_test.go
@@ -1,5 +1,6 @@
 /*
 Copyright 2016 The Kubernetes Authors.
+Copyright 2020 Authors of Arktos - file modified.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


staging/src/k8s.io/apiserver/pkg/authentication/group/group_adder_test.go
```diff
@@ -29,7 +30,7 @@ func TestGroupAdder(t *testing.T) {
 	adder := authenticator.Request(
 		NewGroupAdder(
 			authenticator.RequestFunc(func(req *http.Request) (*authenticator.Response, bool, error) {
-				return &authenticator.Response{User: &user.DefaultInfo{Name: "user", Groups: []string{"original"}}}, true, nil
+				return &authenticator.Response{User: &user.DefaultInfo{Tenant: "test-tenant", Name: "user", Groups: []string{"original"}}}, true, nil
```
pdgetrf, line 33: extra the string into a const

chenqianfzh: Fixed.

PR #418, opened 2020-06-26
Vertical Scaling for VMs (in addition to containers)

vinaykul:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
/kind feature
> /kind flake

**What this PR does / why we need it**: Arktos design supports In-Place restart-free vertical resizing of Pods for both container and VM workloads in a unified way. The full container vertical scaling feature was checked in earlier. This PR adds support for VM vertical scaling on the kubernetes-side (separate PR from Yunwen in runtime repo will bring the runtime-side of the work).

Fixes #

Verified by performing memory and cpu resizing for vmdefault example VM yaml, and verified that vertical scaling has not regressed for contaienrs.

```release-note
Example usage:
FEATURE_GATES="InPlacePodVerticalScaling=true" ./hack/arktos-up.sh
./cluster/kubectl.sh create -f ./test/yaml/vm_default.yaml
./cluster/kubectl.sh patch pod vmdefault --patch '{"spec":{"virtualMachine":{"name":"vm", "resources":{"requests":{"memory":"300Mi"}, "limits":{"memory":"300Mi"}}}}}'


```

### DIFF
--- a/pkg/apis/core/types.go
+++ b/pkg/apis/core/types.go
@@ -2953,6 +2953,25 @@ func (ps *PodSpec) Workloads() []CommonInfo {
 	return ps.WorkloadInfo
 }
 
+func (ps *PodSpec) SetWorkloads() {
+	// Update mutable fields
+	if len(ps.WorkloadInfo) != 0 {
+		if ps.VirtualMachine != nil {
+			ps.VirtualMachine.Image = ps.WorkloadInfo[0].Image
+			ps.VirtualMachine.Resources = ps.WorkloadInfo[0].Resources
+			ps.VirtualMachine.ResourcesAllocated = ps.WorkloadInfo[0].ResourcesAllocated
+			ps.VirtualMachine.ResizePolicy = ps.WorkloadInfo[0].ResizePolicy
+		} else {
+			for i := range ps.Containers {
+				ps.Containers[i].Image = ps.WorkloadInfo[i].Image
+				ps.Containers[i].Resources = ps.WorkloadInfo[i].Resources
+				ps.Containers[i].ResourcesAllocated = ps.WorkloadInfo[i].ResourcesAllocated
+				ps.Containers[i].ResizePolicy = ps.WorkloadInfo[i].ResizePolicy
+			}
+		}
+	}
+}
+
 // HostAlias holds the mapping between IP and hostnames that will be injected as an entry in the
 // pod's hosts file.
 type HostAlias struct {
--- a/pkg/kubelet/kubelet.go
+++ b/pkg/kubelet/kubelet.go
@@ -31,6 +31,7 @@ import (
 	"os"
 	"path"
 	"sort"
+	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
--- a/pkg/kubelet/kubelet.go
+++ b/pkg/kubelet/kubelet.go
@@ -20,6 +20,7 @@ package kubelet
 import (
 	"context"
 	"crypto/tls"
+	"encoding/json"
 	"fmt"
 	arktosv1 "k8s.io/arktos-ext/pkg/apis/arktosextensions/v1"
 	"k8s.io/kubernetes/pkg/kubelet/runtimeregistry"
@@ -30,7 +31,6 @@ import (
 	"os"
 	"path"
 	"sort"
-	"strings"
 	"sync"
 	"sync/atomic"
 	"time"
@@ -45,6 +45,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/diff"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/apimachinery/pkg/util/strategicpatch"
 	"k8s.io/apimachinery/pkg/util/wait"
 	utilfeature "k8s.io/apiserver/pkg/util/feature"
 	clientset "k8s.io/client-go/kubernetes"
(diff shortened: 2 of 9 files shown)


pkg/apis/core/types.go
```diff
@@ -2953,6 +2953,25 @@ func (ps *PodSpec) Workloads() []CommonInfo {
 	return ps.WorkloadInfo
 }
 
+func (ps *PodSpec) SetWorkloads() {
+	// Update mutable fields
```
pdgetrf, line 2957: nit: removable comment.

pdgetrf:
this change should also work with cpu right?

vinaykul:
> this change should also work with cpu right?

yes. it should work with cpu at this time even if libvirt does not support updating cpus because of cgroup management of the vm process. i've tested cpu and memory updation for both containers and vms -- e2e for containers, and up until the cri boundary for vm. the actual updating fo vm requires a fix from yunwen, he has a pr but needs to integrate with this pr and try it.

yb01:
the runtime part has a few changes to support this ( for details, please refer to the PR in runtime repo, a wiki and doc is in progress as well to describe it ):
1. at runtime, resource management is controlled all in runtime itself --- libvirt won't control the CG anymore for vm process
2. at runtime, CG and runtime vm config, libvirt domain settings are updated and sync-ed per resource changes and reporting.

pkg/apis/core/types.go
```diff
@@ -2953,6 +2953,25 @@ func (ps *PodSpec) Workloads() []CommonInfo {
...
+	if len(ps.WorkloadInfo) != 0 {
```
yb01, line 2958: we have a commonInfo which can be used to avoid this if-else block, any reason it is not used here ?

pkg/kubelet/kubelet.go
```diff
@@ -1728,29 +1729,43 @@ func (kl *Kubelet) canResizePod(pod *v1.Pod) (bool, string) {
...
 
-	var containersPatchData string
-	for _, container := range pod.Spec.Containers {
-		var resourcesPatchData string
-		for rName, rQuantity := range container.Resources.Requests {
-			container.ResourcesAllocated[rName] = rQuantity
-			resourcesPatchData += fmt.Sprintf(`"%s":"%s",`, rName, rQuantity.String())
+	podCopy := pod.DeepCopy()
+	for _, workload := range podCopy.Spec.Workloads() {
+		for rName, rQuantity := range workload.Resources.Requests {
```
yb01, line 1737: should this be the limits of the resources or both requests and limits ?

pkg/kubelet/util/podConverter/podConverter.go
```diff
@@ -126,14 +126,15 @@ func ConvertVmPodToContainerPod(pod *v1.Pod) *v1.Pod {
...
 		{
-			Name:            pod.Spec.VirtualMachine.Name,
-			Image:           pod.Spec.VirtualMachine.Image,
-			ImagePullPolicy: pod.Spec.VirtualMachine.ImagePullPolicy,
-			VolumeMounts:    pod.Spec.VirtualMachine.VolumeMounts,
-			VolumeDevices:   pod.Spec.VirtualMachine.VolumeDevices,
-			TTY:             true,
-			Stdin:           true,
-			Resources:       pod.Spec.VirtualMachine.Resources,
+			Name:               pod.Spec.VirtualMachine.Name,
```
yb01, line 129: the vcpu annotation needs to be set as well, in line 115 above.

pkg/kubelet/kuberuntime/kuberuntime_manager.go
```diff
@@ -650,134 +650,150 @@ func (m *kubeGenericRuntimeManager) computePodActions(pod *v1.Pod, podStatus *ku
...
-						// If container is in unknown state, we don't know whether it
-						// is actually running or not, always try killing it before
-						// restart to avoid having 2 running instances of the same container.
-						changes.ContainersToKill[containerStatus.ID] = containerToKillInfo{
-							name:      containerStatus.Name,
-							container: &pod.Spec.Containers[idx],
-							message:   fmt.Sprintf("Container is in %q state, try killing it before restart", containerStatus.State),
-						}
+		// If container does not exist, or is not running, check whether we
+		// need to restart it.
```
yb01, line 669: this logic will need to handle vm cases. i.e. in here the check for workload type in line 653 on left side should NOT be ignored or the VirtualMachineStatus is needed to check.

The prelimary testing with this PR proves that the RemoveContianer and restart VM.

pkg/apis/core/types.go
vinaykul: this is not on apiserver, rather on local object. that's why the comment to clarify.

pkg/apis/core/types.go
vinaykul: this is common info field.

pkg/kubelet/kubelet.go
vinaykul: only requests here. requests is the node capacity reservation done by kubelet. and kubelet always attempts to reconcile desired limits once apiserver has validated it and update the podspec.

pkg/kubelet/kuberuntime/kuberuntime_manager.go
vinaykul: Is RestartPolicy not applicable to VMs? I'm not sure why we need to check VirtualMachineStatus here at all. With the translation, we should apply this check to both containers and VMs unless there is some reason I missed.

pkg/kubelet/util/podConverter/podConverter.go
vinaykul: it is being set in 116, i have not changed that by adding resourcesAllocated. when pod is resized, new vcpu count should be set. what specifically did i miss here?

yb01: indeed. my bad.

pkg/kubelet/kuberuntime/kuberuntime_manager.go
yb01: makes a sense. the restart of VM is caused by other issues as we just investigated. so i just close this one. in term of RestartPolicy enforcement or application to vm types. it can be tracked in the design doc. or applied in a case by case logic.

PR #422, opened 2020-06-27
Not to assign ClusterIP for service whose associated network is of External IPAM type

h-w-chen:
**What type of PR is this?**
/kind feature

modify kube-apiserver to bypass clusterIP assignment for services whose VIP address space is managed by external network provider. It also adds spec.service section to network CRD.

Based on [service IP managed by network provider](https://github.com/futurewei-cloud/arktos/blob/master/docs/design-proposals/multi-tenancy/multi-tenancy-network.md#network-provider-managed-pool), Arktos leaves service ClusterIP as unassigned (eternal network provider's controller should assign them instead), if the associated networks explicitly state the External IPAM type.

This PR temporarily ignores the error that the default network not created (which would be mitigated by tenant controller if the feature of automatic creation of default network for new tenant is enabled). This temporary measure is necessary to keep system not broken before the required dependencies are all in place. However, if service explicitly associates to a network, the network has to exist.

This is part of multi-tenant networking features.

1. start etcd, kube-api-server (http protocol, port 8080). The kube-apiserver start command is
```kube-apiserver -- --etcd-servers=http://127.0.0.1:2379```
2. ensure network CRD created
```kubectl apply -f pkg/controller/artifacts/crd-network.yaml```
3. creates a network CR of External IPAM type using following yaml
```yaml
kind: Network
metadata:
  name: test1
spec:
  type: mizar
  vpcID: 12345-mizar
  service:
    ipam: External
```
4. able to create a service associated to test1 network using following yaml, verify that its will has no clusterIP
```yaml
apiVersion: v1
kind: Service
metadata:
  name: my-nginx
  labels:
    run: my-nginx
    arktos.futurewei.com/network: test1 
spec:
  ports:
  - port: 80
    protocol: TCP
  selector:
    run: my-nginx
```
5. verify creating a service associated to an inexistent network using following yaml fails with error "networks.arktos.futurewei.com test2 not found"
```yaml
apiVersion: v1
kind: Service
metadata:
  name: my-nginx-inexistent-net
  labels:
    run: my-nginx
spec:
  ports:
  - port: 80
    protocol: TCP
  selector:
    run: my-nginx
```

YES
```release-note
1. If a service associates a network by providing arktos.futurewei.com/network label, the named network has to be created already; otherwise, service cannot be created
2. The temporary exception is services not having such network label. This temporary exception will be removed when the dependency features are all enabled.
```


pkg/controller/artifacts/crd-network.yaml
```diff
@@ -31,6 +31,17 @@ spec:
               type: string
             vpcID:
               type: string
+            service:
+              type: object
+              properties:
+                ipam:
+                  type: string
+                  enum: ["Kubernetes", "External"]
```
chenqianfzh, line 39: Minor: "kubernetes", or "Arktos"? I prefer "Arktos" as ae are forking away from K8s.

plugin/pkg/admission/defaultpodnetworkreadiness/admission_test.go
```diff
@@ -84,7 +84,7 @@ func TestAdmit(t *testing.T) {
 				t.Fatalf("Unexpected error returned from admission handler: %v", err)
 			}
 
-			v, ok := pod.Annotations[networkReadiness]
+			v, ok := pod.Annotations["arktos.futurewei.com/network-readiness"]
```
chenqianfzh, line 87: I think there is constant defined for ""arktos.futurewei.com/network-readiness"

h-w-chen: yes there is - but in [another PR](https://github.com/futurewei-cloud/arktos/pull/427) that is already under review.

pkg/controller/artifacts/crd-network.yaml
h-w-chen: what matters is non-External and External; however, adding Arktos also makes sense. Will add "Arktos" too.

plugin/pkg/admission/defaultpodnetworkreadiness/admission_test.go
h-w-chen: it is intentional; in test code, I prefer to using hand-craft string instead of simple reference to const definition.

pkg/registry/core/service/storage/rest.go
```diff
@@ -130,6 +139,53 @@ func (rs *REST) Categories() []string {
...
+		if storage, err := crStorageGetter.GetCustomResourceStorage(tenant, "networks.arktos.futurewei.com", "v1"); err == nil {
+			defer atomic.StoreUint32(&rs.done, 1)
+			rs.networks = storage.CustomResource
+			return rs.networks, nil
+		} else {
+			return nil, fmt.Errorf("custom resource storage not set up yet")
+		}
+	}
+
+	if rs.networks == nil {
```
chenqianfzh, line 174: I am just curious, when the code will hit line 174? When will rs.done == 1 && atomic.loadUnit32(rs.done) == 0?

pkg/registry/core/service/storage/rest.go
```diff
@@ -130,6 +139,53 @@ func (rs *REST) Categories() []string {
...
+			return rs.networks, nil
+		}
+
+		crStorageGetter := extensionsapiserver.GetCustomResourceStoragesGetter()
+		if crStorageGetter == nil {
+			defer atomic.StoreUint32(&rs.done, 1)
+			return nil, fmt.Errorf("api extension server not initialized properly")
+		}
+
+		if storage, err := crStorageGetter.GetCustomResourceStorage(tenant, "networks.arktos.futurewei.com", "v1"); err == nil {
```
chenqianfzh, line 165: A const for "networks.arktos.futurewei.com"?

pkg/registry/core/service/storage/rest.go
```diff
@@ -130,6 +139,53 @@ func (rs *REST) Categories() []string {
...
+			defer atomic.StoreUint32(&rs.done, 1)
+			rs.networks = storage.CustomResource
+			return rs.networks, nil
+		} else {
+			return nil, fmt.Errorf("custom resource storage not set up yet")
```
chenqianfzh, line 170: Let's log the err in the returned message.

pkg/registry/core/service/storage/rest.go
```diff
@@ -130,6 +139,53 @@ func (rs *REST) Categories() []string {
...
+	if rs.done == 0 {
+		if rs.networks != nil {
+			defer atomic.StoreUint32(&rs.done, 1)
+			return rs.networks, nil
+		}
+
+		crStorageGetter := extensionsapiserver.GetCustomResourceStoragesGetter()
+		if crStorageGetter == nil {
+			defer atomic.StoreUint32(&rs.done, 1)
+			return nil, fmt.Errorf("api extension server not initialized properly")
```
chenqianfzh, line 162: I would like the error message rephrase more specifically for convenience in debugging: "Failed to get the Custom Resource Storage Getter: api extension server not ....."

pkg/registry/core/service/storage/rest.go
```diff
@@ -162,8 +218,44 @@ func (rs *REST) Export(ctx context.Context, name string, opts metav1.ExportOptio
...
 }
 
+func (rs *REST) isExternalIPAM(ctx context.Context, service *api.Service) (bool, error) {
+	netName := service.Labels[arktosextensionsv1.NetworkLabel]
+	if netName == "" {
+		// choose the default network name if not specified
+		netName = "default"
+	}
+
+	network, err := rs.getNetwork(context.Background(), service.Tenant, netName)
```
chenqianfzh, line 228: Why discard the ctx in the input param and create a new one?

pkg/registry/core/service/storage/rest.go
```diff
@@ -162,8 +218,44 @@ func (rs *REST) Export(ctx context.Context, name string, opts metav1.ExportOptio
 	return rs.services.Export(ctx, name, opts)
 }
 
+func (rs *REST) isExternalIPAM(ctx context.Context, service *api.Service) (bool, error) {
+	netName := service.Labels[arktosextensionsv1.NetworkLabel]
+	if netName == "" {
+		// choose the default network name if not specified
+		netName = "default"
```
chenqianfzh, line 225: kinda feeling "default" will be a const used a lot in future work. Give it a const definition?

pkg/registry/core/service/storage/rest.go
chenqianfzh: Actually, the tenant info should be available from ctx. It should be the same as the Tenant in service.

pkg/registry/core/service/storage/rest.go
```diff
@@ -162,8 +218,44 @@ func (rs *REST) Export(ctx context.Context, name string, opts metav1.ExportOptio
...
+			if vs, ok := v.(string); ok {
+				return vs == string(arktosextensionsv1.IPAMExternal), nil
+			}
+		}
+	}
+
+	return false, nil
+}
+
 func (rs *REST) Create(ctx context.Context, obj runtime.Object, createValidation rest.ValidateObjectFunc, options *metav1.CreateOptions) (runtime.Object, error) {
```
chenqianfzh, line 253: Looks to me we need some logic in Update() so the system cannot assign IP to an External network.

pkg/registry/core/service/storage/rest_test.go
```diff
@@ -67,6 +69,34 @@ type serviceStorage struct {
 	Err                error
 }
 
+type testNetworkGetter struct {
+	ipamType string
+	err      error
+}
+
+func (n testNetworkGetter) Get(ctx context.Context, name string, options *metav1.GetOptions) (runtime.Object, error) {
```
chenqianfzh, line 77: param "name" and "ctx" are never used. Is it intentional?

BTW, this struct is especially created for mizar ( or say external IPAM type) test. It is better to rename "testNetworkGetter" to indicate it.

pkg/registry/core/service/storage/rest_test.go
```diff
@@ -2280,3 +2312,75 @@ func TestUpdateNodePorts(t *testing.T) {
 		}
 	}
 }
+
+func TestServiceRegistryCreateWithNetworkOfExternalIPAM(t *testing.T) {
+	storage, registry, server := NewTestREST(t, nil)
+	storage.networks = &testNetworkGetter{ipamType: "External"}
+	defer server.Terminate(t)
+
+	svc := &api.Service{
```
chenqianfzh, line 2321: Add Tenant/Namespace in ObjectMeta.

pkg/registry/core/service/storage/rest.go
h-w-chen: see sync.Once code; if this method is called in different goroutines simultaneously, it is possible to reach this line.

pkg/registry/core/service/storage/rest.go
h-w-chen: the service context has service related info, which would lead to misleading error message (which seems derived based on context request info). Having a fresh context mitigate this problem. However, it makes sense if a subcontext can have new request info (get network cr) on top of the ctx param; a todo comment added as reminder for such consideration in future revisits.

pkg/registry/core/service/storage/rest.go
h-w-chen: will have a const as suggested.

pkg/registry/core/service/storage/rest.go
h-w-chen: yes thanks for the feedback. will modify update method.

pkg/registry/core/service/storage/rest_test.go
h-w-chen: ctx & name not used at all as it is a test dummy fixture.
change the struct type name to indicates that it is used as a dummy.

pkg/registry/core/service/storage/rest_test.go
h-w-chen: will add the default tenant/ns explicitly.

pkg/registry/core/service/storage/rest.go
h-w-chen: will add const as suggested.

pkg/registry/core/service/storage/rest_test.go
```diff
@@ -2280,3 +2312,79 @@ func TestUpdateNodePorts(t *testing.T) {
...
+
+func TestServiceRegistryCreateWithNetworkOfExternalIPAM(t *testing.T) {
+	storage, registry, server := NewTestREST(t, nil)
+	storage.networks = &dummyNetworkGetter{ipamType: "External"}
+	defer server.Terminate(t)
+
+	svc := &api.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "foo",
+			Tenant:    "system",
```
chenqianfzh, line 2324: minor: use metav1.TenantSystem instead of "system".

pkg/registry/core/service/storage/rest_test.go
```diff
@@ -2280,3 +2312,79 @@ func TestUpdateNodePorts(t *testing.T) {
...
+	}
+}
+
+func TestServiceRegistryCreateWithoutExistingNetwork(t *testing.T) {
+	storage, _, server := NewTestREST(t, nil)
+	storage.networks = &dummyNetworkGetter{err: fmt.Errorf("network not found")}
+	defer server.Terminate(t)
+
+	svc := &api.Service{
+		ObjectMeta: metav1.ObjectMeta{
```
chenqianfzh, line 2368: Add tenant and namespace in ObjectMeta.

pkg/registry/core/service/storage/rest.go
chenqianfzh: Thanks for sharing the knowledge.

pkg/registry/core/service/storage/rest_test.go
h-w-chen: test data; simple string inplace seems to help understandability

pkg/registry/core/service/storage/rest_test.go
h-w-chen: will add tenant + namespace as suggested

pkg/controller/artifacts/crd-network.yaml
```diff
@@ -31,6 +31,17 @@ spec:
               type: string
             vpcID:
               type: string
+            service:
+              type: object
+              properties:
+                ipam:
+                  type: string
+                  enum: ["Arktos", "Kubernetes", "External"]
```
XiaoningDing, line 39: Maybe combine "Arktos" and "Kubernetes" to a generic term "System"?   system v.s external

h-w-chen: what really matters is being "External" or not (kubernetes, arktos or system all treated equally). For now, it seems "Arktos" sounds most understandable. Anyway, if a more general value is more suitable in the future, change (or adding) won't affect implementation much.

PR #427, opened 2020-06-29
kubelet checks network readiness

h-w-chen:
**What type of PR is this?**
/kind feature

Kubelet syncPod workflow has one more step to check pod's network-readiness annotation value. 

As depicted in [multi-tenancy networking design proposal](https://github.com/futurewei-cloud/arktos/blob/master/docs/design-proposals/multi-tenancy/multi-tenancy-network.md), pod has arktos.futurewei.com/network-readiness annotation served as signal that external network component indicates the preparation of networking for the pod has been done, and pod is able to start. It is necessary for kubelet to add one more check in the syncPod workflow to ensure the signal is not blocking.

This is one feature needed in kubelet to support multi-tenant networking.  

1. start Arktos system by ```hack/arktos-up.sh```
2. create a pod that has network-readiness=false, by ```kubectl create -f pod-net-not-ready.yaml```, with the yaml file 
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: nginx-system-n-n-r
  annotations:
    "arktos.futurewei.com/network-readiness": "false"
spec:
  containers:
  - name: nginx
    image: nginx
    ports:
    - containerPort: 80
```
3. observe the pod just created won't be able to start;
```
$ cluster/kubectl.sh get pods nginx-system-n-n-r
nginx-system-n-n-r   3277677723223307134   0/1     ContainerCreating   0          10m
```
4. change network-readiness value to "true" using ```kubectl edit pod nginx-system-n-n-r```;
5. shortly this pod turns into Running state
```
$ cluster/kubectl.sh get pods nginx-system-n-n-r
nginx-system-n-n-r          3277677723223307134   1/1     Running   0          12m
```

Yes
```release-note
pods annotated by key of "arktos.futurewei.com/network-readiness" and value if not "true" shall be put on hold to start at the work node, until the condition is cleared.
```


yb01:
if network readiness if false, should it stuck at  or right after the RunSandbox() instead of CreateContainer() ?

pkg/kubelet/kubelet.go
```diff
@@ -1689,6 +1691,18 @@ func (kl *Kubelet) syncPod(o syncPodOptions) error {
 		}
 	}
 
+	// If pod is not host network, check network-readiness (only fails the cases with explicit not true value)
```
yb01, line 1694: just to clarify the case when this annotation is needed. in line 1592, the runtimeState.NetworkError() has check for network related errors reported from runtime. is this annotation covers some other cases where runtimeState network is ready but somehow it is not ready for a particular pod under some situations ?

h-w-chen: runtimeState.NetworkError() is about node's status (mostly about cni). network-readiness annotation is about the pod only; some external network provider may need this to sync with arktos (kubelet in this case), in order to prevent premature  pod launch before the network provider finishes _mandatory_ preparation (if there is any)

PR #430, opened 2020-07-01
fix --partition-config to pickup  apiserver-config from /etc/srv/kube…

sonyafenge:
…rnetes

**What type of PR is this?**

> /kind bug

apiserver failed to pickup apiserver-config, change to pick up from /etc/srv/kubenetes

Fixes #

```release-note

```


cluster/gce/gci/configure-helper-common.sh
```diff
@@ -1778,7 +1778,10 @@ function start-kube-apiserver {
   fi
   echo "APISERVER_SERVICEGROUPID: ${APISERVER_SERVICEGROUPID:-0}"
   params+=" --service-group-id=${APISERVER_SERVICEGROUPID:-0}"
-  params+=" --partition-config=${KUBE_HOME}/apiserver.config"
+  if [[ -f "${KUBE_HOME}/apiserver.config" ]]; then
+    mv ${KUBE_HOME}/apiserver.config /etc/srv/kubernetes/
+    params+=" --partition-config=/etc/srv/kubernetes/apiserver.config"
```
Sindica, line 1783: Can /etc/srv/kubernetes/ be a constant? It seems used a lot. It does not need to be done now. It can be done later for code refactor.

PR #448, opened 2020-07-02
Fixed the issue 441 that hallow nodes fail to pick up multiple apiserver configurations

jshaofuturewei:
Please refer to https://github.com/futurewei-cloud/arktos/issues/441


cmd/kubemark/hollow-node.go
```diff
@@ -249,3 +253,23 @@ func run(config *hollowNodeConfig) {
 		hollowProxy.Run()
 	}
 }
+
+// API Server Config Manager is used to watch api server configuration update
+// Upon the inroduction of api server data partition, it needs to connect to update
+// connections when API Server Configuration is updated
+// Hallow kubelet and proxy can be start as static pod - do not fail when it cannot connect to API Server
+func startAPIServerConfigManager(client clientset.Interface, stopCh <-chan struct{}) {
```
Sindica, line 261: This is duplicated from cmd/kubelet/app/server.go:startAPIServerConfigManager()
Please make it a util and update both. We don't need to duplicate the code.

Issue #447, opened 2020-07-02
kubelet needs RBAC permission to get/list network CR


h-w-chen:
**What would you like to be added**:
kubelet running on node is able to get/list network custom resources, across all tenants
API Group: "arktos.futurewei.com"
resource: "networks"

**Why is this needed**:
[Arktos networking design](https://github.com/futurewei-cloud/arktos/blob/master/docs/design-proposals/multi-tenancy/multi-tenancy-network.md) is centered around the network objects, which have their specific DNS service IPs. kubelet will need this network specific DNS data to prepare the pod's /etc/resolv.conf setting (for pods of ClusterFirst DNSpolicy).

h-w-chen:
kubelet uses [node authorizer admission controller] (https://kubernetes.io/docs/reference/access-authn-authz/node/), which does not include networks of "arktos.futurewei.com" api group. Adding a proper set of cluster role / binding should suffice to remedy it..

PR #449, opened 2020-07-02
kubelet prepares pod DNS config with network specific DNS service IP

h-w-chen:
**Do not merge yet**
it requires the feature that kubelet has permission to get network CR, tracked by #447

**What type of PR is this?**
/kind feature

this PR changes pod DNS config (/etc/resolv.conf) - for pod of ClusterFirst DNS Policy, its name server shall be the DNS service IP address of the associated network. 

Based on [multi-tenancy network design](https://github.com/futurewei-cloud/arktos/blob/master/docs/design-proposals/multi-tenancy/multi-tenancy-network.md), pod resides in a certain network; as each network has its own DNS service, pods inside of the network should point to this network specific DNS IP. This is a breaking change - k8s uses cluster-shared DNS ip address for all pods, Arktos assigns pod with its associated network DNS address.

Whenever kubelet needs to get network CR, it sends get request to API server. This approach is simple, but not most efficient. This PR chooses this approach for its simplicity. One direction of improvement is to introduce cache to avoid unnecessary queries. This optimization can be considered in the future.

In this PR,  if the pod does not explicitly associate to a network and the default network not created yet, pod DNS config still uses  the cluster-shared DNS IP, as current k8s does. This temporary measure is not to break system before all multi-tenant dependencies are in place. It should be removed when all are ready.

**How to verify at the local dev env**:
note: in this PR, __kubelet needs permission to get networks custom resource of group "arktos.futurewei.com".__ If the kubelet of your env has not such permission yet, you can temporarily run kubelet in cluster-admin context.
1. start arktos by arktos-up.sh
2. ensure network CRD defined by ```./cluster/kubectl.sh create -f ./pkg/controller/artifacts/crd-network.yaml```
3. create tenant x by ```./cluster/kubectl.sh create tenant x```
4. inside tenant x create network net-y using following yaml
```yaml
kind: Network
metadata:
  name: net-y
  tenant: x
spec:
  type: flat
```
5. manually simulate (e.g. through subresource status update) net-y to have dns service ip assigned, like below
```
./cluster/kubectl.sh get ne --tenant x
NAME    TYPE   VPC   PHASE   DNS
net-y   flat         Ready   1.2.3.4
```
6. create pod inside tenat x using following pod spec:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: nginx-y
  tenant: x
  labels:
    name: nginx
    arktos.futurewei.com/network: net-y
spec:
  containers:
  - name: nginx
    image: nginx
    ports:
    - containerPort: 80
```
7. verify pod-y /etc/resolv.conf having following content - 1.2.3.4 is the DNS service ip defined in net-y:
```
search default.svc.cluster.local svc.cluster.local cluster.local
nameserver 1.2.3.4
options ndots:5
```

Yes
```release-note
Arktos assigns pod /etc/resolv.config with its associated network DNS address, if pod is of ClusterFirst DNSPolicy
```


pkg/kubelet/network/dns/dns.go
```diff
@@ -63,20 +66,50 @@ type Configurer struct {
...
+			return []net.IP{ip}
+		}
+	} else {
+		klog.Errorf("failed to get network %s/%s: %v", pod.Tenant, networkName, err)
+	}
+
+	// for now, fallback to default cluster DNS for pod that has no explicit network and no valid dns service ip of the default network not set yet.
+	// This is temporary measure to keep system unbroken before all multi-tenant network dependencies are in place.
+	// todo: disable the fallback when every components are ready.
+	if networkName == defaultNetwork {
```
XiaoningDing, line 106: The logic seems unclear to me. Why do we need to compare network name to defaultNetwork here?  The default network is not special except that it's pre-created, which also has its own DNS service. Do we only fallback for default networks?

pkg/kubelet/network/dns/dns.go
```diff
@@ -63,20 +66,50 @@ type Configurer struct {
...
 		nodeIP:         nodeIP,
 		clusterDNS:     clusterDNS,
 		ClusterDomain:  clusterDomain,
 		ResolverConfig: resolverConfig,
+		arktosV1Client: arktosV1Client,
 	}
 }
 
+func (c *Configurer) getClusterDNS(pod *v1.Pod) []net.IP {
+	const defaultNetwork = "default"
```
XiaoningDing, line 88: Do we have shared definitions for these strings?

pkg/kubelet/network/dns/dns.go
```diff
@@ -63,20 +66,50 @@ type Configurer struct {
...
+		sip := network.Status.DNSServiceIP
+		ip := net.ParseIP(sip)
+		if ip != nil {
+			return []net.IP{ip}
+		}
+	} else {
+		klog.Errorf("failed to get network %s/%s: %v", pod.Tenant, networkName, err)
+	}
+
+	// for now, fallback to default cluster DNS for pod that has no explicit network and no valid dns service ip of the default network not set yet.
```
XiaoningDing, line 103: Does a cluster still work if no network CRD is installed? I'm not sure if this is only temporary or it should be always there.

pkg/kubelet/network/dns/dns.go
```diff
@@ -63,20 +66,50 @@ type Configurer struct {
...
 	}
 }
 
+func (c *Configurer) getClusterDNS(pod *v1.Pod) []net.IP {
+	const defaultNetwork = "default"
+	networkName, ok := pod.Labels["arktos.futurewei.com/network"]
+	if !ok {
+		networkName = defaultNetwork
+	}
+	if network, err := c.arktosV1Client.NetworksWithMultiTenancy(pod.Tenant).Get(networkName, metav1.GetOptions{}); err == nil {
```
XiaoningDing, line 93: Needs to check if c.arktosV1Client.NetworksWithMultiTenancy(pod.Tenant).Get() will trigger nil pointer exception if the network CRD is not installed.

pkg/kubelet/network/dns/dns.go
XiaoningDing: I'm also thinking if it makes sense to make per-network optional. If we remove "if networkName == defaultNetwork" check here, we will have an effect of " if the network object has its own DNS, use it; otherwise fallback to the per-cluster one".  I'm still thinking about it -- not sure if it covers practical scenarios :)

pkg/kubelet/network/dns/dns.go
h-w-chen: yes we have. the PR was just merged; I'll reuse the consts.

pkg/kubelet/network/dns/dns.go
h-w-chen: intention here is as temporary measure not to break system before the networking components are all ready (particularly the default network objects being created automatically on new tenant). I'll update code to make it clear.

pkg/kubelet/network/dns/dns.go
h-w-chen: c.arktosV1Client.NetworksWithMultiTenancy(pod.Tenant) won't return nil; it just sets up tenant value as client property, which will be used to assemble the rest path of network CR's url. So no nil reference risk here.

PR #450, opened 2020-07-03
increase apimachinery integration test tiemout

dingyin:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note

```

### DIFF
--- a/test/integration/apimachinery/BUILD
+++ b/test/integration/apimachinery/BUILD
@@ -2,6 +2,7 @@ load("@io_bazel_rules_go//go:def.bzl", "go_test")
 
 go_test(
     name = "go_default_test",
+    size = "large",
     srcs = [
         "main_test.go",
         "watch_restart_test.go",
--- a/test/integration/master/BUILD
+++ b/test/integration/master/BUILD
@@ -8,7 +8,7 @@ load(
 
 go_test(
     name = "go_default_test",
-    size = "large",
+    timeout = "long",
     srcs = [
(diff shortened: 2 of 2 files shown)


test/integration/apimachinery/BUILD
```diff
@@ -2,6 +2,7 @@ load("@io_bazel_rules_go//go:def.bzl", "go_test")
 
 go_test(
     name = "go_default_test",
+    size = "large",
```
chenqianfzh, line 5: I think a better solution is
    timeout = "long"

I found the following in the doc https://docs.bazel.build/versions/master/test-encyclopedia.html:
"Unlike timeout, the size also determines the limits for other resources like RAM, as described in Common definitions."

dingyin: I think another reason is more test cases, that is why I set is to 'large'

chenqianfzh: Actually, there is only one test case in apimachinery integ

test/integration/master/BUILD
```diff
@@ -8,7 +8,7 @@ load(
 
 go_test(
     name = "go_default_test",
-    size = "large",
+    timeout = "long",
```
chenqianfzh, line 11: Actually, I think we should keep size="large" here. There are many test cases here, so more RAM is preferred.

PR #458, opened 2020-07-07
fix issue 393

chenqianfzh:
This PR fixes issue https://github.com/futurewei-cloud/arktos/issues/393.

Error will occur if a regular-tenant user use "--tenant system" in the kubectl command, instead of working towards the space of the tenant itself.

### Bug analysis and explanation about fix

The bug was caused by the combination of the following two facts:
1. When client-go creates the HTTP URL, if the tenant is system, the section of "tenants/system" in the HTTP URL is skipped for backwards compatibility. 
2. With the introduction of "short path" in URL resolution, no tenant in the URL means a request to the user tenant. 

To fix the bug 393, a simple way is to put the section of "tenants/system" in the url. It fixes the issue. However, more than 100 test cases will fail as the backwards compatibility is broken. 

I tried different ways to let client-go create different URLs based on the user tenants:
1. the "tenants/system" section in URL is skipped if it is a request from a user of system tenant
2  the "tenants/system" section in URL is NOT skipped if it is a request from a user of regular tenant

However, the component of client-go URL generator and the component to resolve the user tenant from the kubeconfig or command line option in kubectl are far away from each other. They are almost at the two ends of the command process workflow in kubectl. To make the client-go URL generator check or have the info whether the request is from a user tenant, I tried several different ways. However, they either need pervasive code change, or pollute the componentization design in kubectl.

That is why I finally settled down with the solution given in this PR. I let kubectl do the check. "--tenant system" is not a valid command line option is the user tenant is a regular one.

### verification in Dev box

Created a tenant, "qian" and the tenant-admin user, "qian-admin". A corresponding context was also set up. The following commands were tested:

```
qianchen@qianchen-VirtualBox:~$ kubectl --context=qian-admin-context get ns
NAME          STATUS   AGE   TENANT
default       Active   3s    qian
kube-public   Active   2s    qian
kube-system   Active   3s    qian

qianchen@qianchen-VirtualBox:~$ kubectl --context=qian-admin-context get ns --tenant system
error: System space is not accessible for tenant qian

qianchen@qianchen-VirtualBox:~$ kubectl config use-context qian-admin-context
Switched to context "qian-admin-context".

qianchen@qianchen-VirtualBox:~$ kubectl get ns
NAME          STATUS   AGE   TENANT
default       Active   91s   qian
kube-public   Active   90s   qian
kube-system   Active   91s   qian

qianchen@qianchen-VirtualBox:~$ kubectl get ns --tenant system
error: System space is not accessible for tenant qian

qianchen@qianchen-VirtualBox:~$ kubectl get ns --tenant qian
NAME          STATUS   AGE    TENANT
default       Active   104s   qian
kube-public   Active   103s   qian
kube-system   Active   104s   qian
```

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/responsewriters/errors.go
+++ b/staging/src/k8s.io/apiserver/pkg/endpoints/handlers/responsewriters/errors.go
@@ -1,5 +1,6 @@
 /*
 Copyright 2014 The Kubernetes Authors.
+Copyright 2020 Authors of Arktos - file modified.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -65,6 +66,10 @@ func forbiddenMessage(attributes authorizer.Attributes) string {
 
 	if ns := attributes.GetNamespace(); len(ns) > 0 {
 		return fmt.Sprintf("User %q cannot %s resource %q in API group %q in the namespace %q", username, attributes.GetVerb(), resource, attributes.GetAPIGroup(), ns)
+	}
+
+	if tenant := attributes.GetTenant(); len(tenant) > 0 {
+		return fmt.Sprintf("User %q cannot %s resource %q in API group %q in the tenant %q", username, attributes.GetVerb(), resource, attributes.GetAPIGroup(), tenant)
 	}
 
 	return fmt.Sprintf("User %q cannot %s resource %q in API group %q at the cluster scope", username, attributes.GetVerb(), resource, attributes.GetAPIGroup())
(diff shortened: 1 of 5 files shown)


staging/src/k8s.io/apiserver/pkg/endpoints/handlers/responsewriters/errors.go
```diff
@@ -67,6 +68,10 @@ func forbiddenMessage(attributes authorizer.Attributes) string {
 		return fmt.Sprintf("User %q cannot %s resource %q in API group %q in the namespace %q", username, attributes.GetVerb(), resource, attributes.GetAPIGroup(), ns)
 	}
 
+	if tenant := attributes.GetTenant(); len(tenant) > 0 {
+		return fmt.Sprintf("User %q cannot %s resource %q in API group %q in the tenant %q", username, attributes.GetVerb(), resource, attributes.GetAPIGroup(), tenant)
+	}
+
```
XiaoningDing, line 74: This is a special rule enforced on the client side. Usually access rules should be enforced on server side.  I'm wondering if it's possible to handle this way:
1. The explicit "--tenant" parameter in command line alway takes precedence over context configuration. So if users specify "--tenant system" then we emit URL "/tenants/system/...", nothing special.
2. On server side we recognize "system" as a valid space name (probably we already did so?).

PR #473, opened 2020-07-13
Add ETCD data partition into manual

Sindica:


Sindica:
> I went through the ETCD-Data-Partition-Design.md and found the 2nd data partition is from m to Z. It does not work since all the uppercases are less than the lowercases.

Thank you for catching it. I think you meant arktos-apiserver-partition.md

PR #475, opened 2020-07-13
Fixed the issue that kube-up does not pick up etcd3.4.4-arktos.1

jshaofuturewei:
As title

### DIFF
--- a/cluster/gce/manifests/etcd-empty-dir-cleanup.yaml
+++ b/cluster/gce/manifests/etcd-empty-dir-cleanup.yaml
@@ -14,4 +14,4 @@ spec:
   dnsPolicy: Default
   containers:
   - name: etcd-empty-dir-cleanup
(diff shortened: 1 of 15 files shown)


cluster/gce/manifests/etcd-empty-dir-cleanup.yaml
```diff
@@ -14,4 +14,4 @@ spec:
   dnsPolicy: Default
   containers:
   - name: etcd-empty-dir-cleanup
-    image: k8s.gcr.io/etcd-empty-dir-cleanup:3.4.4.0
+    image: k8s.gcr.io/etcd-empty-dir-cleanup:3.4.4-arktos.1-1
```
sonyafenge, line 17: should this change to under gcr.io/{{ project_id }}/etcd-amd64?

PR #484, opened 2020-07-15
per-network service IP allocation - part 1: feature gate and preparation refactoring

h-w-chen:
**What type of PR is this?**
/kind feature

This is preparation for per-network service IP allocation:
1. introduces new feature gate PerNetworkServiceIPAlloc (off by default);
2. refactors api server code to call method instead of fetching field to get service IP alloc range object (which paves the way to per-network ip pool)

Per-network Service IP allocation is depicted in [service IP section, multi-tenant network design proposal](https://github.com/futurewei-cloud/arktos/blob/master/docs/design-proposals/multi-tenancy/multi-tenancy-network.md#service-ip).

**How to verify**:
in fact no change of behavior (unless the feature gate is explicitly turn on).
1. start arktos up by ```./hack/arktos-up.sh```
2. create a service (this works for now) with following yaml
```yaml
apiVersion: v1
kind: Service
metadata:
  name: my-nginx-no-net
  labels:
    run: my-nginx
spec:
  ports:
  - port: 80
    protocol: TCP
  selector:
    run: my-nginx
```
3. verify the named service created inside k8s service IP range
``` 
$ ./cluster/kubectl.sh get svc
NAME           TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)   AGE
kubernetes     ClusterIP   10.0.0.1     <none>        443/TCP   41m
my-nginx-no-net   ClusterIP   10.0.0.105   <none>        80/TCP    8m3s
```
If you want to further verify service that has explicit network, continue with
4. create network CRD by ```./cluster/kubectl.sh create -f ./pkg/controller/artifacts/crd-network.yaml```
5. create network test2 using yaml
```yaml
kind: Network
metadata:
  name: test2
spec:
  type: flat
  service:
    ipam: Arktos
```
6. create a service associated to test2 network, using yaml
```yaml
apiVersion: v1
kind: Service
metadata:
  name: my-nginx-svc
  labels:
    run: my-nginx
spec:
  ports:
  - port: 80
    protocol: TCP
  selector:
    run: my-nginx
```
7.verify the service also gets IP address from k8s pool
```
$ ./cluster/kubectl.sh get svc my-nginx-svc
NAME           TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)   AGE
my-nginx-svc   ClusterIP   10.0.0.185   <none>        80/TCP    26m
``` 

* This PR is intended to keep small and focused for ease of review. There will be another PR providing the method implementation and additional test code. 
* As this PR does not change behavior (when the feature gate if off by default), no new unit test cases is written; test code was changed only to reflect the newly refactored method in prod package.

Partial: No by default; Yes potentially.

```release-note
If kube-apiserver start with --feature-gates=PerNetworkServiceIPAlloc=true, service IP address shall be allocated from network service IP pool (which is on the way)
```


staging/src/k8s.io/arktos-ext/pkg/apis/arktosextensions/v1/types.go
```diff
@@ -93,6 +93,9 @@ const (
 	NetworkLabel = "arktos.futurewei.com/network"
 	// network related annotation keys
 	NetworkReadiness = "arktos.futurewei.com/network-readiness"
+
+	// NetworkDefaul is the default network name
```
chenqianfzh, line 97: NIt type: NetworkDefaul

pkg/registry/core/service/storage/rest.go
```diff
@@ -222,34 +232,27 @@ func (rs *REST) Export(ctx context.Context, name string, opts metav1.ExportOptio
...
 	net := network.(*unstructured.Unstructured)
-	netSpec := net.Object["spec"].(map[string]interface{})
-	if s, ok := netSpec["service"]; ok {
-		netSvc := s.(map[string]interface{})
-		if v, ok := netSvc["ipam"]; ok {
-			if vs, ok := v.(string); ok {
-				return vs == string(arktosextensionsv1.IPAMExternal), nil
-			}
-		}
+	if ipam, ok, err := unstructured.NestedString(net.Object, "spec", "service", "ipam"); err == nil && ok {
```
chenqianfzh, line 254: nice refactoring.

pkg/registry/core/service/storage/rest.go
```diff
@@ -125,6 +126,15 @@ var (
 	_ rest.StorageVersionProvider = &REST{}
 )
 
+// getServiceIPs gets ip allocator object suited for the service on basis of its associated network
+func (rs *REST) getServiceIPs(service *api.Service) (ipallocator.Interface, error) {
```
chenqianfzh, line 130: I have a bit concern about the naming. Here the function does not return the IPs, instead it returns the allocator. Is it better to rename the func as getServiceIPAlloc?

staging/src/k8s.io/arktos-ext/pkg/apis/arktosextensions/v1/types.go
h-w-chen: will correct

pkg/registry/core/service/storage/rest.go
h-w-chen: point made. will change the method name as suggested.

PR #487, opened 2020-07-16
Fixed the kube-up start failure issue

jshaofuturewei:
Since we use service group  yaml files to set up informer ranges, there is no need to maintain a update range via apiserver.config. Removing the range from apiserver.config and start_apiserver function


hack/apiserver.config
```diff
@@ -1,27 +1,27 @@
...
+/registry/poddisruptionbudgets/,,
+/registry/podpresets/,,
+/registry/pods/,,
+/registry/podsecuritypolicies/,,
+/registry/podsecuritypolicy/,,
+/registry/podtemplates/,,
+/registry/replicasets/,,
+/registry/resourcequotas/,,
+/registry/statefulsets/,,
+/registry/storageclasses/,,
```
Sindica, line 25: Please remove /registry/storageclasses/ since it does not have namespace or tenant data

PR #490, opened 2020-07-16
arktos-up script to set forward chain accept by default

### LINKED ISSUE #380: cannot ping other pods from inside a pod
**What happened**:
ping other pod from inside a pod gets back no replies

ping gets back icmp echo packets

context: tenant admin
start  a nginx pod, with yaml spec
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: nginx
spec:
  containers:
  - name: nginx
    image: nginx
```

start another pod, by yaml spec
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: k8s-vanilla-pod
  labels:
    app: vanilla
spec:
  containers:
  - name: vanilla-container
    image: busybox
    command: ['sh', '-c', 'echo Hello Arktos! && sleep 3600']
```
Both are in Running state:
```
$ ./kubectl get pods -o wide
NAME              HASHKEY               READY   STATUS    RESTARTS   AGE    IP            NODE               NOMINATED NODE   READINESS GATES
k8s-vanilla-pod   4081895494302239416   1/1     Running   123        5d3h   10.88.5.152   ip-172-31-27-125   <none>           <none>
nginx             5554014411665225714   1/1     Running   0          10s    10.88.26.82   ip-172-31-27-125   <none>           <none>
```
enter in k8s-vanilla-pod to ping nginx pod, no reply at all
```
$ kubectl exec -it k8s-vanilla-pod sh
   / # ping 10.88.26.82
   PING 10.88.26.82 (10.88.26.82): 56 data bytes
```

system: the bug-bash env (https://18.236.170.164:6443);
security context: my allocated tenat, as tenant-admin

**Environment**:
Client Version: version.Info{Major:"0", Minor:"2+", GitVersion:"v0.2-99+a93f32b74c4285", GitCommit:"a93f32b74c42856be37c74dea42bad41d46ba45b", GitTreeState:"clean", BuildDate:"2020-06-17T01:12:15Z", GoVersion:"go1.12.17", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"0", Minor:"2+", GitVersion:"v0.2-84+29b3b2f1921680-dirty", GitCommit:"29b3b2f1921680c2511abe93c8f58d2acba272a2", GitTreeState:"dirty", BuildDate:"2020-06-10T21:45:04Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"linux/amd64"}
internal bug-bash env
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
bridge
- Others:

h-w-chen:
**What type of PR is this?**
> /kind other (environment setup related issue) 

to set firewall rule allowing traffic forward by default, in the process of arktos-up to set up local dev env

The solution is based on [blog about bridge traffic also traversed iptables rules](https://news.ycombinator.com/item?id=16427686):
_When the sysctl net.bridge.bridge-nf-call-iptables=1 is set (the default on modern Linux kernels), bridged packets will traverse iptables rules.
```  iptables -I FORWARD -j DROP```
by adding that rule, two containers on the same host cannot communicate with each other despite the fact that their communication did not require any traditional routing from the host. This is because the bridge traffic also traversed iptables rules._

Fixes #380 

No for Arktos functionality

### CODE REVIEW

File: hack/arktos-up.sh [hunk 1]
```diff
@@ -59,6 +59,10 @@ if [[ ! -e "${CONTAINERD_SOCK_PATH}" ]]; then
   exit 1
 fi
 
+echo "Ensuring firewall to allow traffic forward by default"
```
XiaoningDing, line 62: Can you add reasons why this is the right way to solve #380?  And also is #380 a bug or an issue of CNI configuration?

h-w-chen: The PR description was modified to include the source of information the fix is based on. In short, the bridge traffic also traversed iptables rules.

This seems a CNI configuration related environment issue.

XiaoningDing: suggest moving this piece to the inside of "install_cni_bridge". It's probably more clear and won't impact host settings if we don't use beidges.

h-w-chen: will update as suggested.

PR #492, opened 2020-07-16
per-network service IP allocation - part 2: implement method of getting service ip allocator per network

h-w-chen:
**What type of PR is this?**
/kind feature

This is part 2 for per-network service IP allocation, follwoing [part 1](https://github.com/futurewei-cloud/arktos/pull/484)

1. provides the method implementation of per network service IP allocation;
2. adds test code;

Per-network Service IP allocation is depicted in [service IP section, multi-tenant network design proposal](https://github.com/futurewei-cloud/arktos/blob/master/docs/design-proposals/multi-tenancy/multi-tenancy-network.md#service-ip).

1. start etcd
2. start kube-apiserver with the feature gate on: ```sudo ./kube-apiserver --etcd-servers=http://127.0.0.1:2379 --feature-gates=PerNetworkServiceIPAlloc=true```
3. create network CRD by ```kubectl create -f <repo-root>/pkg/controller/artifacts/crd-network.yaml```
4. create network test2 using yaml
```yaml
kind: Network
metadata:
  name: test2
spec:
  type: flat
  service:
    ipam: Arktos
    cidrs: [192.168.100.0/24]
```
5. create a service associated to test2 network, using yaml
```yaml
apiVersion: v1
kind: Service
metadata:
  name: my-nginx-svc
  labels:
    run: my-nginx
spec:
  ports:
  - port: 80
    protocol: TCP
  selector:
    run: my-nginx
```
6.verify the service also gets IP address from k8s pool
```
$ ./cluster/kubectl.sh get svc my-nginx-svc
NAME           TYPE        CLUSTER-IP   EXTERNAL-IP   PORT(S)   AGE
my-nginx-svc   ClusterIP   192.168.100.68   <none>        80/TCP    2s
``` 

Partial: No by default; Yes potentially.

```release-note
If kube-apiserver start with --feature-gates=PerNetworkServiceIPAlloc=true, service IP address shall be allocated from network service IP pool.
```


pkg/registry/core/service/storage/rest.go
```diff
@@ -876,3 +890,80 @@ func collectServiceNodePorts(service *api.Service) []int {
 	}
 	return servicePorts
 }
+
+func (rs *REST) getNetworkServiceIPAlloc(service *api.Service) (ipallocator.Interface, error) {
+	netName, ok := service.Labels[arktosv1.NetworkLabel]
+	if !ok {
```
chenqianfzh, line 896: Do we need also set the netName to "default" if it is empty?

pkg/registry/core/service/storage/rest.go
```diff
@@ -876,3 +890,80 @@ func collectServiceNodePorts(service *api.Service) []int {
...
+	}
+
+	networkUnstructured := network.(*unstructured.Unstructured)
+	var cidrs []string
+	cidrs, ok, err = unstructured.NestedStringSlice(networkUnstructured.Object, "spec", "service", "cidrs")
+	if err != nil {
+		return nil, fmt.Errorf("failed to identify spec.service.cidrs of network %s of tenant %s: %v", netName, service.Tenant, err)
+	}
+	if !ok || len(cidrs) == 0 {
+		return nil, fmt.Errorf("no service cidr found for network network %s of tenant %s", netName, service.Tenant)
```
chenqianfzh, line 918: nit: repeated "network" in the error message

pkg/registry/core/service/storage/rest.go
```diff
@@ -876,3 +890,80 @@ func collectServiceNodePorts(service *api.Service) []int {
...
+		return ipAllocator.(ipallocator.Interface), nil
+	}
+
+	var err error
+	ipAllocator := ipallocator.NewAllocatorCIDRRange(cidr, func(max int, rangeSpec string) allocator.Interface {
+		mem := allocator.NewAllocationMap(max, rangeSpec)
+		etcd := serviceallocator.NewEtcd(mem, etcdBaseKey, api.Resource("serviceipallocations"), rs.backendStorageConfig)
+
+		var snapshot *api.RangeAllocation
+		if snapshot, err = etcd.Get(); err != nil || snapshot.Range == "" {
```
chenqianfzh, line 956: Why we need to try Get() before CreateOrUpdate()? etcd is just created, Get() should not find anything.

pkg/registry/core/service/storage/rest_test.go
```diff
@@ -2437,3 +2449,59 @@ func TestServiceRegistryCreateWithoutExistingNetwork(t *testing.T) {
 		t.Fatalf("expected error; got none")
 	}
 }
+
+func TestServiceRegistryCreateWithNetworkOfArktosIPAM(t *testing.T) {
+	origFeatureGate := newFeatureGateWithPerNetworkServiceIPAlloc()
+	defer func() {
+		utilfeature.DefaultFeatureGate = origFeatureGate
```
chenqianfzh, line 2456: Why we need set utilfeature.DefaultFeatureGate = origFeatureGate when the test case exit?

pkg/registry/core/service/storage/rest_test.go
```diff
@@ -2437,3 +2449,59 @@ func TestServiceRegistryCreateWithoutExistingNetwork(t *testing.T) {
...
+
+	storage, registry, server := NewTestREST(t, nil)
+	storage.networks = &dummyNetworkGetter{ipamType: "Arktos", cidr: "3.4.5.0/24"}
+	_, ipRange, _ := net.ParseCIDR("3.4.5.0/24")
+	defer server.Terminate(t)
+
+	svc := &api.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "foo",
+			Tenant:    "system",
```
chenqianfzh, line 2467: As we are working on multi-tenancy, it is better test on a non-system tenant.

pkg/registry/core/service/storage/rest_test.go
```diff
@@ -2437,3 +2449,59 @@ func TestServiceRegistryCreateWithoutExistingNetwork(t *testing.T) {
...
+	if !ipRange.Contains(ip) {
+		t.Errorf("expected ip in %s; got %s", ipRange, created_service.Spec.ClusterIP)
+	}
+
+	srv, err := registry.GetService(ctx, svc.Name, &metav1.GetOptions{})
+	if err != nil {
+		t.Errorf("unexpected error: %v", err)
+	}
+	if srv == nil {
+		t.Errorf("Failed to find service: %s", svc.Name)
```
chenqianfzh, line 2498: should be Fatalf here. If srv==nil, net.ParseIP(srv.Spec.ClusterIP) is doomed to run into failure.

pkg/registry/core/service/storage/rest_test.go
```diff
@@ -2437,3 +2449,59 @@ func TestServiceRegistryCreateWithoutExistingNetwork(t *testing.T) {
...
+	created_service := created_svc.(*api.Service)
+
+	ip := net.ParseIP(created_service.Spec.ClusterIP)
+	if !ipRange.Contains(ip) {
+		t.Errorf("expected ip in %s; got %s", ipRange, created_service.Spec.ClusterIP)
+	}
+
+	srv, err := registry.GetService(ctx, svc.Name, &metav1.GetOptions{})
+	if err != nil {
+		t.Errorf("unexpected error: %v", err)
```
chenqianfzh, line 2495: Should be Fatalf. net.ParseIP(srv.Spec.ClusterIP) has no chance to work.

pkg/registry/core/service/storage/rest_test.go
```diff
@@ -2437,3 +2449,59 @@ func TestServiceRegistryCreateWithoutExistingNetwork(t *testing.T) {
...
+	if srv == nil {
+		t.Errorf("Failed to find service: %s", svc.Name)
+	}
+
+	ipFromGet := net.ParseIP(srv.Spec.ClusterIP)
+	if !ipFromGet.Equal(ip) {
+		t.Fatalf("expected %s; got %s from Get op", ip, ipFromGet)
+	}
+
+	t.Logf("allocated service ip = %s", ip)
```
chenqianfzh, line 2506: It a bit weird to see a t.Logf in the last line of a test case. If the test passes, no log will be printed. If the test case fail, will this line be hit?

h-w-chen: no need to have log statement here indeed. will remove.

pkg/registry/core/service/storage/rest.go
```diff
@@ -876,3 +890,80 @@ func collectServiceNodePorts(service *api.Service) []int {
...
+			return rs.serviceIPs, nil
+		}
+
+		return nil, fmt.Errorf("cannot find network %s of tenant %s: %v", netName, service.Tenant, err)
+	}
+
+	networkUnstructured := network.(*unstructured.Unstructured)
+	var cidrs []string
+	cidrs, ok, err = unstructured.NestedStringSlice(networkUnstructured.Object, "spec", "service", "cidrs")
+	if err != nil {
```
XiaoningDing, line 914: I remember in our last discussion we mentioned service CIDR probably isn't required for flat network objects.  Will these check fail for flat network?

pkg/registry/core/service/storage/rest_test.go
h-w-chen: good point. will change to use Fatalf

pkg/registry/core/service/storage/rest.go
```diff
@@ -876,3 +890,80 @@ func collectServiceNodePorts(service *api.Service) []int {
...
+	if !ok || len(cidrs) == 0 {
+		return nil, fmt.Errorf("no service cidr found for network network %s of tenant %s", netName, service.Tenant)
+	}
+
+	// todo: to use multiple cidr ranges in the future
+	// for now only first cidr take effect
+	cidr := cidrs[0]
+	var ipNet *net.IPNet
+	if _, ipNet, err = net.ParseCIDR(cidr); err != nil {
+		return nil, fmt.Errorf("%q is not a valid range of network %s of tenant %s: %v", cidr, netName, service.Tenant, err)
```
XiaoningDing, line 926: "not a valid range of network" sounds strange.  You mean it's not a valid range of network IPs?

pkg/registry/core/service/storage/rest.go
```diff
@@ -876,3 +890,80 @@ func collectServiceNodePorts(service *api.Service) []int {
...
+	var cidrs []string
+	cidrs, ok, err = unstructured.NestedStringSlice(networkUnstructured.Object, "spec", "service", "cidrs")
+	if err != nil {
+		return nil, fmt.Errorf("failed to identify spec.service.cidrs of network %s of tenant %s: %v", netName, service.Tenant, err)
+	}
+	if !ok || len(cidrs) == 0 {
+		return nil, fmt.Errorf("no service cidr found for network network %s of tenant %s", netName, service.Tenant)
+	}
+
+	// todo: to use multiple cidr ranges in the future
```
XiaoningDing, line 921: Please track these as issues so that we won't forget them.  Thanks.

pkg/registry/core/service/storage/rest_test.go
h-w-chen: will change to Fatalf as suggested

pkg/registry/core/service/storage/rest_test.go
h-w-chen: got your point. However, this case is purposed as the most basic scenario for feature of per-network service ip allocation; in case it fails, we know for sure it is broken at this feature, not around multi-tenancy.

pkg/registry/core/service/storage/rest_test.go
h-w-chen: got your point. However, this case is purposed as the most basic scenario for feature of per-network service ip allocation; in case it fails, we know for sure it is broken at this feature, not around multi-tenancy.

pkg/registry/core/service/storage/rest.go
h-w-chen: etcd may still have data. if it is there, will use the existent data; otherwise to create one.

pkg/registry/core/service/storage/rest.go
h-w-chen: will remove the extra

pkg/registry/core/service/storage/rest.go
h-w-chen: good catch. will update.

pkg/registry/core/service/storage/rest.go
h-w-chen: if flat network does not turn on the feature gate of per-network-service-ip-alloc (the code is in previors part 1 PR), it won't fail. 
this PR is all about turning on the feature, and it will alloc service ip from per-network pool.

pkg/registry/core/service/storage/rest.go
h-w-chen: I meant for "not a valid service ip range for network object". will reword to make it more clear.

pkg/registry/core/service/storage/rest.go
h-w-chen: issue #495 was opened to track it.

h-w-chen:
thanks.

PR #496, opened 2020-07-17
add checks for existance before calling apiserver 

pdgetrf:
Fix for https://github.com/futurewei-cloud/arktos/issues/453

### Problem:
Tenant controller is syncing every 5 minutes during which it will attempt to provision resources through client-go. This adds unnecessary loads to apiserver.

### Fix:
Add checks for resource existence before calling client-go. The checks are against informer cache instead of apiserver directly

### Tests performed:
Verified initial provisioning work correctly without regression, and during periodical syncing, resources creation are skipped

image

### Note to reviewers:
Network object currently is not having this cache check because 1) it's not actually provisioning resource as in "No default network template path; skipped the creation of default network in tenant "system" and 2) informer for network object is in a different format which would taking longer to fix. Considering we are adding this for perf test, it's decided to be left to another PR to fix. There's no perf impact for this at the moment.

I'm also looking into adding additional tests for the scenarios where resource creation is skipped


pkg/controller/tenant/tenant_controller.go
```diff
@@ -278,33 +300,61 @@ func (tc *TenantController) createNamespaces(tenant string) (error, bool) {
...
-		ObjectMeta: metav1.ObjectMeta{
-			Name:   InitialClusterRoleBindingName,
-			Tenant: tenant,
-		},
-		Subjects: []rbacv1.Subject{
-			{Kind: rbacv1.UserKind, Name: initialClusterRoleUser},
-		},
-		RoleRef: rbacv1.RoleRef{APIGroup: rbacv1.GroupName, Kind: "ClusterRole", Name: InitialClusterRoleName},
+	if !shouldSkip {
+		// tenant admin act as cluster admin for tha tenant
```
chenqianfzh, line 317: Nit: typo "tha"

pkg/controller/tenant/tenant_controller.go
```diff
@@ -267,6 +280,15 @@ func (tc *TenantController) createDefaultNetworkObject(tenantName string) (error
 func (tc *TenantController) createNamespaces(tenant string) (error, bool) {
 	failures := []error{}
 	for _, nsName := range tenantDefaultNamespaces {
```
chenqianfzh, line 282: Need to check the tenant's status is active before creating. Otherwise we might run into weird race condition issue when deleting a tenant. 

You can refer to the following logic in createDefaultNetworkObject():
      if tenant.Status.Phase == v1.TenantTerminating {
		klog.Infof("Tenant %q is terminating; skipped the creation of default network", tenantName)
	}

pkg/controller/tenant/tenant_controller.go
```diff
@@ -278,33 +300,61 @@ func (tc *TenantController) createNamespaces(tenant string) (error, bool) {
...
 
 func (tc *TenantController) createInitialRoleAndBinding(tenant string) (error, bool) {
-	// tenant admin act as cluster admin for tha tenant
-	role := &rbacv1.ClusterRole{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:   InitialClusterRoleName,
-			Tenant: tenant,
-		},
-		Rules: initialClusterRoleRules(),
+
```
chenqianfzh, line 303: Same: Need to check the tenant's status is active before creating. Otherwise we might run into weird race condition issue when deleting a tenant.

test/integration/tenantcontroller/tenantcontroller_test.go
```diff
@@ -64,13 +64,15 @@ func setup(t *testing.T) (*httptest.Server, framework.CloseFunc, *tenantcontroll
 	networkClient := arktos.NewForConfigOrDie(configs)
```
chenqianfzh, line 64: I think it is better to add a test case to make sure no resources are created when tenant is terminating.

pkg/controller/tenant/tenant_controller.go
```diff
@@ -115,10 +125,13 @@ func NewTenantController(
 		},
 		resyncPeriod,
```
Sindica, line 126: I am really worried about using AddEventHandlerWithResyncPeriod and resyncPeriod as 5 min. This meaning every five min all tenant creation events will be reevaluated again. When we have more than ten thousand tenants, actively going over all tenants and re-evaluate will consume a lot of resources.

pdgetrf: the resync period is for actions like tenant deletion to work with a periodic check.

pkg/controller/tenant/tenant_controller.go
```diff
@@ -80,10 +88,12 @@ type TenantController struct {
...
 // NewTenantController creates a new iinstance of tenantcontroller
-func NewTenantController(
-	kubeClient clientset.Interface,
+func NewTenantController(kubeClient clientset.Interface,
 	tenantInformer coreinformers.TenantInformer,
-	resyncPeriod time.Duration,
+	namespaceInformer coreinformers.NamespaceInformer,
+	clusterRoleInformer rbacinformers.ClusterRoleInformer,
+	clusterRoleBindingInformer rbacinformers.ClusterRoleBindingInformer,
+	resyncPeriod time.Duration,	// split this controller into tenant creation and deletion controllers if resyncPeriod causes performance degradation
```
pdgetrf, line 96: @chenqianfzh @Sindica added notes here for perf concerns.

pkg/controller/tenant/tenant_controller.go
pdgetrf: fixed

PR #508, opened 2020-07-22
Fixed the a flaky test returns c1 or c2 randomly.

jshaofuturewei:


cmd/kube-scheduler/app/options/options_test.go
```diff
@@ -409,14 +409,14 @@ users:
 					AlwaysAllowPaths:             []string{"/healthz"}, // note: this does not match /healthz/ or /healthz/*
 				},
 			},
-			expectedUsername: "none, http",
+			expectedUsername: []string{"none, http"},
```
Sindica, line 412: Does this need to be "none, http" instead of "none" or "http"?

jshaofuturewei: Please check the mock codes and the  "none, http" is hard-coded here.

insecureserver := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
		username, _, _ = req.BasicAuth()
		if username == "" {
			username = "none, http"
		}
		w.WriteHeader(200)
		w.Write([]byte(`ok`))
	}))

Sindica: Correct, I asked the question because reviously you changed it to []string{"none", "http"}.

jshaofuturewei: I checked the history and there is no []string{"none", "http"} change.

cmd/kube-scheduler/app/options/options_test.go
```diff
@@ -515,7 +515,7 @@ users:
 					AlwaysAllowPaths:             []string{"/healthz"}, // note: this does not match /healthz/ or /healthz/*
 				},
 			},
-			expectedUsername: "c1",
+			expectedUsername: []string{"c1", "c2"},
```
Sindica, line 518: We are not actively supporting multiple kubeconfig from command line. Can this test be disabled instead?

PR #510, opened 2020-07-22
Fixing issue 505

chenqianfzh:
Fixing issue https://github.com/futurewei-cloud/arktos/issues/505.

With this fix, the error "Node ***** not found" error is gone.

Verified: the perf-test is running with this fix.


staging/src/k8s.io/client-go/tools/cache/listwatch.go
```diff
@@ -97,7 +101,11 @@ func NewListWatchFromClientWithMultiTenancy(c Getter, resource string, namespace
 // Option modifier is a function takes a ListOptions and modifies the consumed ListOptions. Provide customized modifier function
 // to apply modification to ListOptions with a field selector, a label selector, or any other desired options.
 func NewFilteredListWatchFromClient(c Getter, resource string, namespace string, optionsModifier func(options *metav1.ListOptions)) *ListWatch {
-	return NewFilteredListWatchFromClientWithMultiTenancy(c, resource, namespace, optionsModifier, metav1.TenantSystem)
+	tenant := metav1.TenantSystem
```
Sindica, line 104: Isn't Sonya using default tenant, which is system tenant, in her testing?

chenqianfzh: This change is that when watching/listing cluster-scope resources like "nodes", the url link should be "/api/v1/nodes", instead of "/api/v1/tenants/system/nodes".  Also, in cases that the K8s would like to list/watch across all the namespaces (which means the whole cluster in K8s), we should let it do across all the tenants ( which is what whole cluster means in Arktos).

It is not about which tenant you are using.

staging/src/k8s.io/client-go/tools/cache/listwatch.go
```diff
@@ -81,9 +81,21 @@ type Getter interface {
 	RESTClients() []restclient.Interface
 }
 
+// For backwards compatibility concerns, if the lister/watcher operates across all namespaces (namely cluster-scoped),
+// we should let it work on all tenants.
+// If it targets a specific namespaces, we should let it target at the system tenant.
+func InferTenantFromNamespace(namespace string) string {
```
XiaoningDing, line 87: Why namespaceALL will lead to tenantALL?  A regular tenant user can still list/watch for namespaceALL, but it should be limited to his own tenant space, right?

staging/src/k8s.io/client-go/tools/cache/listwatch.go
XiaoningDing: "/api/v1/tenants/system/nodes" looks a totally legitimate path to me. What's wrong here?

staging/src/k8s.io/client-go/tools/cache/listwatch.go
XiaoningDing: And we can't assume the default tenant is system tenant.

chenqianfzh: Per offline discussion, we understand that this change unblocks the perf tests but it may have some issue for regular tenants. We cannot assume the tenant is "system" when the tenant info is not given.

Discussion results are recorded in https://github.com/futurewei-cloud/arktos/issues/515.

PR #511, opened 2020-07-22
add internal IP to server cert

sonyafenge:
**What type of PR is this?**

> /kind bug

previously, kube-up only use external ip to server cert, this PR is adding internal IP to server cert.

Fixes #
https://github.com/futurewei-cloud/arktos/issues/498

```release-note

```


cluster/gce/util.sh
```diff
@@ -2738,20 +2738,22 @@ function create-master() {
   declare -a APISERVER_RESERVED_IP
   declare -a APISERVER_RESERVED_INTERNAL_IP
   declare -a APISERVER_NAME
-  CREATE_CERT_APISERVER_IP=""
+
+  # add all external and internal IP to cert
+  CREATE_CERT_SERVER_IP="${MASTER_RESERVED_INTERNAL_IP}"
```
Sindica, line 2743: Why rename CREATE_CERT_APISERVER_IP to CREATE_CERT_SERVER_IP?

sonyafenge: previous, this var only used for additional apiserver, but now, it is used for all servers IP including master internal ip, additional apiserver internal ip and external ip. and also may include extra ETCD cluster ip in the future, so rename to server_IP

PR #514, opened 2020-07-22
Update design doc with VM specific details

yb01:
**What type of PR is this?**
> /kind documentation

Updated the design doc for VMs


docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
```diff
@@ -319,3 +341,129 @@ Other components:
...
+   
+4. Some limitations/constraints for VM workload types
+
+    1. Memory device size: 
+       Currently it is single size, with 128MiB each. i.e. if a VM patch increases the memory as 1Gi, 8 memory devices 
+       will be attached to the VM. The operations will still in one UpdateContainerResource() call though. i.e. Arktos VM 
+       runtime service will handle the attach operation internally with the calculation of memory increment in the patch.
+       
+    2. Max memory scale-up value
+       Currently a VM memory slots are set to 16 in Arktos VM runtime, so 16 * 128 Mi, which is 2Gi max scale-up value.
```
XiaoningDing, line 383: Max 2G memory? Seriously? :). It seems too small

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
```diff
@@ -319,3 +341,129 @@ Other components:
...
+       This is assumed the host has plenty allocatable memory.
+       
+    3. Guest OS Kernel requirements
+       Guest OS kernel must support device hotPlug and hotUnplug.
```
XiaoningDing, line 387: Is it Linux only? For example, will a Windows VM pod eligible for vertical scaling?

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
yb01: this is for the increase for the current implementation. it can be adjusted but i am not sure how much we expect it to be.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
yb01: this is a TBD for requrements for the vm support in general. currently it is Linux that we are mainly testing on.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
```diff
@@ -130,6 +140,15 @@ RestartContainer policy takes precedence over NoRestart policy.
 Additionally, if RestartPolicy is 'Never', ResizePolicy should be set to
 NoRestart in order to pass validation.
 
+It is proposed to allow users to set the MAX resource to better control the behavior of the resource resizing.
+This could be a value or percentage specified by users. prior to this policy setting, the ceiling of resources, such as
+CPU and memory will be set to the allocatable value at the host and rely the proper CGroup to control the VM resource
+consumptions.
```
XiaoningDing, line 146: I don't understand why using allocable as max leads to relying on cgroup. Maybe I miss something. Do we need cGroup control when we use "2*requested" for max vCPU, what's the difference here?

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
```diff
@@ -130,6 +140,15 @@ RestartContainer policy takes precedence over NoRestart policy.
...
+
+TBD:
+Due to different technologies limit different level of memory change size, e.g. Ballooning driver has 4K page
```
XiaoningDing, line 149: From the description it seems ballooning has less capability than memory hotplug. What's the memory increase limit for ballooning?  And I'm the under the impression that we are using memory ballooning.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
yb01: in terms of possibility to exceed the "max" resource level for the vm. 2x or host allocatable is the same. 
the case here is how to determine the max so our code logic and libvirt won't set resource to exceed the "max" of it. this is the reason it is proposed an extra setting in the spec.

we need this for cgroup max-xxx properties. if set to max as host allotable, it is essentially disables the CG control.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
yb01: this is one of decision we will discuss in the review. Max setting is the high mark.  however, Ballooning is essentially targeted for the hypervisor to "get" memory from the VM guest. let's discuss this in the meeting in details.
Hotplug/unplug memory device should be a good feature for Arktos regardless eventually which technology was used for vertical scaling here.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
```diff
@@ -319,3 +341,129 @@ Other components:
...
+
+  If Kubelet enables its CPU manager and set the cpus to a particular pod. the cpuset is sent to the runtime via the 
+  UpdateContainerResource() and the runtime will honer the cpusets in both VM definition and its CGroup.
+
+  Two new APIs, SetVcpus() and SetCurrentMemory() were added to support VM vertical scaling, which extends the call to 
+  Libvirt APIs to update the VM domain xml.
+
+	// Update vcpu for a give domain
+	SetVcpus(int) error
+	// Update current memory for a given domain
```
XiaoningDing, line 406: I leave a comment in code PR that suggests renaming the function to "ChangeXXX", "UpdateXXX" or similar names. The current name implies the input parameter is the new amount, but actually it's the delta.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
```diff
@@ -319,3 +341,129 @@ Other components:
...
+### Changes in Arktos-vm-runtime-service (CRI for VM workload types)
+
+#### Specific for VM workload
+
+1. Hotplug(hotUnplug) vs Memory Ballooning driver for Memory resizing
+
+   Currently per our knowledge these are two technologies that supports live memory resizing for VMs. Both have been there for
+   Linux for years, and both has pros and cons in supporting memory resizing for VMs. There are also some proposals to combine
+   them to better support VM resizing for Linux VMs.
+
```
XiaoningDing, line 353: Can you elaborate this part? Curious to know the key reason why we choose memory hotplug over balooning.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
XiaoningDing: A memoryDevice simulates a memory chip in real world.  128M is too small for today's world.  We can change it to a larger value.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
cathyhongzhang: Could you clarify what is the motivation to let the user define the MAX?  Even the user sets the MAX, your code still needs to check "allocatable value at the host" at the time point of resizing in order to make sure there is enough resource for the new size, right? Also in quite some scanarios, it is hard for the user to pre-define the MAX. for example, in serverless usage scenarios, the serverless function controller does not know the future traffic pattern and would like to resize to the "max allocable value at the host" when there is a burst of requests. BTW, do you use CGroup to control the resource boundary of the POD during the resizing?

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
```diff
@@ -319,3 +341,129 @@ Other components:
...
+   In Arktos VM runtime, it is proposed to use hotplug(hotUnplug) in current release for the following reasons:
+
+    1. Numa support
+    2. Less memory fragmentation
+    3. Better control from VM spec patching
+
```
yb01, line 359: need more investigateion and clearly define the pro/cons

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
yb01: consinder renaming the api names

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
```diff
@@ -55,17 +58,23 @@ some, like VMs are more sensitive, especially when using low number of Pod repli
...
 * Primary: allow to change Pod resource requests & limits without restarting
   its Containers or VMs.
 * Secondary: allow actors (users, VPA, StatefulSet, JobController) to decide
   how to proceed if in-place resource resize is not possible.
+* For VM, NUMA awareness of memory increase 
+* For VM, x86 CPU architecture must be supported, ARM can be stretch goal
+* For VM, major vendors of linux guest OS,
 
 ### Non-Goals
+* For VM, windows guest nor windows host to be supported
```
yb01, line 75: consiner support VM

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
```diff
@@ -319,3 +341,129 @@ Other components:
...
+definition in the Libvirt metadata store. it is runtime's responsibility to keep them in sync. When VM resource is 
+created or updated, CGroup, vmConfig, domain definition are persisted in such order:
+
+1. Update CGroup
+2. Update VMResources( vmConfig and domainDef)
+
+If update VMResources operation failed, then the CGroup update operation will be rolled back and error will be raised.
+
+During UpdateVmResources(), no rollback is enforced. The ContainerStatus() function will sync up the vmConfig and 
+Libvirt domain definition if they are out of sync.
```
cathyhongzhang, line 446: why not do rollback? If these info are out of sync for some period of time, wouldn't it incur some side effect on some operation?

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
```diff
@@ -319,3 +341,129 @@ Other components:
...
+	// Update current memory for a given domain
+	SetCurrentMemory(int64) error
+	
+  Both of them has similar workflow a below:
+
+    1. Get the linux resource from the UpdateContainerResource request
+    2. Get the current configuration from the VM 
+    3. Calculate the delta and determine Attach or Detach operation ( for CPU it will be enable and disable operation)
+    4. Call the Libvirt API to update the domain.
+
```
cathyhongzhang, line 415: Is it one libvirt API to update both CPU and memory domain config? If it is multiple API calls, then atomicity needs to be handled.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
```diff
@@ -319,3 +341,129 @@ Other components:
...
+       Some times the OS just cannot migrated the app and free up the memory pages for the memory device. This can cause
+       memory unplug failures. lower the device size can mitigate but still not guaranteed. The Kubelet will retry in this case
+       till it successfully unplugged.
+
+#### API changes
+  As described in the "CRI changes" section, the Arttos-vm-runtime UpdateContainerResource() were implemented to update
+  the VM CGroup settings and update the VM domain definition with the new resource configuration.
+
+  If Kubelet enables its CPU manager and set the cpus to a particular pod. the cpuset is sent to the runtime via the 
+  UpdateContainerResource() and the runtime will honer the cpusets in both VM definition and its CGroup.
```
cathyhongzhang, line 399: typo: honor

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
yb01: per the discussion today. will update the design with a "multiple devices size support" approach.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
yb01: Arktos will support windows VM. will add this back to our goal to support vertical scaling as well.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
yb01: I think there are a few case we need to clarify:
1.  should the user be expose the option to control the max? some app or service might just want to throttle when a certain ceiling is reached instead of keep allocated resources. for example, the app have fixed size thread pool, the app wont need more cpus than the pool size in general. or they want to control the behavior of the app by limiting the  resources. form this case, we can default to host allocatable, but let user to set if they want.

2. from our own control in the runtime to setup the CG for the VM. for example, for memory usage, it is intended to set the memory_limit_limits_inbytes to the MAX. if the max is host allocatable, this setting will be useless.

we might logically still hold up with the limits in VM resource, but the logical concept should be there in the code.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
yb01: You have a point. I am working on the handler design. will add rollback CG update IF the domain update failed. some details of the reasons for the current design.

First of all, ideally the 3 operations ( CG update, libvirt domain update, runtime config metadata update ) should be in one transaction.

The currently approach is "eventually consistency" mainly due to the downstream behavior of update libvirt domain. resource change ( especially decrease mem ) are done asynchronously and the actual status is available in detachDevice event handler.  Note that the runtime metadata will not change if there is any error returned from the domain update. also, the synchronization logic took the domain definition as the source of truth if there is any differences between libvirt domain and vm runtime metadata.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
yb01: Unfortunately no. we have to handle them separately.
note that they won't need to done in a atomic way. the resource update can be separated. kubelet will check if the desired state of the spec is reached or not and if not, it will sent it again. so the atomic level is on single resource type, not combined.

i will discuss with Vinay to see if we can make an agreement between Kubelet and runtime. if kubelet will send mixed request, runtime will add logic to handle this case.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
yb01: thanks. will change.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
yb01: Updated with some details for the design decision justifications.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
yb01: I have a feeling we might refer to different MAX setting here. the doc has been changed/updated. i explicitly added a section to explain the few settings in terms of memory control in VM and how they affects the design here.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
yb01: I have updated the "design consideration" session for proposal justifications.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
vinaykul: Windows support is not there today (CRI API needs update), but it can be added fairly quickly. The upstream version is awaiting Patrick's review and overall Windows needs to add support for UpdateContainerResources - last I heard from patrick, they were looking to migrate to containerd, will take some time, and we decided not to block alpha on that.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
yb01: updated the doc.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
vinaykul: kubelet handles them separately i.e. sends separate requests to CRI for CPU requests, memory limits, and cpu limits. The reason being that a we may have memory increasing and cpu decreasing in the same patch from user, and we need to order the changes so that decreases are applied before increases.

yb01: that will be great if it is designed such way for vertical scaling. with that, i will resolve this conversation. thanks folks.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
yb01: Updated doc.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
yb01: removed from no-goal. windows VM should be part of it as implementation is OS neutral

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
yb01: FYI. in the updated doc. to avoid misinterpretation, i suggested to use:
MaxBootTimeMemoryAllocation -- this refers to the "memory" setting in libvirt,
MaxExpandableMemoryAllocation -- this refers to the max hotplugmemory a user can do after the vm is configured and started
CurrentMemoryALloction -- this is memory the current VM configured to use.

Now, with the different size of memory device support, MaxExapandableMemory is not a concern anymore.

I won't recommend to tweek the MaxBootTimeMemoryALlocation for the reasons i described in the doc. please review the updated doc and let me know if you have concerns.

Now in terms of if we want to expose a handle to use to set up the MaxbootTimeMemoryALlocation --- i would defer this to later discussion in future work #1 when we build our own memory management component.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
```diff
@@ -0,0 +1,646 @@
...
+    spec:
+      resource:
+       limits:
+         cpu:
+         memory:
+       request:
+         cpu:
+         memory:
+     
+Arktos controls the resource the same way as for containers, i.e. schedule based on requested resource level, and evict
```
XiaoningDing, line 348: You mean "controls **VM pod** resources the same way..."?

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
```diff
@@ -0,0 +1,646 @@
...
+      ...
+      <maxMemory slots='16' unit='KiB'>1524288</maxMemory>
+      <memory unit='KiB'>524288</memory>
+      <currentMemory unit='KiB'>524288</currentMemory>
+      ...
+    </domain>
+
+Memory defines the max memory allocation at the boot time, while the currentMemory defines the real memory allocation for
+the guest OS. The delta between the currentMemory and memory is the space for memory ballooning to "reclaim" memory
+between VMs and the hypervisor. The maxMemory defines the max memory allocation for a VM at runtime, which can be achieved
```
XiaoningDing, line 367: Since currentMemory is the real memory allocated to VM, how can ballooning reclaim memory from delta (i.e. the memory amount beyond currentMemory)? My understanding is (0, currentMemory) is the space that ballooning can reclaim.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
```diff
@@ -0,0 +1,646 @@
...
+with memory hotplug.
+
+Please be noted that there are a few limitations in current design and workflow:
+1. Only spec.resource.limit was passed to the runtime from Kubelet via the linuxContainerResources.MemoryLimitInBytes
+2. The linuxContainerResources.MemoryLimitInBytes was used to set the "memory" in the VM domain xml definition. currentMemory
+   is set to the "memory" setting by default.
+3. Ballooning operation, i.e. to adjust the currentMemory, an explict API call "setmem" need to be called, i.e. no auto 
+   adjust for currentMemory based on the application running in the VM
+4. Adjust "memory" can only be done after VM is restarted
```
XiaoningDing, line 376: Isn't it hotplug?  From your previous demo we didn't see restarting

XiaoningDing, line 376: I'm confused by item #4.  Is it about hotplug or balloning? But neither of them require restart?

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
```diff
@@ -0,0 +1,646 @@
...
+Memory defines the max memory allocation at the boot time, while the currentMemory defines the real memory allocation for
+the guest OS. The delta between the currentMemory and memory is the space for memory ballooning to "reclaim" memory
+between VMs and the hypervisor. The maxMemory defines the max memory allocation for a VM at runtime, which can be achieved
+with memory hotplug.
+
+Please be noted that there are a few limitations in current design and workflow:
+1. Only spec.resource.limit was passed to the runtime from Kubelet via the linuxContainerResources.MemoryLimitInBytes
+2. The linuxContainerResources.MemoryLimitInBytes was used to set the "memory" in the VM domain xml definition. currentMemory
+   is set to the "memory" setting by default.
+3. Ballooning operation, i.e. to adjust the currentMemory, an explict API call "setmem" need to be called, i.e. no auto
```
XiaoningDing, line 374: Consider reorganizing item 3 to make it a complete sentence.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
```diff
@@ -0,0 +1,646 @@
...
+As described in the "CRI changes" section, the Arttos-vm-runtime UpdateContainerResource() were implemented to update
+the VM CGroup settings and update the VM domain definition with the new resource configuration.
+
+If Kubelet enables its CPU manager and set the cpus to a particular pod. the cpuset is sent to the runtime via the 
+UpdateContainerResource() and the runtime will honor the cpusets in both VM definition and its CGroup.
+
+Two new APIs, SetVcpus() and AdjustDomainMemory() were added to support VM vertical scaling, which extends the call to 
+Libvirt APIs to update the VM domain xml.
+
+	// Update vcpu for a give domain
```
XiaoningDing, line 560: It's better to make these two APIs consistent in terms of naming and style (the input parameter is delta or the new amount).

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
```diff
@@ -0,0 +1,646 @@
...
+     
+ 
+#### API changes
+As described in the "CRI changes" section, the Arttos-vm-runtime UpdateContainerResource() were implemented to update
+the VM CGroup settings and update the VM domain definition with the new resource configuration.
+
+If Kubelet enables its CPU manager and set the cpus to a particular pod. the cpuset is sent to the runtime via the 
+UpdateContainerResource() and the runtime will honor the cpusets in both VM definition and its CGroup.
+
+Two new APIs, SetVcpus() and AdjustDomainMemory() were added to support VM vertical scaling, which extends the call to
```
XiaoningDing, line 557: Are these two APIs or just runtime internal methods? If they are not part of CRI APIs, it's better to put it under VM runtime internal implementation.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
```diff
@@ -0,0 +1,646 @@
...
+        coordinated with other features such as number of VMs scheduled on the host, level of over-subscription on the host etc.
+      
+     2. Memory fragmentation
+        The nature of the ballooning mechanism is to reclaim unused memory blocks of the guest OS and render them to host 
+        hypervisor. Given time and with repeated memory increase and decrease, it can left the VM memory fragmentation 
+        and affect its performance.
+      
+     3. Ballooning is not NUMA aware
+
+   Memory hotplug/unplug simulates the physical memory device being plugged into the physical machines for a guest OS.
```
XiaoningDing, line 438: This paragraph requires indent. Also it focuses mostly on something else instead of NUMA.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
yb01: yes, i will update the wording. thanks.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
yb01: yes, i should describe it clearly. This paragraph is mainly to describe the upper boundary of ballooning can achieve in term of scale-up the memory of the VM. to scale-down, it is the "least memory" of the guestOS ( say 512 Mi ) up to currentMemory allocation. i will clarify.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
yb01: rephased this line. thanks.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
yb01: it is about ballooning. with hotplug new memory device, it won't require a restart. and the libvirt/hypervisor honors the new memory devices to the VM.  Ballooning cannot scale-up to the size which "memory" config sets to.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
yb01: This paragraph is not about the #3 above. i thought #3 is self-explained, so there is no explanation under it. i reworded the #3 and this paragraph to avoid such impressions.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
yb01: Good point. those are not CRI APIs. i have changed this section to "Arktos-vm-runtime internal API changes" to reflect this. thanks.

docs/design-proposals/node/Unified-Vertical-Scaling-for-Container-and-VM-Pods.md
yb01: They kept differently due to the different of those two. Libvirt has a setvcpus() which includes the adjustment of the vcpus for a VM and this is the direct relay to it. i have a separated branch for CPU scale-up/scale-down work. and if we ended up use other better methods, we can reconsider it.

PR #520, opened 2020-07-24
Support list from resourceVersion for partitioned API servers.

Sindica:


Sindica:
https://github.com/futurewei-cloud/arktos/issues/516

staging/src/k8s.io/code-generator/cmd/client-gen/generators/generator_for_type.go
```diff
@@ -456,6 +460,73 @@ func (c *$.type|privatePlural$) List(opts $.ListOptions|raw$) (result *$.resultT
...
+					return
+				} else {
+					continue
+				}
+			}
+
+			currentResult, _ := results[j]
+			if result.Kind == "" {
+				result.TypeMeta = currentResult.TypeMeta
+				result.ListMeta = currentResult.ListMeta
```
chenqianfzh, line 515: I guess you add the check "if result.kind == """ if for efficiency concern. If so, you need set the result.Kind in the block after that check.

staging/src/k8s.io/client-go/kubernetes/typed/apps/v1/deployment.go
```diff
@@ -102,6 +103,72 @@ func (c *deployments) List(opts metav1.ListOptions) (result *v1.DeploymentList,
...
+				}
+			}
+		}
+
+		for _, item := range itemsMap {
+			result.Items = append(result.Items, *item)
+		}
+		return
+	}
+
```
chenqianfzh, line 171: Explain the use scenario of the following code.

Sindica: Done

PR #527, opened 2020-07-28
remove cr storage getter caching

h-w-chen:
**What type of PR is this?**
/kind bug
/kind refactoring

It removes the caching of object that gets network CR.

The caching of getter object is unnecessary and faulty. 
Unnecessary: crdHandler.GetCustomResourceStorage() method (which is called by the code) already uses in-memory map to provide similar caching;
Faulty: the previous implementation cached one getter which won't work in multi-tenant cases.

NONE


pkg/registry/core/service/storage/rest_test.go
```diff
@@ -2374,7 +2358,25 @@ func TestUpdateNodePorts(t *testing.T) {
 
 func TestServiceRegistryCreateWithNetworkOfExternalIPAM(t *testing.T) {
 	storage, registry, server := NewTestREST(t, nil)
-	storage.networks = &dummyNetworkGetter{ipamType: "External"}
+	storage.fnGetNetwork = func(ctx context.Context, tenant, name string) (object runtime.Object, err error) {
+		return &unstructured.Unstructured{
+			Object: map[string]interface{}{
+				"apiVersion": "arktos.futurewei.com/v1",
```
chenqianfzh, line 2364: SO many hard-coded values. It is better to define some vars and consts.

h-w-chen: good point. will use consts..

PR #528, opened 2020-07-29
Configure QPS (20) for workload controller manager

Sindica:


cmd/workload-controller-manager/app/config/controllerconfig.go
```diff
@@ -19,11 +19,13 @@ package config
 import (
 	"encoding/json"
 	"io/ioutil"
+	"k8s.io/klog"
 	"os"
 )
 
 type controllerManagerConfig struct {
 	ReportHealthIntervalInSecond int              `json:"reportHealthIntervalInSecond"`
+	QPS                          float32          `json:"qps""`
```
jshaofuturewei, line 28: Is it kube-api-qps in kube-controller-manager? There are multiple qps settings in controller-manager, such as 
bind-pods-qps.

Sindica: Correct, kube controller manager allows different qps setting for different controller. 20 is a generic setting. All rest client if not specified, use the generic qps. We can add controller specific qps later.

PR #541, opened 2020-08-04
Perf-test changes

Sindica:
. Allow perf-test to get all api servers in partitioned api server environment
   . Support list from revision in dynamic client
. Sonya: dynamic to create namespaces and force to average namespace name per apiserver number

Scheduling throughput on 2 partitioned api servers:
{
  "perc50": 20,
  "perc90": 20,
  "perc99": 20.2,
  "max": 23.4
}

for Arktos changes:
https://github.com/Sindica/arktos/commits/2apiserver-sonya-verified

image

### DIFF
--- a/perf-tests/clusterloader2/pkg/framework/config/client_config.go
+++ b/perf-tests/clusterloader2/pkg/framework/config/client_config.go
@@ -51,11 +51,11 @@ func SetAPIServers(apiServerAddresses string) error {
 
 	addrs := strings.Split(apiServerAddresses, ";")
 	for _, item := range addrs {
-		extraAPIServerAddresses = append(extraAPIServerAddresses, item)
+		extraAPIServerAddresses = append(extraAPIServerAddresses, strings.TrimSpace(item))
 	}
 
 	if len(extraAPIServerAddresses) == 0 {
-		klog.Infof("Passed in api server addresses [%s]. Keep kubeconfig value instead.", apiServerAddresses)
+		klog.Infof("Passed in extra api server addresses [%s]. Keep kubeconfig value.", apiServerAddresses)
 	}
 	return nil
 }
--- a/perf-tests/clusterloader2/pkg/framework/config/client_config.go
+++ b/perf-tests/clusterloader2/pkg/framework/config/client_config.go
@@ -18,8 +18,11 @@ package config
 
 import (
 	"fmt"
+	"k8s.io/klog"
 	"net"
 	"net/http"
+	"strings"
+	"sync"
 	"time"
 
 	utilnet "k8s.io/apimachinery/pkg/util/net"
@@ -35,8 +38,33 @@ const (
 	burst       = 200
 )
 
+var muxInitializedMultiClientConfig sync.Mutex
+
+var extraAPIServerAddresses []string
+
+func SetAPIServers(apiServerAddresses string) error {
+	muxInitializedMultiClientConfig.Lock()
+	defer muxInitializedMultiClientConfig.Unlock()
+	if len(extraAPIServerAddresses) > 0 {
+		return fmt.Errorf("Api Server addresses already set [%v]", extraAPIServerAddresses)
+	}
+
+	addrs := strings.Split(apiServerAddresses, ";")
+	for _, item := range addrs {
+		extraAPIServerAddresses = append(extraAPIServerAddresses, item)
+	}
+
+	if len(extraAPIServerAddresses) == 0 {
+		klog.Infof("Passed in api server addresses [%s]. Keep kubeconfig value instead.", apiServerAddresses)
+	}
+	return nil
+}
+
 // PrepareConfig creates and initializes client config.
 func PrepareConfig(path string) (*restclient.Config, error) {
+	muxInitializedMultiClientConfig.Lock()
+	defer muxInitializedMultiClientConfig.Unlock()
+
 	config, err := loadConfig(path)
 	if err != nil {
 		return nil, err
@@ -44,7 +72,28 @@ func PrepareConfig(path string) (*restclient.Config, error) {
 	if err = initializeWithDefaults(config); err != nil {
 		return nil, fmt.Errorf("config initialization error: %v", err)
 	}
-	return config, nil
+
+	if len(extraAPIServerAddresses) == 0 {
+		return config, nil
+	}
+
+	// create config for additional api server partitioning
+	aggConfig := restclient.NewAggregatedConfig(config.GetConfig())
+	for _, apiServerAddr := range extraAPIServerAddresses {
+		newConfig, _ := loadConfig(path)
+		initializeWithDefaults(newConfig)
+		c := newConfig.GetConfig()
+
+		// format: https://34.1.1.1
+		c.Host = apiServerAddr
+		aggConfig.AddConfig(c)
+	}
+
+	for i, config := range aggConfig.GetAllConfigs() {
+		klog.Infof("Kubeconfig %d host %v", i, config.Host)
+	}
+
+	return aggConfig, nil
 }
 
 func restclientConfig(path string) (*clientcmdapi.Config, error) {
--- a/perf-tests/clusterloader2/pkg/framework/framework.go
+++ b/perf-tests/clusterloader2/pkg/framework/framework.go
@@ -29,6 +29,7 @@ import (
 	"k8s.io/kubernetes/perf-tests/clusterloader2/pkg/config"
 	"k8s.io/kubernetes/perf-tests/clusterloader2/pkg/errors"
 	"k8s.io/kubernetes/perf-tests/clusterloader2/pkg/framework/client"
+	"k8s.io/kubernetes/perf-tests/clusterloader2/pkg/util"
 
 	// ensure auth plugins are loaded
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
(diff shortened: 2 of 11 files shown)


staging/src/k8s.io/client-go/dynamic/simple.go
```diff
@@ -303,27 +306,97 @@ func (c *dynamicResourceClient) Get(name string, opts metav1.GetOptions, subreso
...
+	rv := ""
+	for i := 0; i < len(searchClients); i++ {
+		if errs[i] != nil {
+			return nil, errs[i]
+		}
 
-	list, err := uncastObj.(*unstructured.Unstructured).ToList()
-	if err != nil {
-		return nil, err
+		if rv == "" {
```
XiaoningDing, line 372: This if-else check seems unnecessary.  We can simply loop through searchClients.  If rv is empty,  then searchClients will have only one element and we will finish the loop naturally.

Sindica: It is mainly used for initialize. rv is initialized as empty string and will pick up first ResourceVersion

staging/src/k8s.io/client-go/dynamic/simple.go
```diff
@@ -303,27 +306,97 @@ func (c *dynamicResourceClient) Get(name string, opts metav1.GetOptions, subreso
...
+			if err != nil {
+				setResult(resultMap, errMap, lock, pos, nil, err)
+				return
+			} else {
+				setResult(resultMap, errMap, lock, pos, list, nil)
+			}
+
+		}(client, opts, &listLock, i, results, errs)
+	}
+	wg.Wait()
```
XiaoningDing, line 362: Nit:  A shared channel + select is probably is better here, as we can start merging results without waiting for everyone to finish.  But I guess the cost will mostly be networking and server side processing, so the actual difference should be very minor here.

perf-tests/clusterloader2/pkg/framework/config/client_config.go
```diff
@@ -35,16 +38,62 @@ const (
...
+var extraAPIServerAddresses []string
+
+func SetAPIServers(apiServerAddresses string) error {
+	muxInitializedMultiClientConfig.Lock()
+	defer muxInitializedMultiClientConfig.Unlock()
+	if len(extraAPIServerAddresses) > 0 {
+		return fmt.Errorf("Api Server addresses already set [%v]", extraAPIServerAddresses)
+	}
+
+	addrs := strings.Split(apiServerAddresses, ";")
```
chenqianfzh, line 52: need to check to each string is a valid string. 

This func may be helpful:
https://golang.org/pkg/net/#ParseIP

perf-tests/clusterloader2/pkg/framework/config/client_config.go
```diff
@@ -35,16 +38,62 @@ const (
...
+	for _, item := range addrs {
+		extraAPIServerAddresses = append(extraAPIServerAddresses, item)
```
chenqianfzh, line 54: Why adding them one by one? extraAPIServerAddresses is exactly the same as addrs.

perf-tests/clusterloader2/pkg/framework/config/client_config.go
```diff
@@ -35,16 +38,62 @@ const (
...
+	}
+
+	if len(extraAPIServerAddresses) == 0 {
+		klog.Infof("Passed in api server addresses [%s]. Keep kubeconfig value instead.", apiServerAddresses)
```
chenqianfzh, line 58: Rephrase this log please. The info is not clear to the reader. It would better to say "No api server addresses found in the passed-in string [%s] ...."

perf-tests/clusterloader2/pkg/framework/config/client_config.go
```diff
@@ -35,16 +38,62 @@ const (
...
+		initializeWithDefaults(newConfig)
+		c := newConfig.GetConfig()
+
+		// format: https://34.1.1.1
+		c.Host = apiServerAddr
+		aggConfig.AddConfig(c)
+	}
+
+	for i, config := range aggConfig.GetAllConfigs() {
+		klog.Infof("Kubeconfig %d host %v", i, config.Host)
```
chenqianfzh, line 93: The log message will be hard to understand. A reader probably would be puzzled by the following log message:

"kubeconfig 0 host https://34.1.1.1"

staging/src/k8s.io/client-go/dynamic/simple.go
```diff
@@ -21,9 +21,12 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"k8s.io/apimachinery/pkg/util/diff"
```
chenqianfzh, line 24: Nit: put the golang native packages in one group, and k8s packages as another group.

perf-tests/clusterloader2/pkg/util/util.go
```diff
@@ -245,11 +245,20 @@ func CloneMap(src map[string]interface{}) map[string]interface{} {
 }
 
 // RandomDNS1123String generates random string of a given length.
-func RandomDNS1123String(length int) string {
+func RandomDNS1123String(length int, startpos int, endpos int) string {
 	characters := []rune("abcdefghijklmnopqrstuvwxyz0123456789")
+	characters_pos := characters
+	if startpos < endpos {
```
chenqianfzh, line 251: Here you are using condition "startpos < endpos" to check whether there are extra api servers. 
if startpos==endpos , there is no extra api server, otherwise there is.

It is not a good design for a util function. The behavior of this util is weird and hard to understand after this change.

perf-tests/clusterloader2/pkg/framework/framework.go
```diff
@@ -106,13 +107,42 @@ func (f *Framework) CreateAutomanagedNamespaces(namespaceCount int) error {
...
+	} else {
+		namespaceinterval := 0
+		apiservernum := f.clusterConfig.Apiserverextranum + 1
+		if namespaceCount%apiservernum > 0 {
+			namespaceinterval = namespaceCount/apiservernum + 1
+		} else {
+			namespaceinterval = namespaceCount / apiservernum
+		}
+		for server := 1; server <= apiservernum; server++ {
+			endpos = startpos + (26 / apiservernum)
```
chenqianfzh, line 129: What happen if the number of apiservers is really large? If it is really large, endpos == startpos, the following logic will not work as hoped?

Are you going to set a limit on the apiserver num?

perf-tests/clusterloader2/pkg/framework/framework.go
```diff
@@ -248,7 +298,7 @@ func (f *Framework) ApplyTemplatedManifests(manifestGlob string, templateMapping
 }
 
 func (f *Framework) isAutomanagedNamespace(name string) (bool, error) {
-	return regexp.MatchString(f.automanagedNamespacePrefix+"-[1-9][0-9]*", name)
+	return regexp.MatchString("[a-zA-Z0-9]{6,}-"+f.automanagedNamespacePrefix, name)
```
chenqianfzh, line 301: Kubernetes does not allow uppercase lettters in the name of namespaces.

perf-tests/clusterloader2/pkg/framework/framework.go
```diff
@@ -158,14 +188,34 @@ func (f *Framework) deleteNamespace(namespace string) error {
...
-				errList.Append(err)
-				return
-			}
-		})
+	automanagedNamespacesList, staleNamespaces, err := f.ListAutomanagedNamespaces()
+	if err != nil {
+		errList.Append(err)
+		return errList
+	}
+	if len(automanagedNamespacesList) > 0 {
```
chenqianfzh, line 196: Nit: it seems to me the code from line 196 to 218 can be simplified as :

namespacesToDelete := append(automanagedNamespacesList, staleNamespaces)
for _, nsName := namespacesToDelete {
       wg.Start(func() {
				if err := f.deleteNamespace(nsName); err != nil {
					errList.Append(err)
					return
				}
			})
}

staging/src/k8s.io/client-go/dynamic/simple.go
Sindica: The time saved for starting merge earlier is not significant. Would like to use more straight forward code to prevent potential bugs.

perf-tests/clusterloader2/pkg/framework/config/client_config.go
Sindica: The arg will have format as server from kubeconfig. It can be hostname or ip or dns, with or without port number. It will be too complicated to validate each scenario. Plus, the perf-test process will immediately exit if config is not correct and not be able to create a valid connection to api server. I don't think it is worthwhile to deal with argument validation here.

perf-tests/clusterloader2/pkg/framework/config/client_config.go
Sindica: Previously tried to validate each addresses. Will update.

perf-tests/clusterloader2/pkg/framework/config/client_config.go
Sindica: I rephrased the word a bit.

perf-tests/clusterloader2/pkg/framework/config/client_config.go
Sindica: This is used by perf test only. Not for end user.

staging/src/k8s.io/client-go/dynamic/simple.go
Sindica: hmm. I wonder why go fmt did not rearrange it.

PR #558, opened 2020-08-11
Bug fix for clientset generation

Sindica:
This fix kube controller manager leader election issue: https://github.com/futurewei-cloud/arktos/issues/538

Test done:
1. Local: print out rate limiter pointer for rest request
2. Run perf test 500 node with KCM leader election enabled.


cmd/workload-controller-manager/main.go
```diff
@@ -92,7 +92,7 @@ func getConfig() (*controllerManagerConfig.Config, error) {
 	}
 
 	klog.Infof("Current workload controller port %d", workloadControllerPort)
-	klog.Infof("Rest client QPS %v, kubeconfig [%#v]", controllerManagerKubeConfig.GetConfig().QPS, controllerManagerKubeConfig.GetConfig())
```
chenqianfzh, line 95: Nit: I see this bug is about QPS. But for the purpose of information, we need the controllerManagerKubeConfig.GetConfig(), not just QPS.

staging/src/k8s.io/client-go/kubernetes/clientset.go
```diff
@@ -333,160 +333,163 @@ func (c *Clientset) Discovery() discovery.DiscoveryInterface {
...
+		configCopy := *currentConfig
 		if configCopy.RateLimiter == nil && configCopy.QPS > 0 {
 			configCopy.RateLimiter = flowcontrol.NewTokenBucketRateLimiter(configCopy.QPS, configCopy.Burst)
 		}
+		configShallowCopy.AddConfig(&configCopy)
 	}
 	var cs Clientset
 	var err error
-	cs.admissionregistrationV1beta1, err = admissionregistrationv1beta1.NewForConfig(&configShallowCopy)
+	cs.admissionregistrationV1beta1, err = admissionregistrationv1beta1.NewForConfig(configShallowCopy)
```
chenqianfzh, line 347: I am puzzled by this change. I see the signature admissionregistrationv1beta1.NewForConfig(c *rest.Config) in the code. Could you tell me how come this change works?

cmd/workload-controller-manager/main.go
Sindica: GetConfig has a lot of information. including cert data. It is best to print out non sensitive and useful information.

staging/src/k8s.io/client-go/kubernetes/clientset.go
Sindica: configShallowCopy has type *rest.Config

PR #576, opened 2020-08-12
Global Scheduler - Types, scheduler.go, bazel test, bazel test integration

CoderKevinZhang:
**What type of PR is this?**
> /kind design

> This PR contains the global scheduler necessary components

> Some integration tests are still needed to be discussed

```release-note
NONE
```


cathyhongzhang:
Just a couple of comments w.r.t. the new UT: 
The following UTs seem missing:
1.	Scheduler creation failure 
2.	Multiple requests with invalid params
3.	Single creation request with valid host, valid token, invalid POD spec object
4.	Multiple delete instance requests with valid params
5.	Multiple delete instance requests with invalid params

Not sure if the following negative test cases are covered. If not, could you add these tests?  
1.	Timeout/failure of sending the scheduling decision to the selected cluster
2.	Successfully sends out the scheduling decision, but the selected Cluster returns scheduling failure 
3.	Successfully sends out the scheduling decision, but time out before getting cluster scheduling result.

PR #577, opened 2020-08-12
add apiserver data partition to kube-up and start-kubemark

sonyafenge:
**What type of PR is this?**

> /kind feature

add apiserver data partition to kube-up and start-kubemark for multi apiservers

Fixes #

```release-note

```


cluster/gce/config-test.sh
```diff
@@ -42,19 +42,10 @@ NODE_LOCAL_SSDS=${NODE_LOCAL_SSDS:-0}
 NODE_LABELS="${KUBE_NODE_LABELS:-}"
 WINDOWS_NODE_LABELS="${WINDOWS_NODE_LABELS:-}"
 APISERVERS_EXTRA_NUM=${APISERVERS_EXTRA_NUM:-0}
+APISERVER_DATAPARTITION_CONFIG=${APISERVER_DATAPARTITION_CONFIG:-"abcdefghijklmnopqrstuvwxyz"}
```
Sindica, line 45: Can there be a common var file?

cluster/gce/util.sh
```diff
@@ -2895,6 +2895,59 @@ function config-partitionserver() {
 
 }
 
+
+##set apiserver datapartition config by apiserver total number
```
Sindica, line 2899: I would prefer those logic to go to perf test. Environment set up should only have necessary information and do not care how the environment is used.

Sindica: Talk to Xiaoning if you think it is necessary to do it here.

sonyafenge: as my understanding, if we start 3 apiservers, we should have data partition ready for these 3 apiservers, doesn't mentor if we run perf-tests. data partition is part of multi apiserver, not part of perf-tests.

cluster/gce/config-test.sh
sonyafenge: cluster/gce/config-test.sh and cluster/gce/config-default.sh are the common var files for kube-up and start-kubemark. it is same as /hack/lib/common-var-init.sh.

Sindica: Here cluster/gce/config-default.sh and cluster/gce/config-test.sh share many common variables and have the same value.

vinaykul: IIRC they are used in different places. config-default is used in kube-up (admin cluster). config-test is used in start-kubemark (kubemark cluster). they can have different configurations - e.g. 6443 is the api port for admin cluster, 443 is default api port for kubemark iirc.

cluster/gce/util.sh
vinaykul: I agree with Sonya. Configuration of the cluster is a setup thing, and not a test specific thing.

Sindica: Setup can also be multiple parts. API server partition is designed to be flexible, not tied to environments.

cluster/gce/config-test.sh
Sindica: Different places can also share same functional module. I do see many times the two config files had to be changed together for same variables and values. I don't know why they can not be merged.

cluster/kube-up.sh
```diff
@@ -78,7 +78,24 @@ if [[ "${ENABLE_PROXY:-}" == "true" ]]; then
   echo ""
 fi
 
-echo -e "Done, listing cluster services:\n" >&2
+if [[ "${APISERVERS_EXTRA_NUM:-0}" -gt "0" ]]; then
```
XiaoningDing, line 81: Is API Server partitioning optional for this setup script?  Looks like it's mandatory here when we have multiple API Servers. I think we should make it optional even there are multiple API Servers.

sonyafenge: it is optional, only when APISERVERS_EXTRA_NUM >0, that means we have multi apiservers.

sonyafenge:
I will merge this PR for now to enable automatically multi apiserver data partition config. And will continue working to create separated .sh files logic which can be used seperately.

PR #580, opened 2020-08-12
Fix a nil pointer crash caused by a race condition in watch cache

### LINKED ISSUE #555: panic of invalid memory address or nil pointer dereference in API Server cache
**What happened**:

An invalid pointer reference happened during scalability test.  This caused API server process to panic.

There should not be invalid pointer dereference.

This was found by analyzing the API Server log of 10K node scalability tests.  The stack track in the log was put in this issue descritpion.

I0807 02:55:46.292995       1 trace.go:81] Trace[869381899]: "cacher *core.Node: List" (started: 2020-08-07 02:55:45.171326591 +0000 UTC m=+8.054876990) (total time: 1.121590705s):
Trace[869381899]: [1.068µs] [1.068µs] Ready
Trace[869381899]: [1.121577605s] [1.121576537s] watchCache locked acquired
Trace[869381899]: [1.12157789s] [285ns] watchCache fresh enough
Trace[869381899]: [1.121581125s] [3.235µs] Got from cache
Trace[869381899]: [1.121590705s] [9.58µs] END
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        **panic: runtime error: invalid memory address or nil pointer dereference**
[signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0x14f3127]
 
goroutine 77979 [running]:
k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
        /go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:58 +0x105
panic(0x3a300e0, 0xa37d2e0)
        /usr/local/go/src/runtime/panic.go:522 +0x1b5
k8s.io/kubernetes/vendor/k8s.io/apiserver/pkg/storage/cacher.(*cacheWatcher).convertToWatchEvent(0xc033566380, 0x0, 0x0)
        /go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/vendor/k8s.io/apiserver/pkg/storage/cacher/cacher.go:1144 +0x37
k8s.io/kubernetes/vendor/k8s.io/apiserver/pkg/storage/cacher.(*cacheWatcher).sendWatchCacheEvent(0xc033566380, 0x0)
        /go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/vendor/k8s.io/apiserver/pkg/storage/cacher/cacher.go:1177 +0x4d
k8s.io/kubernetes/vendor/k8s.io/apiserver/pkg/storage/cacher.(*cacheWatcher).process(0xc033566380, 0x76ff620, 0xc0206e6d80, 0xc01cbec000, 0x3cf, 0x3cf, 0xb9e36c420680001)
        /go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/vendor/k8s.io/apiserver/pkg/storage/cacher/cacher.go:1226 +0xb3
created by k8s.io/kubernetes/vendor/k8s.io/apiserver/pkg/storage/cacher.(*Cacher).Watch
        /go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/vendor/k8s.io/apiserver/pkg/storage/cacher/cacher.go:493 +0x736

XiaoningDing:
**What type of PR is this?**

> /kind bug

In 10K-node scalability test, API Server crashed due to a nil pointer dereference.  It was because a nil item was returned from watch cache.  This PR fixes the issue.

In the watch cache, revCache acts as a pointer to the actual items in Cache. When we delete one item, we should first delete (or overwrite) the pointer. If we delete the object before overwriting the pointer, another thread could read a pointer but couldn’t find the associated item in the map Cache, and therefore get a nil item.

Fixes #555

No

No

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/storage/cacher/watch_cache.go
+++ b/staging/src/k8s.io/apiserver/pkg/storage/cacher/watch_cache.go
@@ -147,10 +147,13 @@ func newWatchCache(
 	getAttrsFunc func(runtime.Object) (labels.Set, fields.Set, error),
 	versioner storage.Versioner) *watchCache {
 	wc := &watchCache{
-		capacity:            capacity,
-		keyFunc:             keyFunc,
-		getAttrsFunc:        getAttrsFunc,
-		cache:               make(map[uint64]*watchCacheEvent, capacity),
+		capacity:     capacity,
+		keyFunc:      keyFunc,
+		getAttrsFunc: getAttrsFunc,
+
+		// One extra slot is needed because we delete old item after
+		// inserting new item.
+		cache:               make(map[uint64]*watchCacheEvent, capacity+1),
 		revCache:            make([]uint64, capacity),
 		startIndex:          0,
 		endIndex:            0,
@@ -274,15 +277,22 @@ func (w *watchCache) processEvent(event watch.Event, resourceVersion uint64, upd
 
 // Assumes that lock is already held for write.
 func (w *watchCache) updateCache(event *watchCacheEvent) {
+	var toRemove bool
+	var revToRemove uint64
+
 	if w.endIndex == w.startIndex+w.capacity {
 		// Cache is full - remove the oldest element.
-		revToRemove := w.revCache[w.startIndex%w.capacity]
-		delete(w.cache, revToRemove)
+		revToRemove = w.revCache[w.startIndex%w.capacity]
+		toRemove = true
 		w.startIndex++
 	}
 	w.cache[event.ResourceVersion] = event
 	w.revCache[w.endIndex%w.capacity] = event.ResourceVersion
 	w.endIndex++
+
+	if toRemove {
+		delete(w.cache, revToRemove)
+	}
 }
 
 // List returns list of pointers to <storeElement> objects.
@@ -487,7 +497,7 @@ func (w *watchCache) GetAllEventsSinceThreadUnsafe(resourceVersion uint64) ([]*w
 		rev := w.revCache[(w.startIndex+first+i)%w.capacity]
 		result, isOK := w.cache[rev]
 		if !isOK {
-			klog.Errorf("Expected event not found in cache. Rev %v", rev)
+			return nil, fmt.Errorf("Expected event not found in cache. Rev %v", rev)
 		} else {
 			results[i] = result
 		}


staging/src/k8s.io/apiserver/pkg/storage/cacher/watch_cache.go
```diff
@@ -274,15 +274,22 @@ func (w *watchCache) processEvent(event watch.Event, resourceVersion uint64, upd
 
 // Assumes that lock is already held for write.
 func (w *watchCache) updateCache(event *watchCacheEvent) {
+	var toRemove bool
```
Sindica, line 277: Prefer rename as isRemove and initialize as false.

staging/src/k8s.io/apiserver/pkg/storage/cacher/watch_cache.go
```diff
@@ -487,7 +494,7 @@ func (w *watchCache) GetAllEventsSinceThreadUnsafe(resourceVersion uint64) ([]*w
 		rev := w.revCache[(w.startIndex+first+i)%w.capacity]
 		result, isOK := w.cache[rev]
 		if !isOK {
-			klog.Errorf("Expected event not found in cache. Rev %v", rev)
+			return nil,fmt.Errorf("Expected event not found in cache. Rev %v", rev)
```
Sindica, line 497: I think go fmt would complain

staging/src/k8s.io/apiserver/pkg/storage/cacher/watch_cache.go
XiaoningDing: Golang will guarantee it's initialized to false, which is the zero value for a bool type.

staging/src/k8s.io/apiserver/pkg/storage/cacher/watch_cache.go
XiaoningDing: I already ran "gofmt" on this changed file.  This is the file after formatting.

staging/src/k8s.io/apiserver/pkg/storage/cacher/watch_cache.go
Sindica: I agree golang would do that but I personally do not rely on compiler since it does not save any performance.

XiaoningDing: It's not about compiler implementation. It's defined by the go language specification.  Initializing to false here actually is a redundancy, which I don't like :)

Sindica: I found it is clear :)

XiaoningDing: :) I will resolve this since it's personal preference.

staging/src/k8s.io/apiserver/pkg/storage/cacher/watch_cache.go
XiaoningDing: You are right. I ran gofmt but didn't add "-w" parameter so the change wasn't applied.  Will fix.  Thanks for catching this.

PR #582, opened 2020-08-13
perf-tests/clusterloader2/run-e2e-arktos.sh: run perf-test e2e script without the Kubernetes softlink workaround

chenqianfzh:
**Issue** : perf-tests/clusterloader2/run-e2e.sh is the entry script to run perf-tests in kubernetes. However, our project is named "arktos" yet the code has many dependencies in "k8s.io/kubernetes", running this script in arktos leads to build failures. 

**Cause**: The failure is due to that the script uses command "go build" to build the binary clusterloader, which follows a specific order to find the dependency packages. I tried different ways but every one runs into some issues.

**Workaround**: Perf test team found a workaround, which creates a kubernetes soft link in parallel to the "arktos" folder and run the script from "kubernetes" folder. This workaround is surely not a clean solution. 

**Solution**: building using "make" does not suffer this issue. So I create a new script, which does the same thing as the run-e2e.sh except the binary building is changed to use "make". I did not remove the old script, though.

**Verification**:
In my gce machine:
image

### DIFF
--- a/pkg/cloudfabric-controller/controllerframework/cim_test.go
+++ b/pkg/cloudfabric-controller/controllerframework/cim_test.go
@@ -23,7 +23,7 @@ import (
 	"testing"
 )
 
-func newControllerInstance(controllerType string, controllerKey int64, workloadNum int32, isLocked bool) *v1.ControllerInstance {
+func newControllerInstance(cim *ControllerInstanceManager, controllerType string, controllerKey int64, workloadNum int32) *v1.ControllerInstance {
 	controllerInstance := &v1.ControllerInstance{
 		ObjectMeta: metav1.ObjectMeta{
 			ResourceVersion: "100",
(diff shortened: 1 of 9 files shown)


cathyhongzhang:
Could you also update https://github.com/futurewei-cloud/arktos/tree/master/docs/setup-guide to add this new script so that people know which is the correct script to use for performance test? Thanks.

chenqianfzh:
> Could you also update https://github.com/futurewei-cloud/arktos/tree/master/docs/setup-guide to add this new script so that people know which is the correct script to use for performance test? Thanks.
> 
>

 Qian: Sure, will do it in a separate PR after the reviewers agree on this change.

sonyafenge:
From code, it looks good. I will take some time to test this change and then update this thread.

Sindica:
> > Could you also update https://github.com/futurewei-cloud/arktos/tree/master/docs/setup-guide to add this new script so that people know which is the correct script to use for performance test? Thanks.
> 
> Qian: Sure, will do it in a separate PR after the reviewers agree on this change.

Sonya has a pending PR for how to run perf test:
https://github.com/futurewei-cloud/arktos/pull/545
Please update the same document instead of creating a new one. Thank you!

PR #587, opened 2020-08-13
E2e perf test

chenqianfzh:
Enable prometheus e2e-tests after perf-tests are integrated into arktos.

Verification:
run arktos/perf-tests/clusterloader/run-e2e.sh with env var ENABLE_PROMETHEUS_SERVER=true in gce machine and confirmed:
1. the script finished successfully
2. The prometheus perf results are reported.
3. The test results are approximately same as the tests done with the perf-tests before migration.

### DIFF
--- a/perf-tests/clusterloader2/pkg/imagepreload/imagepreload.go
+++ b/perf-tests/clusterloader2/pkg/imagepreload/imagepreload.go
@@ -36,7 +36,7 @@ import (
 
 const (
 	informerTimeout = time.Minute
-	manifest        = "$GOPATH/src/k8s.io/perf-tests/clusterloader2/pkg/imagepreload/manifests/daemonset.yaml"
+	manifest        = "$GOPATH/src/k8s.io/arktos/perf-tests/clusterloader2/pkg/imagepreload/manifests/daemonset.yaml"
 	namespace       = "preload"
 	daemonsetName   = "preload"
 	pollingInterval = 5 * time.Second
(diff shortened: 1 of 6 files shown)


perf-tests/clusterloader2/pkg/imagepreload/imagepreload.go
```diff
@@ -36,7 +36,7 @@ import (
 
 const (
 	informerTimeout = time.Minute
-	manifest        = "$GOPATH/src/k8s.io/perf-tests/clusterloader2/pkg/imagepreload/manifests/daemonset.yaml"
+	manifest        = "$GOPATH/src/k8s.io/arktos/perf-tests/clusterloader2/pkg/imagepreload/manifests/daemonset.yaml"
```
Sindica, line 39: Can this be changed to "$GOPATH/src/arktos/perf-tests/... ?

Sindica: This change needs to work with PR 583 to get Promethus data, correct? Can they be merged?

chenqianfzh: For the first comment, "$GOPATH/src/k8s.io/arktos/" is the correct path to the arktos repo in perf-test env. You should run into some problems if you put it as $GOPATH/src/arktos/.

For the second comment, these two changes are separate. This one deals with the perf-test tool, 583 is about kubelet prometheus registration. They might work together to produce the perf results, but lets's not mix them in one PR.

PR #594, opened 2020-08-16
Update logging for more info to investigate stress related issues

yb01:
**What type of PR is this?**
Update log for stress related issues

Need more info from log to investigate stress test issues

### DIFF
--- a/pkg/kubelet/kubelet.go
+++ b/pkg/kubelet/kubelet.go
@@ -1999,16 +1999,16 @@ func (kl *Kubelet) syncLoopIteration(configCh <-chan kubetypes.PodUpdate, handle
 			// once we have checkpointing.
 			handler.HandlePodAdditions(u.Pods)
 		case kubetypes.UPDATE:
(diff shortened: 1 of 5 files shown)


pkg/kubelet/kubelet.go
```diff
@@ -1999,16 +1999,16 @@ func (kl *Kubelet) syncLoopIteration(configCh <-chan kubetypes.PodUpdate, handle
 			// once we have checkpointing.
 			handler.HandlePodAdditions(u.Pods)
 		case kubetypes.UPDATE:
-			klog.V(2).Infof("SyncLoop (UPDATE, %q): %q", u.Source, format.PodsWithDeletionTimestamps(u.Pods))
+			klog.V(2).Infof("SyncLoop (UPDATE, %q): %q", u.Source, format.PodsWithDeletionTimestampsAndResourceVersions(u.Pods))
```
XiaoningDing, line 2002: You can print just one log in the beginning of the loop, instead of changing every case.

PR #596, opened 2020-08-17
Increase UT test coverage for api server/etcd data partition - part 1

Sindica:
Issue https://github.com/futurewei-cloud/arktos/issues/568

### DIFF
--- a/pkg/master/controller_test.go
+++ b/pkg/master/controller_test.go
@@ -1391,7 +1391,10 @@ func TestMasterCountRemoveEndpoints(t *testing.T) {
 				if !reflect.DeepEqual(err, test.expectedErr) {
 					t.Errorf("case %q: expected error: %v; got error: %v", test.testName, test.expectedErr, err)
 				}
+			} else if err != nil {
+				t.Errorf("case %q: unexpected err: %v", test.testName, err)
 			}
+
 			if test.expectUpdate != nil {
 				if e, a := test.expectUpdate, actualEndpoints; !reflect.DeepEqual(e, a) {
 					t.Errorf("case %q: expected update:\n%#v\ngot:\n%#v\n", test.testName, e, a)
--- a/pkg/master/controller_test.go
+++ b/pkg/master/controller_test.go
@@ -18,6 +18,7 @@ limitations under the License.
 package master
 
 import (
+	"k8s.io/apimachinery/pkg/api/errors"
 	"net"
 	"reflect"
 	"testing"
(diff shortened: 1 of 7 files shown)


pkg/registry/core/controllerinstance/storage/storage_test.go
```diff
@@ -118,3 +119,40 @@ func TestList(t *testing.T) {
...
+	defer server.Terminate(t)
+	defer storage.DestroyFunc()
+	controllerInstance := validNewControllerInstance()
+	ctx := genericapirequest.NewContext()
+	_, err := storage.Store.Create(ctx, controllerInstance, rest.ValidateAllObjectFunc, &metav1.CreateOptions{})
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+
+	object, err := storage.Get(ctx, "foo", &metav1.GetOptions{})
```
chenqianfzh, line 134: Let's define "foo" as a const var outside any funcs.

pkg/registry/core/controllerinstance/storage/storage_test.go
```diff
@@ -118,3 +119,40 @@ func TestList(t *testing.T) {
...
+	}
+
+	// read back and compare
+	newObj, err := storage.Get(ctx, "foo", &metav1.GetOptions{})
+	if err != nil {
+		t.Errorf("unexpected error: %v", err)
+	}
+	newInstance := newObj.(*api.ControllerInstance)
+	if newInstance.ControllerKey != math.MaxInt64 {
+		t.Fatalf("controller key update failed")
```
chenqianfzh, line 155: Log the real newInstance.ControllerKey in the error message.

pkg/master/controller_test.go
```diff
@@ -1290,3 +1291,112 @@ func TestCreateOrUpdateMasterService(t *testing.T) {
...
+				}
+			}
+
+			r := reconcilers.NewMasterCountEndpointReconciler(3, clientset.CoreV1())
+			err := r.RemoveEndpoints(test.serviceName, "", net.ParseIP(test.ip), test.endpointPorts)
+			if err != nil {
+				t.Errorf("case %q: unexpected error: %v", test.testName, err)
+			}
+			actualEndpoints, err := clientset.CoreV1().Endpoints(corev1.NamespaceDefault).Get(test.serviceName, metav1.GetOptions{})
+			if test.expectedErr != nil {
```
chenqianfzh, line 1390: Need to add a test for case where test.expected == nil but err != nil

pkg/registry/core/controllerinstance/storage/storage_test.go
Sindica: Should the previous value which is 1. But I updated it anyway.

PR #600, opened 2020-08-18
kubelet should not skip pod synchronization when network is not ready

### LINKED ISSUE #595: kubelet should run scheduled cni pod when node is not ready because cni has not been installed
**What happened**:
kubelet could not start flannel cni pod scheduled to the node

kubelet starts the flannel pod, and eventually bring up the node to ready state

1. ensures to use the code that scheduler is able to schedule cni pod (see PR #593 )
2. prepare arktos-up.sh to comment out bridge cni installation line ```#source ${KUBE_ROOT}/hack/arktos-cni.rc```
3. start arktos lab env by ```./hack/arktos-up.sh```, wait till it displays ```Waiting for node ready at api server```
4. on another console, run ```./cluster/kubectl.sh apply -f https://github.com/coreos/flannel/raw/master/Documentation/kube-flannel.yml```
5. now, the pod status is still Pending (and scheduled), its detailed status is
```yaml
...
status:
  conditions:
  - lastProbeTime: null
    lastTransitionTime: "2020-08-17T16:11:53Z"
    status: "True"
    type: PodScheduled
  phase: Pending
  qosClass: Guaranteed
```

looking at /tmp/kubelet.log, it seems kubelet skips pod process loop:
```text
I0817 16:30:47.791828   27669 kubelet.go:1933] skipping pod synchronization - container runtime status check may not have completed yet
I0817 16:30:52.792058   27669 kubelet.go:1933] skipping pod synchronization - container runtime status check may not have completed yet
I0817 16:30:57.792298   27669 kubelet.go:1933] skipping pod synchronization - container runtime status check may not have completed yet
```

**Environment**:
- Arktos version (use `kubectl version`): GitCommit:"07296427f661c0b6d8862c3b897c3a0a53858b21"
- Cloud provider or hardware configuration: GCP n1-statndard type vm (1 cpu, 3.75G ram, 40GB disk)
- OS (e.g: `cat /etc/os-release`):Ubuntu 18.04.5 LTS
- Kernel (e.g. `uname -a`):Linux lab-arktos-vm-1 5.3.0-1032-gcp #34~18.04.1-Ubuntu SMP Tue Jul 14 22:07:36 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
- Install tools: arktos-up
- Network plugin and version (if this is a network-related bug): flannel (to be installed)
- Others:

h-w-chen:
**What type of PR is this?**
/kind bug

This PR fixes the bug that kubelet mistakenly skips syncLoopIteration when cni is not installed yet.

Some cni plugins, like Flannel etc, are installed after the cluster has been started (without cni installed initially, however). If kubelet is unable to start cni pods, it will dead lock, and the cluster cannot run any pods at all. 

This PR removes the premature error returning code introduced in implementation of runtime partial readiness feature, hence breaks the dead lock.

Fixes #595

NONE

### DIFF
--- a/pkg/kubelet/kubelet.go
+++ b/pkg/kubelet/kubelet.go
@@ -2352,6 +2352,7 @@ func (kl *Kubelet) updateRuntimeUp() {
 
 	s := toKubeRuntimeStatus(s1)
 	if err := kl.updateOneRuntimeUp(s); err != nil {
+		klog.Errorf("node updateOneRuntimeUp had error: %v", err)
 		return
 	}
 
@@ -2360,18 +2361,20 @@ func (kl *Kubelet) updateRuntimeUp() {
 }
 
 func (kl *Kubelet) updateOneRuntimeUp(s *kubecontainer.RuntimeStatus) error {
-	msg := "either runtime or network is not ready. flag caller to return"
+	msg := "runtime is not ready. flag caller to return"
 	kl.updateRuntimeMux.Lock()
 	defer kl.updateRuntimeMux.Unlock()
 
 	// TODO(random-liu): Consider to send node event when optional
 	// condition is unmet.
 	klog.V(4).Infof("Container runtime status: %v", s)
+
 	networkReady := s.GetRuntimeCondition(kubecontainer.NetworkReady)
 	if networkReady == nil || !networkReady.Status {
+		// do not return error when network is not ready yet, as it would prevent cni pod from starting,
+		// which forms a dead lock in bootstrapping. Logging is just fine.
 		klog.Errorf("Container runtime network not ready: %v", networkReady)
 		kl.runtimeState.setNetworkState(fmt.Errorf("runtime network not ready: %v", networkReady))
-		return fmt.Errorf(msg)
 	} else {
 		// Set nil if the container runtime network is ready.
 		kl.runtimeState.setNetworkState(nil)


pkg/kubelet/kubelet.go
```diff
@@ -2360,18 +2361,20 @@ func (kl *Kubelet) updateRuntimeUp() {
...
 	// condition is unmet.
 	klog.V(4).Infof("Container runtime status: %v", s)
+
 	networkReady := s.GetRuntimeCondition(kubecontainer.NetworkReady)
 	if networkReady == nil || !networkReady.Status {
+		// do not return error when network is not ready yet, as it would prevent cni pod from starting,
+		// which forms a dead lock in bootstrapping. Logging is just fine.
 		klog.Errorf("Container runtime network not ready: %v", networkReady)
 		kl.runtimeState.setNetworkState(fmt.Errorf("runtime network not ready: %v", networkReady))
-		return fmt.Errorf(msg)
```
yb01, line 2374: if this does't return and the compute condition is ready, it will lead the line 2358 and 2359 t be run for each loop in the caller function. could be a potential issue if so.

h-w-chen: this fix is specifically about the dead lock as reported in #595. The dead lock was formed because (s *runtimeState) runtimeErrors() returned error, as there was no lastBaseRuntimeSync timestamp refreshed.

In order to break the dead lock, we need to ensure runtimeState.lastBaseRuntimeSync is touched. The fix is to exactly ensure line 2358 and 2359 be execute, to touch the time stamp, so to break the lock.

There may exist more optimal frequency refreshing the timestamp; however, it seems not a real concern right now, unless it really impacts system performance.

yb01: sounds good. I just checked the community version of this function they does the same thing for only return the function when the compute is not ready. we should be good here.

h-w-chen: Exactly, this regression bug was introduced in commit 1e698f6.
```golang
        if networkReady == nil || !networkReady.Status {
                klog.Errorf("Container runtime network not ready: %v", networkReady)
                kl.runtimeState.setNetworkState(fmt.Errorf("runtime network not ready: %v", networkReady))
+               return fmt.Errorf(msg)
        } else {
```

PR #626, opened 2020-08-26
update design document for cloud-edge communication framework feature

ZongbaoLiu:
This PR updates the design document with follow changes:
1. add design proposal and API definition
2. add implementation steps


docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
```diff
@@ -11,18 +11,255 @@ Because the cloud and edge is located in the network boundary of different secur
...
 
 * Support communication access capability of cloud-edge basic management components
-
 * Support independent IP addressing and IP address convergence of edge management components/services
-
 * Minimize attack surface, zero trust security
 
+## Proposal
+
+The following two solutions do not conflict with each other. The former can converge IP addresses to resolve IP address conflicts. However, the communication needs to be first sent to the gateway and then to the destination service. The latter cannot converge IP addresses but can resolve IP address conflicts. The latter provides end-to-end communication.
```
XiaoningDing, line 27: What does "converge IP addresses" here mean?

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
```diff
@@ -11,18 +11,255 @@ Because the cloud and edge is located in the network boundary of different secur
...
+		"dnsName": "url1.test.com",
+	    "eSites": ["TokyoChibaSite1"],
+	    "ePolicys": ["Policy1", "Policy2"]
+	}
+
+
+#### API Structs
+
+Here are the ESite data struct introduced into arktos，Represent each site, and the central cloud region also belongs to the site：
+
```
XiaoningDing, line 140: It's clearer to use "Edge" as the prefix instead of "E".

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
```diff
@@ -11,18 +11,255 @@ Because the cloud and edge is located in the network boundary of different secur
...
+	
+		// Ip address of the gateway
+		Ip string
+	
+		// Virtual presence ip address cidr of this gateway
+		VirtualPresenceIPcidr string
+	
+		// ESiteName associated to the gateway
+		ESiteName string
+	}
```
XiaoningDing, line 160: Have we considered the approach of defining all objects as CRDs?

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
```diff
@@ -11,18 +11,255 @@ Because the cloud and edge is located in the network boundary of different secur
...
+### Proposal A
+
+The communication of cloud edge components/services is realized based on L4/L7 proxy, which is similar to kubedge's cloud-edge message channel.
+
+Cloudgateway and edgegateway are introduced into the central cloud and each edge site respectively. The edge site edgegateway supports connecting to the cloud gateway of the central cloud when it is started. The service communication between cloud and edge needs to go through these two gateways before arriving. Service access interfaces are provided on these two gateways to receive requests and route requests to target services.
+
+* Cloudgateway and edgegateway are installed and run in the central cloud and each edge site respectively
+* Edgegateway of Edge site can actively establish encrypted links with cloudgateway on the central side, and support two-way message transmission
+* Register the services of edge site and central cloud to cloudgateway
+* Edgegateway in the edge site acts as the service in the central cloud, and exposes the service access interface of the central cloud to the service in the edge site
```
XiaoningDing, line 38: I'm trying to understand the difference between option A and B.  In option A it's also mentioned "...exposes the service access interface of the central cloud to the service in the edge site". What's the difference of "service interface" here than the virtual presentation in option B? Is the service interface in A only network-level?

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
```diff
@@ -11,18 +11,255 @@ Because the cloud and edge is located in the network boundary of different secur
...
+
+### Proposal B
+
+The access of cloud edge components/services is realized based on L3 layer network communication, but the network/IP is not exposed to the remote end, but the components/services are exposed to the remote end.
+
+![](Proposal-B1.png)
+
+Consider providing the concept of virtual presence, which is the local virtual presentation of remote services. In this way, virtual presence can be regarded as a local service with local IP addresses. These IP addresses have nothing to do with the network where the remote service is located, and can be allocated independently. In this way, the problem of unified planning of IP address of VPN solution in the whole network can be solved at the network level. From the perspective of security, virtual presence is completely isolated from the remote service. The client cannot directly access the network where the remote service is located. If the client is attacked, it can not directly attack the server.
+
+![](Proposal-B2.png)
```
XiaoningDing, line 49: My understanding is: the services in cloud have internal IPs that can be accessed from cloud gateways, but not internet or edge sites. If this is the case, shouldn't "FW(ACL)" be put on the cloud service side in the digram?

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
nkwangjun: It means expose only the proxy（or gateway/hub） IP address to the cloud, not all internal IP addresses of the site. The advantage of this method is that the firewall does not need to be configured with a large number of rules and does not need to be frequently changed due to the changes of the edge site network. Converge IP addresses can help minimize the attack surface.

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
nkwangjun: Yes, the idea now is to provide these objects and interfaces through the CRD method.

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
nkwangjun: Service access interface just mean the api endpoint in option A must be corrected.

In simple terms, Option A works at a higher network layer (Layer 4/Layer 7). It is similar to Nginx. It provides backend service agents and transfers cloud or edge sites to other sites.

Option B is a type of (no man-in-the-middle, no session termination) and Virtual Presence is a local virtual presentation of remote services, It can provide full isolation of remote servers, it use of locally assigned virtual presence IPs and External IPs to avoid exposure of site internal addresses. that work at the Layer 3 network layer.
we want to use translation mechanisms to emulate local traffic flow, to implement it.

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
nkwangjun: Yes. The FW is between the cloud service and cloud gateway. Traffic that needs to access cloud services at all edge sites must pass through the cloud gateway, pass through the FW, and finally reach the cloud service. The FW is used for traffic ACL from CloudGateway to CloudService.

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
```diff
@@ -11,18 +11,255 @@ Because the cloud and edge is located in the network boundary of different secur
...
+
+Gateway (cloudgateway/edgegateway) is introduced into the cloud and each edge. Virtual presence of remote service is provided on these gateways, and service components that need communication between cloud and edge site are presented respectively to realize communication access of cloud edge service：
+
```
cathyhongzhang, line 52: Option A also has Cloudgateway and edgegateway which provide the virtual presence. What is the difference?

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
```diff
@@ -11,18 +11,255 @@ Because the cloud and edge is located in the network boundary of different secur
...
+* Edgegateway is installed and run in the central cloud and each edge site respectively, so that these edge sites can access to the central cloud through edgegateway
```
cathyhongzhang, line 53: Does Edgegateway run in both central cloud and edge cloud?

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
```diff
@@ -11,18 +11,255 @@ Because the cloud and edge is located in the network boundary of different secur
...
+* Create edgesite objects to the system and associate edgesite with these edgegateway
+* All components in the cloud and edge that need mutual access are registered with the central cloud cloudgateway in the form of service, and the access policies of these services are configured
```
cathyhongzhang, line 55: Here it is said that  cloud gateway is used in central cloud. I am confused.

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
```diff
@@ -11,18 +11,255 @@ Because the cloud and edge is located in the network boundary of different secur
...
+* The central cloud synchronizes the policy to the corresponding edgegateway, which provides virtual presence for the edge through translation mechanisms to simulate local traffic flow
+
```
cathyhongzhang, line 57: I would suggest that you draw diagrams with example message flows from central to edge and from edge to central to clarify the key difference between option A and option B. From high level, they look kind of similar.

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
nkwangjun: Both options are provide the communication abilities for the component or services running in the cloud and edge site, so they all need to provide "proxy" or "gateway" to help the component and services to communicate. Here we call the "proxy" or "gateway" in Option A and Option B are the Cloudgateway and Edgegateway, which is running in the cloud called cloudgateway, and running in the edge , we call it  edgegateway.
The main differences between Option A and Option B is the "proxy method", Option A just like the nginx, the middle man, it proxy the service visit between the cloud and edge with reverse proxy and running in the network layer 4 or 7. Option B we want to use the virtual presence to presentation the service to local, there is no middle man, the communication between the client and service is directly, t is running in network layer 3

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
nkwangjun: We want to just running the edgegateway in the edge site, the cloudgateway running in the central cloud. the cloudgateway is both the control brain and the traffic gateway in the cloud

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
nkwangjun: Ok, I will add it, thx

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
nkwangjun: Yes, the cloudgateway is running in the cloud, and the edgegateway is running in the edge(docs/design-proposals/cloud-edge-communication/Proposal-B2.png)

deepak-vij:
I am trying to get a grasp on all this. Would it be fair to say that option B is something like a decentralized proxy approach using something like a side-car container? If true, good thing about this approach is that mechanism such as well established mTLS coupled with decentralized PKI based solution could be leveraged for doing AuthN for each service invocation. This is as opposed to reinventing the wheel for doing AuthN. Same side-car proxy can also perform the other gateway functions as well, such as AuthZ, Protocol Translation, NAT etc.

Let me know if I misunderstood it all. Thanks.

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
```diff
@@ -11,18 +11,255 @@ Because the cloud and edge is located in the network boundary of different secur
 However, there may be many problems when VPN is used to directly connect the components/services of different security levels:
```
pdgetrf, line 11: "can be connected through the VPN layer" 

Is this the only other option? is there any review done that collects and compares "current" solutions?

deepak-vij:
Relying on network perimeter technologies (VPNs etc.) for securing infrastructure is definitely both a security risk - since the damage a rouge actor that breaches the perimeter can be enormous - but it also implies that "reachability is authorization", which is typically far too course grained a mechanism for reliable ACLing.

Instead, we need to think in terms of cryptographic authentication and authorization mechanism at the application layer level for inter-service communication. This provides strong access control at an abstraction level and granularity that administrators and services can naturally understand.

Now at a basic service-to-service communication level, each service that runs on the infrastructure needs to assigned an associated verifiable service identity. A service is provided cryptographic credentials that it can use to prove its identity when making or receiving remote procedure calls (RPCs) to other services. These identities are used by clients to ensure that they are talking to the correct intended server, and by servers to limit access to methods and data to particular clients.

The owner of a service can use access management features provided by the infrastructure to specify exactly which other services can communicate with it. For example, a service may want to offer some APIs solely to a specific whitelist of other services. That service can be configured with the whitelist of the allowed service account identities and this access restriction is then automatically enforced by the infrastructure.

As I mentioned earlier that well established mTLS mechanism coupled with decentralized PKI based solution could be leveraged for doing Authentication for each service invocation. Additionally, a side-car proxy can also perform the other gateway functions as well, such as Authorization, NAT etc.

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
pdgetrf: "Edgegateway in the edge site acts as the service in the central cloud, and exposes the service access interface of the central cloud to the service in the edge site" this is quite hard to understand.

what's central cloud? why the word "central"? is it just the cloud?
"acts as the service" --> "acts as a service"
if "service access interface" means api, can we just say API?
if edgegateway acts as a service in the cloud, how can it expose API of cloud to service in the edge site?

consider add some more explanation or an example for this. I still can't understand it.

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
```diff
@@ -11,18 +11,255 @@ Because the cloud and edge is located in the network boundary of different secur
...
+
+Cloudgateway and edgegateway are introduced into the central cloud and each edge site respectively. The edge site edgegateway supports connecting to the cloud gateway of the central cloud when it is started. The service communication between cloud and edge needs to go through these two gateways before arriving. Service access interfaces are provided on these two gateways to receive requests and route requests to target services.
+
+* Cloudgateway and edgegateway are installed and run in the central cloud and each edge site respectively
+* Edgegateway of Edge site can actively establish encrypted links with cloudgateway on the central side, and support two-way message transmission
+* Register the services of edge site and central cloud to cloudgateway
+* Edgegateway in the edge site acts as the service in the central cloud, and exposes the service access interface of the central cloud to the service in the edge site
+* Cloudgateway in the central cloud acts as the service in each edge site and exposes the service access interface of the edge site in the central cloud
```
pdgetrf, line 39: same for this statement. please add an example or diagram for better explanation.

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
```diff
@@ -11,18 +11,255 @@ Because the cloud and edge is located in the network boundary of different secur
...
 
+## Proposal
+
+The following two solutions do not conflict with each other. The former can converge IP addresses to resolve IP address conflicts. However, the communication needs to be first sent to the gateway and then to the destination service. The latter cannot converge IP addresses but can resolve IP address conflicts. The latter provides end-to-end communication.
+
+### Proposal A
+
+The communication of cloud edge components/services is realized based on L4/L7 proxy, which is similar to kubedge's cloud-edge message channel.
+
+Cloudgateway and edgegateway are introduced into the central cloud and each edge site respectively. The edge site edgegateway supports connecting to the cloud gateway of the central cloud when it is started. The service communication between cloud and edge needs to go through these two gateways before arriving. Service access interfaces are provided on these two gateways to receive requests and route requests to target services.
```
pdgetrf, line 33: what does "before arriving" mean?

"Service access interfaces are provided on these two gateways to receive requests and route requests to target services." --> these two gateways provide service access interface (API?) that receives requests and routes requests to target services.

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
```diff
@@ -11,18 +11,255 @@ Because the cloud and edge is located in the network boundary of different secur
...
+
+* Cloudgateway and edgegateway are installed and run in the central cloud and each edge site respectively
+* Edgegateway of Edge site can actively establish encrypted links with cloudgateway on the central side, and support two-way message transmission
```
pdgetrf, line 36: what does actively mean here?

pdgetrf: "message transmission" --> communication

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
pdgetrf, line 27: Please add a reason why two options that are not conflicting with each other are presented? are we supposed to pick one? or are we suppose to find away to merge them? The statement of the intention would help guide reading each of the proposals.

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
```diff
@@ -11,18 +11,255 @@ Because the cloud and edge is located in the network boundary of different secur
...
+* Register the services of edge site and central cloud to cloudgateway
+* Edgegateway in the edge site acts as the service in the central cloud, and exposes the service access interface of the central cloud to the service in the edge site
+* Cloudgateway in the central cloud acts as the service in each edge site and exposes the service access interface of the edge site in the central cloud
+
+### Proposal B
+
+The access of cloud edge components/services is realized based on L3 layer network communication, but the network/IP is not exposed to the remote end, but the components/services are exposed to the remote end.
```
pdgetrf, line 43: "realized based on L3 layer network communication" this is a broad statement that gives little information.  I'm assuming it's supposed to establish a comparison with the L4/L7 method in proposal A. If true, the difference between this and proposal A is what's more worthwhile to be clarified here

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
```diff
@@ -11,18 +11,255 @@ Because the cloud and edge is located in the network boundary of different secur
...
+
+![](Proposal-B1.png)
```
pdgetrf, line 45: there's no mentioning of any of the elements such as C1, S1, S2, S3, S4 from this diagram in the doc. why do we need this graph at all? also it's confusing. what's the difference between S1, S2 and S3 S4? what does dotted line and solid line mean?

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
pdgetrf: agree with @cathyhongzhang this sentence doesn't make sense. it says "Edgegateway is installed and run in the central cloud and each edge site respectively". where does it even mention "cloudgateway"?

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
```diff
@@ -11,18 +11,255 @@ Because the cloud and edge is located in the network boundary of different secur
...
+
+![](Proposal-B2.png)
+
+Gateway (cloudgateway/edgegateway) is introduced into the cloud and each edge. Virtual presence of remote service is provided on these gateways, and service components that need communication between cloud and edge site are presented respectively to realize communication access of cloud edge service：
+
+* Edgegateway is installed and run in the central cloud and each edge site respectively, so that these edge sites can access to the central cloud through edgegateway
+* Create edgesite objects to the system and associate edgesite with these edgegateway
```
pdgetrf, line 54: lots of handwaving here. what does "associate edgesite with these edgegateways" mean? associate how?

also are these 4 bullet points supposed to be consecutive steps?

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
pdgetrf: how does the actual service maintain the mapping with their stub/shadown/phatom or "virtual presence"? let's say if service 3 dies, what happens to its "virtual presence"?

pdgetrf: also, I still don't quite get what's the benefit of proposal B over proposal A? what does the concept of virtual presence really offer other than adding more complexity of mapping with their real presence.

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
```diff
@@ -11,18 +11,255 @@ Because the cloud and edge is located in the network boundary of different secur
...
+		DnsName string
+	
+		// ESite name list
+		ESites []string
+	
+		// EPolicys name list of the service will be allowed
+		EPolicys []string
+	}
+
+## Implementation Steps：
```
pdgetrf, line 245: so at this point which proposal is being used?

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
nkwangjun: what's central cloud? why the word "central"? is it just the cloud?
-- yes, it is just the cloud

"acts as the service" --> "acts as a service"
if "service access interface" means api, can we just say API?
if edgegateway acts as a service in the cloud, how can it expose API of cloud to service in the edge site?
-- In proposal A, edgegateway acts as the proxy of the services running in the cloud

consider add some more explanation or an example for this. I still can't understand it.
-- ok, I will add some example for it

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
nkwangjun: yes，you are right，I will change the description here.edgateway is only running in the edge site，and cloudgateway is running in the cloud.

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
nkwangjun: how does the actual service maintain the mapping with their stub/shadown/phatom or "virtual presence"? let's say if service 3 dies, what happens to its "virtual presence"?
-- In this case, nothing will happen to its "virtual presence". The data connection from service 1 to service 3 will be closed(Because the service 3 dies)

nkwangjun: "virtual presence" is a local virtual presentation of remote services, the benefit of proposal B over proposal A:
a) no man-in-the-middle, no session termination, end-to-end encryption(for example, rest calls(https) between cloud and edge, this communication data is encrypted) 
b) it can easily support some kinds of services(use http、https、kafka、rabbitmq or something else) communications between cloud and edge， if we want to use option A to implement it， we need to implemente based on each protocol we want to support.

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
nkwangjun: Use the vpn for  the data communication between cloud and edge is not the only option， Direct Connect or just use public ip over internet are other options. Vpn is a common choice for data communication between different site (small data center) .
I will add some descriptions here, thx

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
nkwangjun: Ok, I will add an example here for better explanation, thx

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
nkwangjun: "before arriving" will be deleted, thx
These two gateways provide API that receives requests and routes requests to target services.

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
nkwangjun: Here “actively” means “Proactive", I will modify the description for better understanding.

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
nkwangjun: We want to choose proposal B, the description here will be deleted for no ambiguity, thx

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
nkwangjun: I will add description here for the proposal comparison

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
nkwangjun: I will add more description on this diagram here, for better understanding

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
nkwangjun: Proposal B, I will add it before the implementation steps, thx

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
nkwangjun: yes, I think I need to add a more detail description for this proposal, thx

nkwangjun:
> I am trying to get a grasp on all this. Would it be fair to say that option B is something like a decentralized proxy approach using something like a side-car container? If true, good thing about this approach is that mechanism such as well established mTLS coupled with decentralized PKI based solution could be leveraged for doing AuthN for each service invocation. This is as opposed to reinventing the wheel for doing AuthN. Same side-car proxy can also perform the other gateway functions as well, such as AuthZ, Protocol Translation, NAT etc.
> 
> Let me know if I misunderstood it all. Thanks.

Thanks for your comments.
It likes "a side-car container" for the whole site, it proxy the data communication between the sites.
Because it is used for the communication between services on infrastructure in the edge cloud，and these services may not be deployed through kubernetes or arktos yet, even it is not running in a container.

deepak-vij:
> > I am trying to get a grasp on all this. Would it be fair to say that option B is something like a decentralized proxy approach using something like a side-car container? If true, good thing about this approach is that mechanism such as well established mTLS coupled with decentralized PKI based solution could be leveraged for doing AuthN for each service invocation. This is as opposed to reinventing the wheel for doing AuthN. Same side-car proxy can also perform the other gateway functions as well, such as AuthZ, Protocol Translation, NAT etc.
> > Let me know if I misunderstood it all. Thanks.
> 
> Thanks for your comments.
> It likes "a side-car container" for the whole site, it proxy the data communication between the sites.
> Because it is used for the communication between services on infrastructure in the edge cloud，and these services may not be deployed through kubernetes or arktos yet, even it is not running in a container.
> And we want to use mTLS for the gateways communication，and other gateway functions like NAT as you mentioned, thanks.

One of the challenges of mTLS is management of PKI certificates in a hierarchical cloud environment (edge->CentralCloud; edge->edge). Current centralized PKI system responsible for minting, issuing and managing certificates is really problematic. This is due to the fact that the connectivity between edge clouds and centralized cloud may be intermittent. In such case, edge clouds may work in offline mode. Public Key Infrastructure (PKI) approach may not be viable in such an environment. 

A standardized SPIFFE/SPIRE approach addresses this issue. In order for Public Key Infrastructure (PKI) approach to function properly, following is the deployment topology for SPIFFE in order to manage “Availability” in the edge cloud environment:
●	To have main SPIRE server deployed along with the Root CA at the centralized cloud level.
●	Each gateway server (a designated edge node or a regional edge site) hosts the intermediate SPIRE server along with the intermediate CA. Gateway server in this case is a logical entity. (Edge Cloud Side)
●	The intermediate SPIRE is the actual SPIRE server that issues out workload identities (SPIFFE Verifiable Identity Documents) at the edge cloud level.
●	In order for services to communicate with other across intermediate SPIRE servers, trust chain is established to the Root CA. All intermediate SPIRE servers trust the Root CA.

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
```diff
@@ -30,28 +30,67 @@ edge server -----> Proxy A(in the edge) <----------> Proxy B(in the cloud) -----
...
+
+So if the rest call(http/https request) is the only communication type between the cloud and the edge, maybe it is a simple/good choice. 
+
+Option 2: The proxy works at network layer 4, do the udp/tcp proxy.
+
+For example, we want to support rest calls(http/https), message queue connection(rabbitmq/kafka), and something else based on the udp protocol.
+The proxy works as a udp server, and it forward the udp connection/traffic to the destination site.
+
+It works just like a nginx, but there are some differences:
+1. It is not proxy the traffic to the destination service directly, but to the other proxy in other site
```
pdgetrf, line 52: It is not ---> It does not

pdgetrf: to this statement, can nginx tell whether it's proxying to a destination service directly or another proxy?

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
```diff
@@ -30,28 +30,67 @@ edge server -----> Proxy A(in the edge) <----------> Proxy B(in the cloud) -----
...
+It needs to solve the port allocation and traffic encryption/decryption problems
+
+![](Proposal-B4.png) 
 
 ## Proposal B
 
-We want to implement it based on network layer 3, and we also need to solve the ip address conflict and security problems.
+We want to implement the communication framework based on the network layer 3, and also we need to solve the ip address conflict and security problems. We can use ip translation to do this. 
+
+We need to add the gateway/proxy for the edge site and the cloud for the communication, we call it edgegateway in each edge site, and in addition to communication, cloudgateway in responsible for central control. We don't expose the network to the cloud or the edge site(cause too much attach surface), we expose the services to the gateway.
```
pdgetrf, line 62: in responsible --> is responsible

attach surface --> attack surface

so in short. Both cloud and edge sides have gateways. Services in both sides are behind the gateways.

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
```diff
@@ -30,28 +30,67 @@ edge server -----> Proxy A(in the edge) <----------> Proxy B(in the cloud) -----
...
+Here we want to introduce a concept "virtual presence":
 * It is a local virtual presentation of remote services
 * It is full isolation of remote servers
-* It exposes services to remote sites, do not expose networks
-* It uses locally assigned virtual presence ips to avoid exposure of site internal addresses
-* It use translation mechanisms to emulate local traffic flow
-* There is no man-in-the-middle, no session termination, it provide end-to-end encryption
 
-![](Proposal-B1.png)
+If the service is exposed to the edge site, there will be a virtual presence for the service in this edge site. The clients in this edge site which want to communicate to the service will be told to communicate to the virtual presence of the service which is on the gateway of the edge site. And our gateway will forward traffic in secured tunnel using local virtual presence ip address to the destination gateway and services.
```
pdgetrf, line 68: just so I understand correctly: the difference between this solution and the current VPN solution is that clients on the edge site do not communicate with the service in the cloud using their actual ip, but the ip of the gateway instead. gateway still talks to the cloud (the cloudgateway actually) through VPN. This reduces the attack surface to just the edgegateway.

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
```diff
@@ -30,28 +30,67 @@ edge server -----> Proxy A(in the edge) <----------> Proxy B(in the cloud) -----
...
+1. There is ip conflict in edge site A and edge site B, the nova-compute server use the same ip address for communication to the cloud.
+2. We can expose the rabbitmq service which is in the cloud to the edge site A and edge site B, it will expose the rabbitmq service url for the edge site. There will be a virtual presence for the rabbitmq service in edge site A, and another virtual presence for the rabbitmq service in edge site B.
+The virtual presence will be assigned a local virtual ip address of the edge site by the controller(CloudGateway), the local virtual ip is in the edge site address spaces which is predefined for the edge site, and it is just used in the edge site, so it is no need to global planning for the cloud.
```
pdgetrf, line 75: "local virtual ip" I think it should be just "local ip" no?

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
```diff
@@ -30,28 +30,67 @@ edge server -----> Proxy A(in the edge) <----------> Proxy B(in the cloud) -----
...
+3. The nova compute need to use the local dns which is a component of the edgegateway A. Edgegateway A will resolve the rabbitmq service url to the virtual presence ip address which is routed to the edgegateway A.
```
pdgetrf, line 76: so all nova compute service in all the edge sites will have the same url for rabbitmq. edgegateway will resolve and direct them to their local virtual presence

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
nkwangjun: Yes, I will remove the "virtual"

docs/design-proposals/cloud-edge-communication/cloud-edge-communication-framework.md
nkwangjun: "gateway still talks to the cloud (the cloudgateway actually) through VPN. "    gateway can talk to the cloud through VPN, or with security tunnel which is build by the gateways.  Currently we choose the second option, we don't want to use the VPN for our cloud-edge communication.

pdgetrf:
@nkwangjun About "data plane using VPN" question that @XiaoningDing asked at our last meeting, do we have any update? The question is that if data plane will continue to use VPN for cloud-edge communication, does this gateway design really solve the security threat goal?

nkwangjun:
> @nkwangjun About "data plane using VPN" question that @XiaoningDing asked at our last meeting, do we have any update? The question is that if data plane will continue to use VPN for cloud-edge communication, does this gateway design really solve the security threat goal?

"data plane" between cloud and edge， currently using encrypted VXLAN tunnel，currently the data plane gateway in the edge may be attacked and some basic protection is performed. For example, the key cannot be persisted and is periodically updated.

PR #629, opened 2020-08-26
e2e tests for deployment/replicaset/serviceacount/token/endpoints controllers

chenqianfzh:
This PR adds the e2e test suites for the following multi-tenancy controllers:
1. deployment & replicaset controller
2. serviceaccount & token controller
3. endpoints controllers.

The test suites were verified in my dev box. The following are the screenshots:
image

image

Tracking issue: https://github.com/futurewei-cloud/arktos/issues/559

### DIFF
--- a/test/e2e/arktos/multi_tenancy/run-e2e-test.sh
+++ b/test/e2e/arktos/multi_tenancy/run-e2e-test.sh
@@ -23,7 +23,10 @@ script_root=$(dirname "${BASH_SOURCE}")
 repo_root=$(cd $(dirname $0)/../../../.. ; pwd)
 
 #put the test suite file names below, one line one suite. The test suites will be run in the order defined.
-test_suite_files="tenant_init_delete_test.yaml"
+test_suite_files="multi_tenancy_controller/test_sa_token_controller.yaml \
+				  multi_tenancy_controller/test_endpoints_controller.yaml \
+				  multi_tenancy_controller/test_deployment_replicaset_controller.yaml \
+				  tenant_init_delete_test.yaml"
 test_suite_file_directory=${repo_root}/test/e2e/arktos/multi_tenancy/test_suites/
 
 # The values of timeouts and retry intervals are in the unit of second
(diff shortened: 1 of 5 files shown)


test/e2e/arktos/multi_tenancy/run-e2e-test.sh
```diff
@@ -23,7 +23,10 @@ script_root=$(dirname "${BASH_SOURCE}")
 repo_root=$(cd $(dirname $0)/../../../.. ; pwd)
 
 #put the test suite file names below, one line one suite. The test suites will be run in the order defined.
-test_suite_files="tenant_init_delete_test.yaml"
```
h-w-chen, line 26: it seems test code is sort-of mixed with test data files. Could they be further separated, so that when new test cases are added, we don't have to touch this test code file at all?

test/e2e/arktos/multi_tenancy/test_suites/multi_tenancy_controller/test_deployment_replicaset_controller.yaml
```diff
@@ -0,0 +1,141 @@
+#~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ Deployment & ReplicaSet Controller Tests ~~~~~~~~~~~~~~~~~~~~~~
```
h-w-chen, line 1: like this BDD style: easy to understand; end user should be able to add test cases without problem.

test/e2e/arktos/multi_tenancy/test_suites/multi_tenancy_controller/test_deployment_replicaset_controller.yaml
```diff
@@ -0,0 +1,141 @@
...
+
+######################################################
+# test variables
+######################################################
+Variables:
+  test_ns: random_8
+  test_tenant: random_8
+
+###########################################################################################################
+# test setup
```
h-w-chen, line 13: comment here tries to tell test steps aprt; nice attempt. Could it be part of the test case output too, hence more visible and useful when looking at test case output?

test/e2e/arktos/multi_tenancy/test_suites/multi_tenancy_controller/test_deployment_replicaset_controller.yaml
chenqianfzh: :-)

test/e2e/arktos/multi_tenancy/test_suites/multi_tenancy_controller/test_deployment_replicaset_controller.yaml
chenqianfzh: That is good suggestion. I will add this feature to the test tool.

test/e2e/arktos/multi_tenancy/run-e2e-test.sh
chenqianfzh: Actually, the code and the test suite files are separate. The code after under "cmd" subfolder, while the test suite files are under subfolder "test_suite".

PR #633, opened 2020-08-27
replace the docs file of global scheduler architecture design specification with md file 

cathyhongzhang:
**What type of PR is this?**

> /kind design

This PR upload the global scheduler architecture design spec in md file format which replces previously uploaded docs file format

change the design spec file format


docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
```diff
@@ -0,0 +1,416 @@
...
+- High scalability to support a high volume of incoming scheduling requests.
+- Rapid decision making and low scheduling latency to support a high allocation rate
+- Globally optimal scheduling to support high resource utilization across all cloud data centers and cloud edge sites
+- Satisfying the user-supplied placement constraints
+
+These goals usually conflict with each other. As analyzed in the previous sessions, each type of scheduler architecture only meets a few of the above goals, but not all the goals.
+
+# 2. Global Scheduler Overall Architecture Design
+
+## 2.1 Key Requirement Analysis
```
nkwangjun, line 130: Whether to add batch scheduling requirements
Unlike Kubernetes, you can apply for a VM or container at a time. Multiple resources can be applied for at a time. All resources can be successfully scheduled. Otherwise, the scheduling fails and no VM is created
I think this may be a key technology and requirement

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
```diff
@@ -0,0 +1,416 @@
...
+
+In our design, all the Global Schedulers operate in a shared-state manner. That is, they shared one global resource DB which holds the cluster-level resource information of all the clusters. The DB size will be manageable since the DB only holds cluster-level resource information, such as available cluster CPU, available cluster memory, supported VM flavor list, and network topology of clusters, network proximity of clusters, etc.. Unlike the Shared-State Multi-Scheduler Architecture outlined in section 1.3.3, the global resource DB is not replicated by each Global Scheduler. Instead, they share and read/update information from one global resource DB.  In our new design, the Global Schedulers run in an optimistic concurrent way instead of using pessimistic concurrency control, which removes the locking latency.
```
nkwangjun, line 154: Is there a line missing in the high-level architecture shown in the following figure? Is the global scheduler obtaining information from the global resource database? Is this the Dynamic resource info cache?

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
```diff
@@ -0,0 +1,416 @@
...
+
+Unlike existing multi-scheduler architecture design as described in section 1.3.2 and 1.3.3, in which the incoming scheduling requests are either distributed to all schedulers or a scheduler in a round-robin way, in our design, the incoming scheduling requests are distributed to a Global Scheduler based on the request's profile. Section 2.5.4 illustrates how this distribution mechanism combined with the Global Scheduler partitioning scheme (Section 2.5.3) will mitigate the scheduling conflict when all the Global Schedulers run concurrently.
+
+In summary, this design eliminates the issues associated with existing Two-Level Scheduler Architecture, Siloed Multi-Scheduler Architecture, and Shared-State Multi-Scheduler Architecture. It renders good scalability, low latency, and global optimal scheduling while satisfying the user-supplied placement constraints.
+
+On the southbound, it integrates with different types of clusters via a flexible plugin design.
+
+Figure.1 is an illustration of the high-level architecture.
+![image.png](/images/2.3.png)
```
nkwangjun, line 163: 1. Does the distributor in the figure need to watch vm/container resources through the API server? Is the arrow from the API server to the distributor in reverse?
2. How does the Distributor communicate with the global scheduler?
Distributor and Global Scheduler are different processes, right?
3. Indicates whether the request dispatcher supports batch operations. Currently, batch operations may also be a major requirement scenario.

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
```diff
@@ -0,0 +1,416 @@
...
+    <td>
+      <li>Register with API server to get notification of new or changes of VM/Container object</li>
+      <li>Distribute the VM/Container scheduling requests to an appropriate Home Global Scheduler</li>
+      <li>Support multiple partitions for scalability. If multiple partitions are needed, incoming requests from the API servers will be load balanced to those Distributor partitions. That is, each Distributor will be connected to a subset of API Servers and will handle a subset of incoming requests</li>
+      <li>Each Distributor is connected to all Global Schedulers and will distribute an incoming VM/Container request to a Home Global Scheduler based on tag matching</li>
+    </td>
+  </tr>
+  
+  <tr>
+    <td>Dynamic Partition Mapper</td>
```
nkwangjun, line 228: 1. Check whether each cluster is allocated to only one partition and scheduled by only one scheduler.
2. If the current partition scheduling mode is used, is batch scheduling not supported? For example, a resource request for batch scheduling may need to apply for resources in multiple partitions based on geographical distribution, which cannot be met by a single partition or a single scheduler.

nkwangjun (commented):
Questions about the overall architecture

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
```diff
@@ -0,0 +1,416 @@
+- [1. Introduction](#1-introduction)
+  - [1.1 Problem Statement](#11-problem-statement)
+  - [1.2 Goals and Scope](#12-goals-and-scope)
```
allison-wangnannan, line 3: Our design constraints 1K DCs+Edge sites ,the number of site depend on how we define the edge ： edge DC or  edge device。

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
```diff
@@ -0,0 +1,416 @@
...
+    - [1.3.3 Shared-State Multi-Scheduler](#133-shared-state-multi-scheduler)
+    - [1.3.4 Siloed Multi-Scheduler with Intercommunication](#134-siloed-multi-scheduler-with-intercommunication)
+    - [1.3.5 Siloed Multi-Scheduler with Arbitrator](#135-siloed-multi-scheduler-with-arbitrator)
+    - [1.3.6 Two-Level Scheduler](#136-two-level-scheduler)
+    - [1.3.7 Summary of Scheduler Design Pain Points](#137-summary-of-scheduler-design-pain-points)
+- [2. Global Scheduler Overall Architecture Design](#2-global-scheduler-overall-architecture-design)
+  - [2.1 Key Requirement Analysis](#21-key-requirement-analysis)
+  - [2.2 Key Points of this New Scheduler Architecture Design](#22-key-points-of-this-new-scheduler-architecture-design)
+  - [2.3 High-Level Architecture Design](#23-high-level-architecture-design)
+  - [2.4 Architecture Component Description](#24-architecture-component-description)
```
allison-wangnannan, line 16: The API of global scheduler may only support schedule ability like allocation /release.Scheduler itself cannot do the CRUD of VM.

allison-wangnannan, line 16: We have considered to use redis as a  memory cache before, but if we chose etcd now , may be we need to consider the performance which is often be challenged with etcd.

allison-wangnannan, line 16: Global scheduler: This interaction need cluster scheduler have the capability to accept scheduler request but not do the create VM task.

allison-wangnannan, line 16: Migration:This interaction need cluster scheduler have the capability to accept scheduler request but not do the create VM task.

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
```diff
@@ -0,0 +1,416 @@
...
+  - [2.5 Key Global Scheduler Mechanism Design](#25-key-global-scheduler-mechanism-design)
+    - [2.5.1 Shared State Lock-Free Optimistic Concurrent Scheduling Operation](#251-shared-state-lock-free-optimistic-concurrent-scheduling-operation)
+    - [2.5.2 Multi-Dimensional Optimization Based Scheduling Algorithm](#252-multi-dimensional-optimization-based-scheduling-algorithm)
```
allison-wangnannan, line 19: That means global scheduler must collect quantity information of all nodes, and  that conflict with the aim  which we don’t want to have a  huge DB.
2、	What the finish condition of  scan the  largest available node?  Is ’t just a quantity filter?
3、	If  we  cannot get the detail like NUMA info，we   may not avoid the case that the  total quantity enough but  actually cannot host the VM .

allison-wangnannan, line 19: Global scheduler do not  decision the placement of a  VM onto a host, I think we don’t need to consider the dimension of per node capacity.

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
```diff
@@ -0,0 +1,416 @@
...
+
+Our global scheduler does the scheduling/placement of a VM/Container to the granularity of a cluster. To avoid a huge DB, we should not store every node's allocable capacity information in the global DB. But if we only store each cluster's accumulated allocable capacity, "_quantity check_" cannot be done properly. This is because even a cluster's accumulated allocable capacity is larger than a VM/Container's quantity requirement, the cluster may have no single node which has enough resource to host/run the VM/Container. An "average per node allocable capacity" is a better way. But "_quantity check_" against the "average per node allocable capacity" is not good enough either. As shown in Figure 2, the new VM requires 128M memory and Cluster1's "average per node available memory resource" is 110M memory. Although Node 3 in Cluster 1 has 256M, which meets the VM's memory requirement, the "_quantity check_" against Cluster 1 will fail and Cluster 1 will be incorrectly filtered out of the candidate list.
```
allison-wangnannan, line 320: 1、That means global scheduler must collect quantity information of all nodes, and  that conflict with the aim  which we don’t want to have a  huge DB.
2、	What the finish condition of  scan the  largest available node?  Is ’t just a quantity filter?
3、	If  we  cannot get the detail like NUMA info，we   may not avoid the case that the  total quantity enough but  actually cannot host the VM .

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
```diff
@@ -0,0 +1,416 @@
...
+![image.png](/images/2.5.2.png)
+
+After the scheduling algorithm finds the candidate list of clusters for a VM/Container, it goes to the second phase---scoring phase.  In the scoring phase, the scheduling algorithm runs a set of functions to score each cluster in the candidate list and picks a cluster with the highest score.
+
+Factors that need to be considered for scoring include the following (more can be added in the future):
+
+1. average per node allocable capacity. The more allocable capacity, the higher the score. But a VM/Container's resource requirement spans more than one resource type. A cluster may have more "per node available memory resource" but less "per node available vCPU resource" than another cluster. How to define "more" or "less"? Section 2.5.2.1 describes the way to define this.
```
allison-wangnannan, line 329: Global scheduler do not  decision the placement of a  VM onto a host, I think we don’t need to consider the dimension of per node capacity.

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
```diff
@@ -0,0 +1,416 @@
...
+  <tr>
+    <th>Subsystem or Component</th>
+    <th>Description</th>
+    <th>Roles and Responsibilities</th>
+  </tr>
+  
+  <tr>
+    <td>API Server</td>
+    <td>
+    	API server exposes VM/Container CRUD APIs to clients, who use the Global Scheduler Service to request/read/update/delete VM/Container resource
```
allison-wangnannan, line 181: The API of global scheduler may only support schedule ability like allocation /release.Scheduler itself cannot do the CRUD of VM.

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
```diff
@@ -0,0 +1,416 @@
...
+      <li>Distribute data store for high availability</li>
+      <li>Support Data Persistency</li>
+      <li>Support both strong and eventual data consistency</li>
+      <li>Support multiple partitions for scalability</li>
+      <li>Notify object changes to API server which will notify the VM/Container Request Distributor component</li>
+    </td>
+  </tr>
+  
+  <tr>
+    <td>Dynamic Cluster Resource Info Cache</td>
```
allison-wangnannan, line 205: Dynamic Cluster Resource Info Cache:We have considered to use redis as a  memory cache before, but if we chose etcd now , may be we need to consider the performance which is often be challenged with etcd.

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
```diff
@@ -0,0 +1,416 @@
...
+      <li>Determine the set of geolocations/regions/AZs each Global Scheduler is associated with</li>
+      <li>Construct the corresponding resource profile tags for each Global Scheduler</li>
+      <li>Monitor the workload of all the Global Schedulers</li>
+      <li>Dynamically adjust the assignment of geolocations/regions/AZs and corresponding profile tags to avoid overloading any single Global Scheduler</li>
+      <li>Update each Global Scheduler's workload, geolocation, resource profile tags in the DB cache</li>
+    </td>
+  </tr>
+  
+  <tr>
+    <td>Global Scheduler</td>
```
allison-wangnannan, line 240: Global scheduler: This interaction need cluster scheduler have the capability to accept scheduler request but not do the create VM task.

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
```diff
@@ -0,0 +1,416 @@
...
+      <li>Monitor each VM's application flow volume</li>
+      <li>Retrieve the application flow's origins based on their source IPs</li>
+      <li>Analyzes the application flow's geolocation and volume information to determine whether there is a need to migrate the VM to another location or to scale out more VMs to support the application flow</li>
+      <li>Determine the number of new VMs that need to be created to serve the application traffic flow</li>
+      <li>Send migration instruction to the Migration Manager for VM migration or send scale-out/in instruction to the VM Horizontal Auto Scaler for auto-scaling out/in</li>
+    </td>
+  </tr>
+  
+  <tr>
+    <td>Migration Manager</td>
```
allison-wangnannan, line 278: Migration:This interaction need cluster scheduler have the capability to accept scheduler request but not do the create VM task.

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
```diff
@@ -0,0 +1,416 @@
...
+
+# 1. Introduction
+
+## 1.1 Problem Statement
+
+Edge Cloud brings computing resources to the edge of the network and provides secure, low latency, and high QoS computing service to mobile applications/devices at the edge and mitigates the network congestion on the Internet. Edge Cloud has several characteristics: the number of Edge Cloud Sites is large, they are all distributed, and each one has a small limited resource pool. It could happen that a resource request, e.g. a VM or container request, cannot be served by an Edge Cloud site due to resource limitation. So, resource sharing and coordination among multiple Edge Cloud Sites are needed, i.e. Edge Computing must globally support resource scheduling.
+
+Global scheduling is also needed in the Public Cloud and Hybrid Cloud. At any given time, one cloud data center may have resource strain while another cloud data center may be in light usage. Global scheduling makes better scheduling decisions and renders better resource utilization and service performance.
+
+## 1.2 Goals and Scope
```
allison-wangnannan, line 48: 1、Our design constraints 1K DCs+Edge sites ,the number of site depend on how we define the edge ： edge DC or  edge device。
2、Do we need to  clarify QPS

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
cathyhongzhang: Yes, I will add that in.

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
cathyhongzhang: We can discuss this in next meeting. Here is one way to do it---The Dynamic resource info cache is part of the global DB. The difference is that these info does not need persistency so "create" and "update" operation will be much faster. We will leverage the API server to handle the interaction with both DB and cache, similar to what K8S does. That is, the Global Scheduler obtains info from the global DB and cache through API server. API server has cache itself and may return the info from its cache if it is there.  Global Scheduler will have its own cache for the info and may get the info from its own cache.

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
cathyhongzhang: 1. yes
2. Good question. This is open for discussion. There are two ways to implement this. 
a) Each scheduler is a separate process and the distributor is another process which interacts with the API Server to get the VM/Container events. The distributor can list&watch VM/Container events with API server through existing K8S "informer" mechanism. The interface between the distributor and each scheduler can be based on the same "informer" mechanism, i.e. each scheduler will register with the distributor a "watch for its associated resource profile tags" and get filtered notification of those VM/Containers that match the registered tags. The interface between the distributor and each scheduler can alternatively be gRPC.  
b) Each scheduler is a thread and has its own scheduling event queue. All schedulers share the same process but have separate scheduling event queues. The distributor runs inside the scheduling process and list&watch VM/Container events with API server through existing K8S "informer" mechanism. Then the distributor puts filtered VM/container requests into each scheduler's queue based on match between the VM/Container's profile and the scheduler queue's profile tags.

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
cathyhongzhang: I have not got time to think a lot about batch scheduling. Originally we only consider support for replicaset scheduling. For batch scheduling, we need to first define if batch scheduling requires atomic operation or not. If the requests in the batch has different profiles, then they will be distributed to different schedulers (or schediuling partitions).  If the batch needs atomic operation, i.e. either all succeed or all fail, then the batch object status should not be updated until all requests associated with this batch is received. Some structure outside these schedulers must track the scheduling result of all the batch requests and do the update. In implementation, we can probably put this tracking functionality inside the distributor.

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
cathyhongzhang: No, it does not need to collect info of all nodes, only info of the largest node which the cluster provides.

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
cathyhongzhang: This is the scroing phase.  Although the global scheduler does not place a VM onto a node, a cluster that has high average per-node capacity is better than a cluster that has low average per-node capacity if the policy is to have even resource distribution across clusters. Of course, the average per-node capacity is just one factor in the final score and its weight can be adjusted.

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
cathyhongzhang: Ok. will update

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
cathyhongzhang: Database selection can be redis or any other cache DB. The interface API to Redis and ETCD are different, but the general design is the same.

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
allison-wangnannan: How to defined the largest? I mean what is the terminate condition when i rank the host.

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
nkwangjun: Yes, I think our batch scenario, needs atomic operation.

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
nkwangjun: Ok, I think the global scheduler needs to obtain the resource information required for scheduling. Therefore, the global scheduler needs to connect to the database directly or access the database through the API server.

cathyhongzhang: The global scheduler will access the DB through the API server, so it will connect directly with the API server, not the DB.

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
cathyhongzhang: As stated in the document, we will collect the resource information of several largest available nodes from each cluster (a few with the largest CPU, a few with the largest memory, etc.). Let's call these nodes "potentially feasible nodes". Then the filtering algorithm will run through these "potentially feasible nodes" to select candidate list.

allison-wangnannan: We need to deployment a group of resources on the edge  like   2VM of  4U8G ,do we only use 4U8G to match the largest host or use the total cpus 8U16G?

cathyhongzhang: Are you talking about bacth scheduling? If these two VMs are assigned to the same Home Scheduler partition, then the 8U16G will be used. If there is no cluster whose lagest node has this capacity, then further check will be done to see if there are two clusters in this partition whose largest nodes both have 4U8G capacity.

allison-wangnannan: I suggest to use local-agent do the capacity estimate, it can use the detail infomation of each node  or  each NUMA  to improve the accurancy ,and do not need to put the stats to global scheduler directly .

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
allison-wangnannan: The goals of  global scheduler is manager 1K cluster include the  edge and region , support 100 QPS request. So ,we can make this estimate:
1、100 QPS resquest need  100/s full table query or condition query.
2、1K+cluster  update info per min,so need 1K write/s.

docs/design-proposals/global-scheduler/GlobalSchedulerArchitectureDesignSpecificationFirstDraft.md
cathyhongzhang: Yes, I am thinking about using local agent too.

dingyin (approved):

PR #640, opened 2020-08-31
Fix API server config update "concurrent map iteration and map write" (bug 639)

Sindica:


staging/src/k8s.io/client-go/rest/config.go
```diff
@@ -191,10 +191,10 @@ func (ag *Config) updateConfig() {
 
 	// assign new hosts
 	i := 0
-	klog.V(6).Infof("Len of new server %d; len of new config %d", numOfServers, len(ag.config))
+	klog.V(4).Infof("Len of new server %d; len of new config %d", numOfServers, len(ag.config))
 	for _, ss := range newApiServers {
 		ag.config[i].Host = fmt.Sprintf("%s://%s:%d/", ss.Ports[0].Name, ss.Addresses[0].IP, ss.Ports[0].Port)
```
chenqianfzh, line 196: minor: remove line 198 and change line 195 to "for i, ss := ...."

staging/src/k8s.io/client-go/apiserverupdate/update_test.go
```diff
@@ -52,6 +77,70 @@ func TestSetAPIServerConfig(t *testing.T) {
...
+			go setAndReadAPIServerConfig(&wg, epMap1)
+			go setAndReadAPIServerConfig(&wg, epMap2)
+		}
+		wg.Wait()
+	}
+
+	// final test
+	readEPMap := GetAPIServerConfig()
+	assert.NotNil(t, readEPMap)
+	assert.True(t, len(readEPMap) <= 2)
```
chenqianfzh, line 122: seems to me that we also need to check assert.True(t, len(readEPMap) >= 1)

staging/src/k8s.io/client-go/apiserverupdate/update_test.go
```diff
@@ -52,6 +77,70 @@ func TestSetAPIServerConfig(t *testing.T) {
 	assert.Equal(t, "", ss2.Addresses[0].Hostname)
 }
 
+func setAndReadAPIServerConfig(wg *sync.WaitGroup, epMap map[string]v1.EndpointSubset) {
+	wg.Add(1)
+	SetAPIServerConfig(epMap)
+	readEPMap := GetAPIServerConfig()
+	for sg, ss := range readEPMap {
+		time.Sleep(10 * time.Millisecond)
+		klog.V(6).Infof("Make sure read server group %s and endpoints %v", sg, ss)
```
chenqianfzh, line 86: incomplete log message: make sure what?

Sindica: "read server group and endpoints"

staging/src/k8s.io/client-go/apiserverupdate/update_test.go
Sindica: line 123 does the same thing.

PR #657, opened 2020-09-01
Use wait group in AggregatedWatcher to avoid panic "send on closed channel"

Sindica:


staging/src/k8s.io/client-go/rest/config.go
```diff
@@ -157,9 +157,11 @@ func (ag *Config) WatchUpdate() *bcast.Member {
 
 func (ag *Config) ToString() string {
 	text := ""
+	ag.mux.RLock()
```
jshaofuturewei, line 160: Is it added for the unit test TestInClusterConfig ?

Sindica: This function is currently used only in UT. Just add lock for general purpose. It can be used by other code in the future.

staging/src/k8s.io/apimachinery/pkg/watch/watch.go
```diff
@@ -161,13 +163,16 @@ func (a *AggregatedWatcher) AddWatchInterface(watcher Interface, err error) {
 
 func (a *AggregatedWatcher) closeWatcher(watcher Interface, stopCh *bcast.Member) {
 	watcher.Stop()
+	a.wg.Done()
 	a.stopChGrp.Leave(stopCh)
 	a.removeWatcherAndError(watcher)
 
 	if !a.allowWatcherReset && a.stopChGrp.MemberCount() == 0 {
-		//klog.Infof("Close watcher %v caused aggregated channel %v closed", watcher, a.aggChan)
+		//klog.V(4).Infof("Close watcher %v caused aggregated channel %v closed", watcher, a.aggChan)
```
jshaofuturewei, line 171: My suggestion is to uncomment the logs. It might be good to collect information in case  there are other close channel issues.

PR #671, opened 2020-09-03
Add mizar-pod-controller and Grpc client

Hong-Chang:
mizar-pod-controller is one of the mizar controllers that listen to cluster events and send corresponding info to mizar backend.

**What type of PR is this?**
/kind feature


cmd/kube-controller-manager/app/mizarcontrollers.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"net/http"
+
+	"k8s.io/klog"
+	controllers "k8s.io/kubernetes/pkg/controller/mizarcontrollers"
+)
+
+const (
+	workers = 4
```
h-w-chen, line 27: is there a reason we choose 4 here?

pkg/controller/mizarcontrollers/mizar-pod-controller.go
```diff
@@ -0,0 +1,194 @@
...
+	if err != nil {
+		return err
+	}
+
+	obj, err := oc.lister.PodsWithMultiTenancy(namespace, tenant).Get(name)
+	if errors.IsNotFound(err) {
+		klog.V(4).Infof("%v %v cannot be found", controllerFor, key)
+		return nil
+	} else {
+		klog.V(4).Infof("Handling %v %s/%s/%s hashkey %v for event %v", controllerFor, tenant, namespace, obj.Name, obj.HashKey, eventType)
```
h-w-chen, line 182: Isuggest to add logic of processing potential errors

pkg/controller/mizarcontrollers/util.go
```diff
@@ -0,0 +1,19 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package mizarcontrollers
+
+type KeyWithEventType struct {
```
h-w-chen, line 16: this small file seems to able to incorporate in other files, as it only has very limited stuff.

h-w-chen (commented):
Thanks for the effort of adding mizar pod controller skeletal code; good first move. 
code generally looks fine; left a few feedback. Will take a look again after they are addressed.

chenqianfzh:
I have an overall suggestion. Can we change the name of controller to MizarPodController, instead of just pod controller?

Though there is no other controller watching pod now in Arktos, but it is likely there may be another watching-pod controller in the future, since pod is such a fundamental resource.

cmd/kube-controller-manager/app/mizarcontrollers.go
```diff
@@ -0,0 +1,39 @@
...
+)
+
+func startMizarPodController(ctx ControllerContext) (http.Handler, bool, error) {
+	controllerName := "mizar-pod-controller"
+	klog.V(2).Infof("Starting %v", controllerName)
+
+	go controllers.NewPodController(
+		ctx.InformerFactory.Core().V1().Pods(),
```
chenqianfzh, line 35: Is this controller just about mizar pods? If Mizar pods have their own label, I suggest we use a filtered informer here. That should benefit the performance, especially if the percentage of mizar pods is not high.

pkg/controller/mizarcontrollers/mizar-pod-controller.go
```diff
@@ -0,0 +1,194 @@
...
+
+		if eventType == "create" {
```
chenqianfzh, line 184: I think it is ideal to use "switch-case" here.

Besides, the handling of "delete" and non-recognized event-type are missing.

pkg/controller/mizarcontrollers/mizar-pod-controller.go
```diff
@@ -0,0 +1,194 @@
...
+	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/kubernetes/pkg/controller"
+
+	"k8s.io/klog"
+)
+
+const (
+	controllerKind = "pod"
+	controllerFor  = "mizar_pod"
```
chenqianfzh, line 39: If you decide to follow my suggestion to use the term MizarPodController consistently, we need only one const here.

pkg/controller/mizarcontrollers/mizar-pod-controller.go
chenqianfzh: +1

Errors other than NotFound are not handled.

cmd/kube-controller-manager/app/controllermanager.go
```diff
@@ -421,6 +422,8 @@ func NewControllerInitializers(loopMode ControllerLoopMode) map[string]InitFunc
 	controllers["root-ca-cert-publisher"] = startRootCACertPublisher
 	controllers["network"] = startNetworkController
 
+	controllers["mizar-pod"] = startMizarPodController
```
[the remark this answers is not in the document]
Hong-Chang: I'm implementing a mechanism that watching configMap, and start controller after configMap shows it's ready for starting controller.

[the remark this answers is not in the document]
Hong-Chang: I think it's better to hold the review until I make new iteration. I'll close the pr for now.

pkg/controller/mizarcontrollers/mizar-pod-controller.go
```diff
@@ -0,0 +1,194 @@
...
+	coreinformers "k8s.io/client-go/informers/core/v1"
+	clientset "k8s.io/client-go/kubernetes"
+	v1core "k8s.io/client-go/kubernetes/typed/core/v1"
+	corelisters "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/client-go/tools/record"
+	"k8s.io/client-go/util/workqueue"
+	"k8s.io/kubernetes/pkg/controller"
+
+	"k8s.io/klog"
```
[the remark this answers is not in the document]
Hong-Chang: Updated.

pkg/controller/mizarcontrollers/mizar-pod-controller.go
```diff
@@ -0,0 +1,194 @@
...
+	listerSynced cache.InformerSynced
+
+	// To allow injection for testing.
+	handler func(key string, eventType string) error
+
+	// Queue that used to hold thing to be handled.
+	queue workqueue.RateLimitingInterface
+}
+
+// NewPodController configures a new controller instance
```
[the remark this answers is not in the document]
Hong-Chang: Updated.

pkg/controller/mizarcontrollers/mizar-pod-controller.go
```diff
@@ -0,0 +1,194 @@
...
+func NewPodController(informer coreinformers.PodInformer, kubeClient clientset.Interface) *PodController {
+	eventBroadcaster := record.NewBroadcaster()
```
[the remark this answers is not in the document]
Hong-Chang: I see other controllers are using their own eventBroadcaster. Some module has multiple controllers but they still don't share.

pkg/controller/mizarcontrollers/mizar-pod-controller.go
```diff
@@ -0,0 +1,194 @@
...
+	eventBroadcaster.StartLogging(klog.Infof)
+	eventBroadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: kubeClient.CoreV1().EventsWithMultiTenancy(metav1.NamespaceAll, metav1.TenantAll)})
+
+	oc := &PodController{
```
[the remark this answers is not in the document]
Hong-Chang: Changed name from oc to c.

pkg/controller/mizarcontrollers/mizar-pod-controller.go
```diff
@@ -0,0 +1,194 @@
...
+// Run begins watching and handling.
+func (oc *PodController) Run(workers int, stopCh <-chan struct{}) {
+	defer utilruntime.HandleCrash()
+	defer oc.queue.ShutDown()
+
+	klog.Infof("Starting %v controller", controllerFor)
+	defer klog.Infof("Shutting down %v controller", controllerFor)
+
+	if !controller.WaitForCacheSync(controllerKind, stopCh, oc.listerSynced) {
+		return
```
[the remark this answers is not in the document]
Hong-Chang: Inside WaitForCacheSync there is enough log for troubleshooting.

cmd/kube-controller-manager/app/mizarcontrollers.go
Hong-Chang: From other controllers, I see if pod/event happen frequently, the number is higher. For rarely happened controller, the number is 1. Considering pod is core object and its quantity could be huge, I think the number 4 is proper.

I made an update to don't put it in const. Because other mizar controllers are not necessary use 4.

pkg/controller/mizarcontrollers/mizar-pod-controller.go
Hong-Chang: Nice catch. Actually IsNotFound is not special here. Original I reserved it for deletion scenario but actually it doesn't apply. I changed code to return error when it happens. After error returns, it will be logged and retried.

pkg/controller/mizarcontrollers/util.go
Hong-Chang: There are some planned func coming. For example, I need a function to transfer from pod object to pod grpc contract object.

pkg/controller/mizarcontrollers/mizar-pod-controller.go
Hong-Chang: Updated to switch.
"delete" event planed to be handled in future release.

cmd/kube-controller-manager/app/mizarcontrollers.go
Hong-Chang: The pods are all scope. The name "mizar pod" just indicates this controller serves for mizar purpose.

Hong-Chang:
> I have an overall suggestion. Can we fix the name of controller to MizarPodController? I see the term of "mizar-pod-controller" and "podController" are used interchangably.
> 
> Though there is no other controller watching pod now in Arktos, but it is likely there may be another watching-pod controller in the future, since pod is such a fundamental resource.
> 
> I suggest to make changes like renaming the function name from controllers.NewPodController to controllers.NewMizarPodController, and other things like that.

Good point. I did some renaming. Now Mizar and Pod are always together.

pkg/controller/mizarcontrollers/mizar-pod-controller.go
Hong-Chang: Updated.

pkg/controller/mizar/mizar-pod-controller.go
```diff
@@ -0,0 +1,208 @@
...
+
+	// A store of objects, populated by the shared informer passed to MizarPodController
+	lister corelisters.PodLister
+	// listerSynced returns true if the store has been synced at least once.
+	// Added as a member to the struct to allow injection for testing.
+	listerSynced cache.InformerSynced
+
+	// To allow injection for testing.
+	handler func(keyWithEventType KeyWithEventType) error
+
```
clu2xlu, line 51: This part of the code I was not able to run on my end.  System complains unrecognized type  KeyWithEventType

pkg/controller/mizar/mizar-pod-controller.go
```diff
@@ -0,0 +1,208 @@
...
+	informer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc:    c.createObj,
+		UpdateFunc: c.updateObj,
+		DeleteFunc: c.deleteObj,
+	})
+	c.lister = informer.Lister()
+	c.listerSynced = informer.Informer().HasSynced
+
+	c.handler = c.handle
```
clu2xlu, line 80: Could you change the naming here? It's kinda confusing

pkg/controller/mizar/util.go
```diff
@@ -0,0 +1,59 @@
...
+
+type KeyWithEventType struct {
+	EventType string
+	Key       string
+}
+
+type StartHandler func(interface{}, string)
+
+func ConvertToServiceContract(service *v1.Service) *BuiltinsServiceMessage {
+	return &BuiltinsServiceMessage{
```
clu2xlu, line 26: I will move this part to my Service controller

cmd/kube-controller-manager/app/mizarcontrollers.go
```diff
@@ -0,0 +1,54 @@
...
+
+func startMizarPodController(ctx ControllerContext, grpcHost string) (http.Handler, bool, error) {
+	controllerName := "mizar-pod-controller"
+	klog.V(2).Infof("Starting %v", controllerName)
+
+	go controllers.NewMizarPodController(
+		ctx.InformerFactory.Core().V1().Pods(),
+		ctx.ClientBuilder.ClientOrDie(controllerName),
+		grpcHost,
+	).Run(4, ctx.Stop)
```
chenqianfzh, line 52: let's not hard the number of worker as  4 here. Give it a variable, we may need to tune this in the future.

h-w-chen (commented):
all of my comments had been addressed; the skeletal code of mizar-pod controller looks fine to me.

Can this PR refrain from doing more than skeletal controller? The grpc related code is better in its separate PR, and team members who are familiar with grpc can help to provide meaningful reviews.

[Hong] I agree with you. Our pr should be small and precise. We should split our task to small pieces and independent. 
In this case, it's not ideal because in the final stage of the release I jumped in the tasks and make them done. These tasks should have been done one by one but now I have to work on them all at once. The later task depend on prev task. And since I do them together (in the weekend of labor day), I have to send only one pr. I guess we don't have time to split them to 3 pr and review one by one.
I will make sure every part will be reviewed by you or Chen Qian.

pkg/controller/mizar/grpc-adaptor.go
```diff
@@ -0,0 +1,272 @@
...
+	"google.golang.org/grpc"
+
+	v1 "k8s.io/api/core/v1"
+)
+
+const (
+	port = "50052"
+)
+
+// GrpcCreateService is to invoking grpc func of CreateService
```
chenqianfzh, line 30: Guess we can have simpler code by abstrction using func as a param. For example, for those Service related functions:

func GrpcServiceOperation(grpcHost string, service *v1.Service, operation func(context.Context, *BuiltinsServiceMessage) (*ReturnCode, erroRr) ) *ReturnCode {
	client, ctx, conn, 
        cancel, err := getGrpcClient(grpcHost)
	if err != nil {
		return getReturnCodeFromError(&err)
	}
	defer conn.Close()
	defer cancel()
	returnCode, err := operation(ctx, ConvertToServiceContract(service))
	if err != nil {
		return getReturnCodeFromError(&err)
	}
	return returnCode
}

func GrpcCreateService(grpcHost string, service *v1.Service) *ReturnCode {
       return GrpcServiceOperation(grpcHost, service, client.CreateService)
}

...

chenqianfzh: Actually, more abstraction can be done over the different types of resources. Yet it involves more type conversion and check, I had a brief look and it seems to me that the signature of those conversion funcs need to be updated to handler the possible conversion errors.

So I just stop here, you can play with it more to have even simpler code.

pkg/controller/mizar/grpc-adaptor.go
```diff
@@ -0,0 +1,272 @@
...
+func getReturnCodeFromError(err *error) *ReturnCode {
+	return &ReturnCode{
+		Code:    CodeType_TEMP_ERROR,
+		Message: fmt.Sprintf("Grpc call failed: %s", (*err).Error()),
+	}
+}
+
+func getGrpcClient(grpcHost string) (BuiltinsServiceClient, context.Context, *grpc.ClientConn, context.CancelFunc, error) {
+	address := fmt.Sprintf("%s:%s", grpcHost, port)
+	conn, err := grpc.Dial(address, grpc.WithInsecure(), grpc.WithBlock())
```
chenqianfzh, line 264: Why insecure?

pkg/controller/mizar/grpc-adaptor.go
```diff
@@ -0,0 +1,272 @@
...
+	returnCode, err := client.DeleteNode(ctx, ConvertToNodeContract(node))
+	if err != nil {
+		return getReturnCodeFromError(&err)
+	}
+	return returnCode
+}
+
+func getReturnCodeFromError(err *error) *ReturnCode {
+	return &ReturnCode{
+		Code:    CodeType_TEMP_ERROR,
```
chenqianfzh, line 257: I think HTTP Internal Error is more suitable in this case, which seems the usual behavior in K8s.

chenqianfzh: One thing more:  getReturnCodeFromError and getGrpcClient are very likely to be used a lot by other team members. Consider change them to public functions

pkg/controller/mizar/grpc-adaptor.go
```diff
@@ -0,0 +1,272 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package mizar
+
+import (
+	"context"
+	fmt "fmt"
```
chenqianfzh, line 18: fmt "fmt" -> "fmt"

pkg/controller/mizar/grpc-adaptor.go
```diff
@@ -0,0 +1,272 @@
...
+	"time"
+
+	"google.golang.org/grpc"
+
+	v1 "k8s.io/api/core/v1"
```
chenqianfzh, line 23: v1 "k8s.io/api/core/v1" ->  "k8s.io/api/core/v1"

pkg/controller/mizar/mizar-pod-controller.go
chenqianfzh: +1, a common func in this case is sync(). Does it fit?

pkg/controller/mizar/mizar-pod-controller.go
```diff
@@ -0,0 +1,208 @@
...
+		// Two different versions of the same object will always have different RVs.
+		return
+	}
+	if curObj.DeletionTimestamp != nil {
+		// Object is being deleted. Don't handle update anymore.
+		return
+	}
+
+	key, _ := controller.KeyFunc(curObj)
+	c.queue.Add(KeyWithEventType{Key: key, EventType: "update"})
```
chenqianfzh, line 124: Nit: make a new type, an enum of strings, for EventType,

pkg/controller/mizar/mizar-pod-controller.go
```diff
@@ -0,0 +1,208 @@
...
+	return nil
+}
+
+func processGrpcReturnCode(c *MizarPodController, returnCode *ReturnCode, keyWithEventType KeyWithEventType) {
+	key := keyWithEventType.Key
+	switch returnCode.Code {
+	case CodeType_OK:
+		klog.Infof("Mizar handled request successfully for %v. key %s, eventType %s", controllerForMizarPod, key, keyWithEventType.EventType)
+	case CodeType_TEMP_ERROR:
+		klog.Infof("Mizar hit temporary error for %v. key %s. %s, eventType %s", controllerForMizarPod, key, returnCode.Message, keyWithEventType.EventType)
```
chenqianfzh, line 205: Not sure whether it is OK to log the error as "temporary". When we say "temporary", it is usually an error that will go after retrial.

pkg/controller/mizar/mizar-starter-controller.go
```diff
@@ -0,0 +1,165 @@
...
+	}
+
+	obj, err := c.lister.ConfigMapsWithMultiTenancy(namespace, tenant).Get(name)
+	if err != nil {
+		return err
+	}
+
+	klog.V(4).Infof("Handling %v %s/%s/%s hashkey %v", controllerForMizarStarter, tenant, namespace, obj.Name, obj.HashKey)
+
+	if key == "default/mizar-grpc-service" {
```
chenqianfzh, line 160: Don't check the key directly. Instead check the value of tenant/ns/name. The key format may be changed in the future.

cmd/kube-controller-manager/app/mizarcontrollers.go
```diff
@@ -0,0 +1,54 @@
...
+		ctx.InformerFactory.Core().V1().ConfigMaps(),
+		ctx.ClientBuilder.ClientOrDie(controllerName),
+		ctx,
+		startHandler,
+	).Run(2, ctx.Stop)
+	return nil, true, nil
+}
+
+func startHandler(controllerContext interface{}, grpcHost string) {
+	ctx := controllerContext.(ControllerContext)
```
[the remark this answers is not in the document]
Hong-Chang: It won't. Here in line 39 I have to declare controllerContext as interface{} instead of type ControllerContext. It's because the context will be pass into start controller then pass back when starting other controllers. Inside start controller I cannot import ControllerContext's package to avoid circle import.

pkg/controller/mizarcontrollers/mizar-pod-controller.go
Hong-Chang: Now delete event is being handled.

cmd/kube-controller-manager/app/mizarcontrollers.go
Hong-Chang: Updated. Added const.

pkg/controller/mizar/grpc-adaptor.go
Hong-Chang: Updated.

h-w-chen (approved):
approved based on arktos skeletal controller code.

Sindica (changes requested):
I am not seeing any unit test or integration test code.

pkg/controller/mizar/mizar-starter-controller.go
```diff
@@ -157,7 +157,7 @@ func (c *MizarStarterController) handle(key string) error {
 
 	klog.V(4).Infof("Handling %v %s/%s/%s hashkey %v", controllerForMizarStarter, tenant, namespace, obj.Name, obj.HashKey)
 
-	if key == "default/mizar-grpc-service" {
+	if namespace == "default" && name == "mizar-grpc-service" {
```
chenqianfzh, line 160: minor: "default" -> metav1.NamespaceDefault

chenqianfzh (approved):
LGTM.

Hong, can you update the thread that we synced offline? so we know they are addressed.

Hong-Chang (commented):
> I am not seeing any unit test or integration test code.

I plan to write test code after this merged. Our release date is 9/15, and I originally didn't join the release. Later when I see there is risk on schedule so I joined the release to work on these code, and these code is like structural code that is dependent by other team members. So I need to code to be merged first to unblock other people. Then I'll write test code for it.

Sindica:
> I am not seeing any unit test or integration test code.

Author got offline approved from Dr. Xiong.

Hong-Chang:
> LGTM.
> 
> Hong, can you update the thread that we synced offline? so we know they are addressed.

Yes, we majorly discussed the refactoring of grpc-adaptor. Because currently it's not decided yet how other controller will use the grpc-adaptor, I will hold the refactoring until the code is accepted by other members. Qian agrees.

Sindica (approved):

PR #680, opened 2020-09-04
Revert "kubelet prepares pod DNS config with network specific DNS service IP"

h-w-chen:
Reverts futurewei-cloud/arktos#449

kubemark test is having problem that keubelet cannot be started due to changes in this PR. reverting it to unblock scalability testing.

Will investigate the regression ASAP.


Sindica (approved):

h-w-chen:
there is a patch to address the identified panic of hollow-node; if it is verified, we don't have to revert it. we'll see the result of patch verification.

PR #683, opened 2020-09-04
arktos network controller deploy dns per network

h-w-chen:
**What type of PR is this?**
/kind feature

This PR renames flat-network-controller to arktos-network-controller, and extends to cope with other types of networks (not flat networks only).

YES
```release-note
whenever new custom network object is created, the dns service object of the network will be created, and dns deployment of that network will be triggered, too.
```


h-w-chen:
seems having to manually resolve conflicts. closing for now.

PR #688, opened 2020-09-05
arktos network controller deploy dns per network

h-w-chen:
**What type of PR is this?**
/kind feature

This PR renames flat-network-controller to arktos-network-controller, and extends to cope with other types of networks (not flat networks only).

YES
```release-note
whenever new custom network object is created, the dns service object of the network will be created, and dns deployment of that network will be triggered, too.
```


cmd/arktos-network-controller/app/controller.go
```diff
@@ -158,6 +158,33 @@ func (c *Controller) process(item interface{}) {
 	c.queue.Forget(item)
 }
 
+// manageNonFlatNetwork is the core logic to manage a non-flat typed network object
+func manageNonFlatNetwork(net *v1.Network, netClient arktos.Interface, svcClient kubernetes.Interface, toDeployDNS bool, domainName string) error {
+	var svc *corev1.Service
+	var err error
+	if len(net.Status.DNSServiceIP) == 0 {
```
chenqianfzh, line 165: I think there is a logic error. If len(net.Status.DNSServiceIP) >0, line 166 will not be run, so svc will be nil, then 177 will hit nil pointer error.

It seems to me the logic here should be the same as line 163-177 in func manageFlatNetwork(). It that is the case, let's have a small helper func for those code.

cmd/arktos-network-controller/app/controller.go
```diff
@@ -158,6 +158,33 @@ func (c *Controller) process(item interface{}) {
...
+		}
+	}
+
+	if toDeployDNS {
+		if err = deployDNSForNetwork(net, svcClient, domainName); err != nil {
+			return fmt.Errorf("failed to deploy per-network DNS: %v", err)
+		}
+	}
+
+	if len(svc.Spec.ClusterIP) > 0 {
```
chenqianfzh, line 177: The DNS service is of type ClusterIP, so if len(svc.Spec.ClusterIP) > 0, we need to return an error.

chenqianfzh:
A general comment: A URL to the multi-tenancy networking design doc in the beginning of controller.go may help the reader.

cmd/arktos-network-controller/app/controller.go
h-w-chen: talked offline; will add a comment to explain the reason why cluster ip might be empty, to avoid confusion.

cmd/arktos-network-controller/app/controller.go
h-w-chen: good point; need to protect from nil svc object.
for the code reuse, as the logic seems quite different from the flat typed network, it is fine to have separate code body. Having said that; if in the future the code structure is changed and code reuse makes more sense, we should refactor code accordingly.

chenqianfzh (approved):
lgtm

PR #704, opened 2020-09-11
Wait until pod get hostIP filled

Hong-Chang:
When pod is just created, the HostIP value hasn't been fill into the object.
Wait for at most 2 seconds to get the pod with HostIP value.

**What type of PR is this?**
/kind bug

For newly created pod, there is empty HostIP sent to grpc server


chenqianfzh (changes requested):
I am concerned about the controller behavior when something goes wrong with the HostIP assignment in Miza. When it happens, every pod processing will cost 2 seconds. Suppose there are 4 workers, it means only 120 pods can be processed per minute, namely the controller is blocked.

Hong-Chang:
Found better way to handle issue

PR #705, opened 2020-09-11
Fix issue that no hostIP sent to grpc server

Hong-Chang:
Currently there is no hostIP sent to grpc server for newly created pod.
When a pod got created, there will be 1 create event triggered and 5 update events triggered. At the time the create event and first update event happens, there is no hostIP yet. And at this time mizar pod controller is sending the data to grpc server without hostIP data. The followed 4 update events have hostIP data, but in the event queue, the update event is being handled and existing in the queue, then the 4 followed update events are dedup'd by queue, which means the 4 followed update events cannot enqueue. Hence when first update event handle completed, there is no more event to be handled in the queue, so eventually there is no hostIP data sent to grpc server.
The solution is to add an EventTime field in the queue data which can distinguish all the different update events. In this way all the update events will be handled and sent to grpc server. Even the first event has no hostIP data, the following events have hostIP data and grpc server will get hostIP. 

What type of PR is this?
/kind bug

Which issue(s) this PR fixes:
For newly created pod, there is empty HostIP sent to grpc server

### DIFF
--- a/pkg/controller/mizar/mizar-pod-controller.go
+++ b/pkg/controller/mizar/mizar-pod-controller.go
@@ -103,7 +103,7 @@ func (c *MizarPodController) Run(workers int, stopCh <-chan struct{}) {
 
 func (c *MizarPodController) createObj(obj interface{}) {
 	key, _ := controller.KeyFunc(obj)
-	c.queue.Add(KeyWithEventType{Key: key, EventType: EventType_Create})
+	c.queue.Add(KeyWithEventType{Key: key, EventType: EventType_Create, EventTime: time.Now().Format(time.RFC3339Nano)})
 }
 
 // When an object is updated.
@@ -121,13 +121,13 @@ func (c *MizarPodController) updateObj(old, cur interface{}) {
 	}
 
 	key, _ := controller.KeyFunc(curObj)
-	c.queue.Add(KeyWithEventType{Key: key, EventType: EventType_Update})
+	c.queue.Add(KeyWithEventType{Key: key, EventType: EventType_Update, EventTime: time.Now().Format(time.RFC3339Nano)})
 }
 
 func (c *MizarPodController) deleteObj(obj interface{}) {
 	key, _ := controller.KeyFunc(obj)
 	klog.Infof("%v deleted. key %s.", controllerForMizarPod, key)
-	c.queue.Add(KeyWithEventType{Key: key, EventType: EventType_Delete})
+	c.queue.Add(KeyWithEventType{Key: key, EventType: EventType_Delete, EventTime: time.Now().Format(time.RFC3339Nano)})
 }
 
 // worker runs a worker thread that just dequeues items, processes them, and marks them done.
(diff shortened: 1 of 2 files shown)


pkg/controller/mizar/mizar-pod-controller.go
```diff
@@ -121,13 +121,18 @@ func (c *MizarPodController) updateObj(old, cur interface{}) {
...
 
 func (c *MizarPodController) deleteObj(obj interface{}) {
 	key, _ := controller.KeyFunc(obj)
 	klog.Infof("%v deleted. key %s.", controllerForMizarPod, key)
-	c.queue.Add(KeyWithEventType{Key: key, EventType: EventType_Delete})
+	c.queue.Add(KeyWithEventType{Key: key, EventType: EventType_Delete, ResourceVersion: getResourceVersion(obj)})
+}
+
+func getResourceVersion(obj interface{}) string {
+	pod := obj.(*v1.Pod)
```
chenqianfzh, line 134: Let's re-use the apimachinery util func to do the job:
```
import "k8s.io/apimachinery/pkg/api/meta"

func GetResourceVersion(obj interface{}) string {
     objAccessor, _ := meta.Accessor(obj)
    return objAccessor.GetResourceVersion()
}
```

pkg/controller/mizar/mizar-pod-controller.go
```diff
@@ -121,13 +121,18 @@ func (c *MizarPodController) updateObj(old, cur interface{}) {
...
 
 	key, _ := controller.KeyFunc(curObj)
-	c.queue.Add(KeyWithEventType{Key: key, EventType: EventType_Update})
+	c.queue.Add(KeyWithEventType{Key: key, EventType: EventType_Update, ResourceVersion: curObj.ResourceVersion})
 }
 
 func (c *MizarPodController) deleteObj(obj interface{}) {
 	key, _ := controller.KeyFunc(obj)
 	klog.Infof("%v deleted. key %s.", controllerForMizarPod, key)
-	c.queue.Add(KeyWithEventType{Key: key, EventType: EventType_Delete})
```
chenqianfzh, line 130: Do we need a resourceVersion here as there is only one delete event for each pod? 

If that is the case, we can just put an empty string as the ResourceVersion.

chenqianfzh (approved):
I believe the PR should unblock the mizar integration issue.

Leaving some comments about coding.

pkg/controller/mizar/mizar-pod-controller.go
Hong-Chang: Yes, I think empty string is better. Updated.

pkg/controller/mizar/mizar-pod-controller.go
Hong-Chang: Updated. Thanks for writing the code.

pkg/controller/mizar/mizar-pod-controller.go
```diff
@@ -103,7 +103,7 @@ func (c *MizarPodController) Run(workers int, stopCh <-chan struct{}) {
 
 func (c *MizarPodController) createObj(obj interface{}) {
 	key, _ := controller.KeyFunc(obj)
-	c.queue.Add(KeyWithEventType{Key: key, EventType: EventType_Create})
+	c.queue.Add(KeyWithEventType{Key: key, EventType: EventType_Create, ResourceVersion: getResourceVersion(obj)})
```
chenqianfzh, line 106: Do we need a resourceVersion here as there is only one create event for each pod?

If that is the case, we can just put an empty string as the ResourceVersion.

Hong-Chang: Updated

Issue #698, opened 2020-09-09
Panic: "invalid memory address or nil pointer dereference" in daemon_controller.go


sonyafenge:
**What happened**:
run perf-tests with 5K nodes and disable workload-controller-manager, get the panic:
```
E0909 14:14:34.420852       1 runtime.go:73] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 12740 [running]:
k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/util/runtime.logPanic(0x378f120, 0x7207f30)
        /go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:69 +0x7b
k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
        /go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/vendor/k8s.io/apimachinery/pkg/util/runtime/runtime.go:51 +0x82
panic(0x378f120, 0x7207f30)
        /usr/local/go/src/runtime/panic.go:522 +0x1b5
k8s.io/kubernetes/vendor/k8s.io/api/core/v1.(*PodSpec).Workloads(0xc02589fd38, 0x8, 0x989680, 0xc0391b8a20)
        /go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/vendor/k8s.io/api/core/v1/types.go:3286 +0x89
k8s.io/kubernetes/pkg/scheduler/nodeinfo.calculateResource(0xc02589fc00, 0x0, 0x0, 0x0, 0x0, 0x0, 0x8, 0x1)
        /go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/pkg/scheduler/nodeinfo/node_info.go:583 +0x6a
k8s.io/kubernetes/pkg/scheduler/nodeinfo.(*NodeInfo).AddPod(0xc0390e5520, 0xc02589fc00)
        /go/src/k8s.io/kubernetes/_output/dockerized/go/src/k8s.io/kubernetes/pkg/scheduler/nodeinfo/node_info.go:502 +0x40
[...]

PR #706, opened 2020-09-11
Initialize WorkloadInfo during defaulting

vinaykul:
**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**: There is a nil ptr or invalid memory deref issue when populating WorkloadInfo in PodSpec. This is caused either by race condition during deserialization, or due to memory corruption in daemon controller. Setting WorkloadInfo during defaulting allows it to be carried with the PodSpec all the time, while continuing to allow latest values of mutable fields to be accessed.

**Which issue(s) this PR fixes**: #698 

Fixes #

```release-note

```


yb01:
having a few thoughts on this PR:
1. is it really need to be guarded with the feature gate? understood it intends to minimize the impact to 930 should this carries regressions, but a feature gate seems to be not fit in this purpose.
2. suggest to do a bit future just to set the workload info, i.e.
   a.  update the workloads() function in the type.go, treat the workloads() as a getter and the setWorkloads() function purely a setter, to initialize the slice. update the two functions to reflect the getter and setter nature 
   b. workloads() only do get or update the slice. and assert if the slice is not set
   c.  setWorkloads() can have a lock ( logically not necessary ) in case it was called in other places.

vinaykul:
> having a few thoughts on this PR:
> 
> 1. is it really need to be guarded with the feature gate? understood it intends to minimize the impact to 930 should this carries regressions, but a feature gate seems to be not fit in this purpose.
> 2. suggest to do a bit future just to set the workload info, i.e.
>    a.  update the workloads() function in the type.go, treat the workloads() as a getter and the setWorkloads() function purely a setter, to initialize the slice. update the two functions to reflect the getter and setter nature
>    b. workloads() only do get or update the slice. and assert if the slice is not set
>    c.  setWorkloads() can have a lock ( logically not necessary ) in case it was called in other places.

This change is unique because WorkloadInfo is a cache of Container/VM fields. Featuregate + enable in arktos-up / kube-up is a low-cost way to enable its initialization and avoid special-casing the protobuf, round-tripping, and all the other compatibility tests that are failing because the tests fill the non-nil fields with random data and expect to see that random data instead of the default-initialized data.

pkg/apis/core/v1/defaults.go
```diff
@@ -203,6 +205,11 @@ func SetDefaults_PodSpec(obj *v1.PodSpec) {
 	if obj.SchedulerName == "" {
 		obj.SchedulerName = v1.DefaultSchedulerName
 	}
+	if utilfeature.DefaultFeatureGate.Enabled(features.WorkloadInfoDefaulting) {
```
yb01, line 208: Continue on the discussion if the feature gate is needed as in the PR conversation here:
my understanding for fixing issue #698 , this feature-gate will need to be set true anyways, so from that perspective, this feature gate ( at least per the code here ) is not much useful.  if there are other code path in Arktos that will need this flag to be false, i think we should consider to fix that code path, vm and container support as the unified platform is essential in Arktos, i.e. from that perspective, it is not just a "feature", right ?

pkg/apis/core/v1/defaults.go
```diff
@@ -203,6 +205,11 @@ func SetDefaults_PodSpec(obj *v1.PodSpec) {
...
+		if len(obj.WorkloadInfo) == 0 {
+			_ = obj.Workloads()
```
yb01, line 210: in the admission controller for vertical scaling you added a few weeks ago, there is already a setter level function in the types.go file for this purpose,  suggested to make this consistent, i.e. for workloadinfo slice life cycle: init(), set() in the admission controller, and get()afterward in other Arktos code except the admission controller ( when the pod is first time created in the system ). likely a more readable and maintainable code as such.

yb01 (commented):
Regarding the two suggestions, let discuss them separately in the code review level.

yb01 (approved):
OK for 930 to unblock the load testing. a couple suggestion might need more code refactor work around the workloads() and worklaodinfo slice, they can be done post 930.

pkg/apis/core/v1/defaults.go
vinaykul: Yes. I'm setting it to true in arktos-up and kube-up to cover our deployment use cases. This is apiserver defaulting code, and there is no other place where this is required. Once we initialize this, the deserialization of PodSpec should have WorkloadInfo allocated and set. If we were to do this without featuregate, it will require investigating and changing to the test infra to consider this special-case, and fixing all the failing unit tests.

pkg/apis/core/v1/defaults.go
vinaykul: This is not set. We are calling get here to allocate and initialize WorkloadInfo for first time as defaulting.

PR #707, opened 2020-09-12
adds feature gate MandatoryArktosNetwork to enforce network exists for each non-host pod 

h-w-chen:
**What type of PR is this?**
/kind feature

This PR adds an optional feature gate, MandatoryArktosNetwork, for the mandatory check of network object of every non-host pod.

When this gate is explicitly turn on, pods on the implicit default network won't be started - until the default network is created.

By defaul, this gate if off; pods on the implicit network could still use the cluster global DNS as their name server, so the backward compatability of k8s is maintained.

**How to verify**
1. explictly set env var
```
export FEATURE_GATES="AllAlpha=false,MandatoryArktosNetwork=true"
```
2. start arktos lab env by ```./hack/arktos-up.sh```
3. start a pod (implicitly associated to default network), by ```./cluster/kubectl.sh run nnn image=nginx```
4. observe pod unable to run
5. create the default network, and sets its status to Ready with a valid dnsServiceIP, using follwoing yaml (may need to manually update the status, if arktos-network-controller is absent)
```yaml
kind: Network
metadata:
  name: default
spec:
  type: flat
```
6. observe the pod is in Running state
```
./cluster/kubectl.sh get pod
NAME                   HASHKEY               READY   STATUS    RESTARTS   AGE
nnn-757db88665-jnc76   7420869659319877687   1/1     Running   0          34m
```
Yes, when the --feature-gates=MandatoryArktosNetwork=true is one of args of kubelet
```release-note
If kubelet starts with --feature-gates=MandatoryArktosNetwork=true, every non-host pod won't start unless its associated network object already exists.
```


pkg/kubelet/network/dns/dns.go
```diff
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	"k8s.io/kubernetes/pkg/features"
```
chenqianfzh, line 24: nit: this package belongs to the k8s group, should be moved down.

h-w-chen: will reorgnize imports

chenqianfzh (approved):
Looks good.

PR #708, opened 2020-09-12
[Mizar-Endpoints-Controller] Implemented 

kimeunju108:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it:**

This PR presents an endpoints controller for integrating Arktos and Mizar systems.

-Endpoints: built-in kubernetes object which is the list of addresses (ip and port) of endpoints of a Service.

-Major functions of controllers (currently partially implemented and will be updated soon)
(1) Controller detect endpoints' events such as add, update, delete, and resume.
(2) Controllers send the events' information such as service name, the list of endpoints' ip addresses and ports to Mizar operator using gRPC protocols

-Planned, but not yet completed
(1) Integration testing with other modules such as mizar operator proxy
(2) Implementing test cases for this controller


pkg/controller/mizar/BUILD
```diff
@@ -11,6 +11,10 @@ go_library(
         "builtins.pb.go",
         "grpc-adaptor.go",
         "mizar-pod-controller.go",
+<<<<<<< HEAD
+        "mizar-endpoints-controller.go",
+=======
+>>>>>>> 70f10144a36d9de4e9f45373f5d4ffb2d36743fd
```
phudtran, line 17: Leftover stuff from rebase/merge that needs to be cleaned up.

PR #710, opened 2020-09-12
mizar grpc proto schema change

Hong-Chang:
This is mizar grpc proto change. Server side needs network instead of vpc

**What type of PR is this?**
/kind feature


chenqianfzh (approved):
It looks good to me, just make sure the format meets the needs in the Mizar side.

PR #727, opened 2020-09-15
mizar node/endpoints controller

Hong-Chang:
**What type of PR is this?**
/kind feature


chenqianfzh:
I think this PR is not rebased to the latest Arktos master branch. For instance, pkg/controller/mizar/mizar-service-controller.go is already checked in commit https://github.com/futurewei-cloud/arktos/commit/ce15e44c110280ffc78c406cd8524ca47acceff5, yet it is still shown as a new file in this PR.

Could you rebase and send out the PR again?

chenqianfzh (changes requested):
I think this PR is not rebased to the latest Arktos master branch. For instance, pkg/controller/mizar/mizar-service-controller.go is already checked in commit ce15e44, yet it is still shown as a new file in this PR.

Could you rebase and send out the PR again?

PR #728, opened 2020-09-16
arktos-up: install network crd before waiting for node ready

h-w-chen:
**What type of PR is this?**
/kind bug

Arktos-up.sh installs network CRD before entering the loop of waiting for node be ready

NONE


chenqianfzh (approved):
looks good.

PR #729, opened 2020-09-16
Adding retry code for network and service controllers

clu2xlu:
**What type of PR is this?**
/kind feature

Controller will retry if a TEMP_ERROR return code is send from mizar side

no
```release-note

```

### DIFF
--- a/cmd/kube-controller-manager/app/mizarcontrollers.go
+++ b/cmd/kube-controller-manager/app/mizarcontrollers.go
@@ -128,28 +128,27 @@ func startMizarServiceController(ctx *ControllerContext, grpcHost string) (http.
 
 func startArktosNetworkController(ctx *ControllerContext, grpcHost string) (http.Handler, bool, error) {
 	controllerName := "mizar-arktos-network-controller"
-        klog.V(2).Infof("Starting %v", controllerName)
+	klog.V(2).Infof("Starting %v", controllerName)
 
-        netKubeconfigs := ctx.ClientBuilder.ConfigOrDie(controllerName)
-        for _, netKubeconfig := range netKubeconfigs.GetAllConfigs() {
-                netKubeconfig.QPS *= 5
-                netKubeconfig.Burst *= 10
-        }
-        crConfigs := *netKubeconfigs
-        for _, cfg := range crConfigs.GetAllConfigs() {
-                cfg.ContentType = "application/json"
-                cfg.AcceptContentTypes = "application/json"
-        }
-        networkClient := arktos.NewForConfigOrDie(&crConfigs)
-        svcKubeClient := clientset.NewForConfigOrDie(netKubeconfigs)
-        informerFactory := externalversions.NewSharedInformerFactory(networkClient, 0)
+	netKubeconfigs := ctx.ClientBuilder.ConfigOrDie(controllerName)
+	for _, netKubeconfig := range netKubeconfigs.GetAllConfigs() {
+		netKubeconfig.QPS *= 5
+		netKubeconfig.Burst *= 10
+	}
+	crConfigs := *netKubeconfigs
+	for _, cfg := range crConfigs.GetAllConfigs() {
+		cfg.ContentType = "application/json"
+		cfg.AcceptContentTypes = "application/json"
+	}
+	networkClient := arktos.NewForConfigOrDie(&crConfigs)
+	svcKubeClient := clientset.NewForConfigOrDie(netKubeconfigs)
+	informerFactory := externalversions.NewSharedInformerFactory(networkClient, 0)
 
-        go controllers.NewMizarArktosNetworkController(
-                networkClient,
-                svcKubeClient,
-                informerFactory.Arktos().V1().Networks(),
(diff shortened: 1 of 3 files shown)


pkg/controller/mizar/mizar-service-controller.go
```diff
@@ -264,9 +319,16 @@ func (c *MizarServiceController) processServiceDeletion(key string) error {
...
 	code := response.Code
-	if code != CodeType_OK {
-		klog.Errorf("Return Code: %v", code)
-		return errors.New("Service Deletion failed on mizar side")
+	switch code {
+	case CodeType_OK:
+		klog.Infof("Mizar handled service deletion successfully: %s", key)
+	case CodeType_TEMP_ERROR:
+		klog.Infof("Mizar hit temporary error for service deletion for service: %s", key)
+		c.queue.AddRateLimited(eventKey)
```
chenqianfzh, line 327: Does it work? The processNextWorkItem() expects the item got from the queue is of type KeyWithEventType. Yet you add an object of type string to the queue here. 

```
func (c *MizarServiceController) processNextWorkItem() bool {
	workItem, quit := c.queue.Get()
	...
       eventKey := workItem.(KeyWithEventType)
```

Seems this issue exist in all the c.queue.AddRateLimited() in this PR.

chenqianfzh (changes requested):
The type of object added to the queue for retrial does not look right. I am afraid the re-trial does not work as expected.

cmd/kube-controller-manager/app/mizarcontrollers.go
```diff
@@ -73,11 +73,26 @@ func startMizarServiceController(ctx *ControllerContext, grpcHost string) (http.
 	klog.V(2).Infof("Starting %v", controllerName)
 
 	svcKubeconfigs := ctx.ClientBuilder.ConfigOrDie(controllerName)
+	for _, svcKubeconfig := range svcKubeconfigs.GetAllConfigs() {
+		svcKubeconfig.QPS *= 20
```
Sindica, line 77: What is the original value? *20 seems too aggressive

cmd/kube-controller-manager/app/mizarcontrollers.go
```diff
@@ -73,11 +73,26 @@ func startMizarServiceController(ctx *ControllerContext, grpcHost string) (http.
...
 	svcKubeClient := clientset.NewForConfigOrDie(svcKubeconfigs)
 
+	crConfigs := *svcKubeconfigs
+	for _, cfg := range crConfigs.GetAllConfigs() {
+		cfg.ContentType = "application/json"
+		cfg.AcceptContentTypes = "application/json"
+	}
+	networkClient := arktos.NewForConfigOrDie(&crConfigs)
+
+	informerFactory := externalversions.NewSharedInformerFactory(networkClient, 10*time.Minute)
```
Sindica, line 89: The 2nd arg means the interval to rescan all events again. Please make sure this is what you wanted.

pkg/controller/mizar/mizar-arktos-network-controller.go
```diff
@@ -190,9 +191,18 @@ func (c *MizarArktosNetworkController) processNetworkCreation(network *v1.Networ
 	code := response.Code
 	context := response.Message
 
-	if code != CodeType_OK {
-		return errors.New("Fail to update VPC with Arktos Network name on the Mizar cluster")
+	switch code {
+	case CodeType_OK:
+		klog.Infof("Mizar handled arktos network creation successfully: %s", key)
+	case CodeType_TEMP_ERROR:
+		klog.Infof("Mizar hit temporary error for arktos network creation: %s", key)
```
chenqianfzh, line 198: Infof -> Warningf

pkg/controller/mizar/mizar-arktos-network-controller.go
```diff
@@ -190,9 +191,18 @@ func (c *MizarArktosNetworkController) processNetworkCreation(network *v1.Networ
 	code := response.Code
 	context := response.Message
 
-	if code != CodeType_OK {
-		return errors.New("Fail to update VPC with Arktos Network name on the Mizar cluster")
+	switch code {
+	case CodeType_OK:
+		klog.Infof("Mizar handled arktos network creation successfully: %s", key)
```
chenqianfzh, line 196: log: only create events will hit the code in line 196

clu2xlu: During arktos network creation, on mizar side, we need associate akrtos network obj name with VPC ID. 
I agree the log is not very clear, I will update it

chenqianfzh (approved):
Overall it is OK to me, leaving some minor comments.

cmd/kube-controller-manager/app/mizarcontrollers.go
XiaoningDing: We do this multiplication for namespace controller and tenant controller because they are very chatty during nested deletion scenario.  I don't think we need to multiply it for regular object controllers.

PR #734, opened 2020-09-16
fix kubectl-get-net error

chenqianfzh:
This PR fixes the issue that "kubectl get net" return error complaining the "net" is not a valid resource type, though CRD networks.arktos.futurewei.com has already been installed in the arktos-up.sh.

Analysis & Solution:
The bug is due to that resource discovery cache is not updated after the new CRD type is created. So this script adds a line of "kubectl api-resources" to force a cache refresh.

Verification:
In the local dev box, right after the Arktos cluster is up by arktos-up.up, run the following command:
```
qianchen@qianchen-VirtualBox:~/gopath/qianc/arktos$ kubectl get net
No resources found.
```

The error message before this change was gone.

### DIFF
--- a/hack/arktos-up.sh
+++ b/hack/arktos-up.sh
@@ -523,7 +523,8 @@ if [[ "${DEFAULT_STORAGE_CLASS}" = "true" ]]; then
 fi
 
 ${KUBECTL} --kubeconfig="${CERT_DIR}/admin.kubeconfig" apply -f "${KUBE_ROOT}/pkg/controller/artifacts/crd-network.yaml"
-
+# refresh the resource discovery cache after the CRD is created
+${KUBECTL} --kubeconfig="${CERT_DIR}/admin.kubeconfig" api-resources &>/dev/null
 echo "*******************************************"
 echo "Setup Arktos components ..."
 echo ""


h-w-chen (approved):
as a workaround, this approach is quite progamatic.
I have no problem with this approach.

hack/arktos-up.sh
```diff
@@ -523,7 +523,8 @@ if [[ "${DEFAULT_STORAGE_CLASS}" = "true" ]]; then
 fi
 
 ${KUBECTL} --kubeconfig="${CERT_DIR}/admin.kubeconfig" apply -f "${KUBE_ROOT}/pkg/controller/artifacts/crd-network.yaml"
-
+# refresh the resource discovery cache after the CRD is created
+${KUBECTL} --kubeconfig="${CERT_DIR}/admin.kubeconfig" api-resources &>/dev/null
```
Sindica, line 527: It only works in arktos-up, correct? What about cluster installed by other means such as kubadm or GCE cluster in perf test?

chenqianfzh: The change "${KUBECTL} --kubeconfig="${CERT_DIR}/admin.kubeconfig" apply -f "${KUBE_ROOT}/pkg/controller/artifacts/crd-network.yaml" does NOT happen in kubadm or GCE cluster, so this change is not needed.

Sindica: I am confused. Are you saying those will not be supported in automatically set up cluster or those are not currently supported? It looks line 525 and 527 should be used together to support crd network. Isn't it? If so, I would suggest let user to do both kubectl action so that they understand it better. Otherwise, Arktos is giving wrong impression of how those are used.

chenqianfzh: synced offline. This action only happens in dev cluster only so far. Not ready for generic clusters yet.

PR #744, opened 2020-09-18
coredns pod connects to kube-apiserver directly

### LINKED ISSUE #742: core-dns pod in customer vpc failing health checks
**What happened**:
We noticed an issue with customer vpc (vpc1) network: 

In a customer vpc, the core-dns pod failed to start due to failing the health check, see log below: 

As you can see from the log, the expected behavior should be: vpc1 core-dns pod should ping kubernetes-my-net-mizar-vpc1 ip address 12.2.0.2
However, vpc1 core-dns pod is pinging 10.0.0.1, which is default/Kubernetes ‘s ip address. 

the expected behavior should be: vpc1 core-dns pod should ping kubernetes-my-net-mizar-vpc1 ip address 12.2.0.2

1. Create a custom vpc (called vpc1) and subnet
2. Create a arktos network that is associated with that customer vpc
3. Check the dns services
4. kubectl get pods -AT
    where it shows core-dns pods
5. kubectl --namespace kube-system logs coredns-my-net-mizar-vpc1
6. you will see health check failure but the ip address is not custom vpc's kubernetes service. Instead, it uses default kubernetes

the default vpc (vpc0) works as designed. Only customer vpc does not work

**Environment**:
- Arktos version (0.3`):
- Cloud provider or hardware configuration: AWS
- OS (e.g: `cat /etc/os-release`): Ubuntu
- Kernel (e.g. `uname -a`): 5.6.0-rc2
- Install tools:
- Others:

h-w-chen:
**What type of PR is this?**
/kind bug

This PR makes coredns pods of per-network DNS deployment connecting to the specified kube-apiserver.

When a network resource is created, arktos network controller creates its per-network DNS deployment; however, at this moment, the controller might not have any idea of the per-network kubernetes service IP (for external IPAM networks) - actually it uses 10.0.0.1 by default,  leading to coredns unable to get hold of the api server, thus cannot start properly.

Making the DNS deployment to define its pod template so that coredns pod connects to kube-apiserver directly solve this kubernetes service ip not yet assigned issue.

In the long run, the controller may need to wait for per-network kubernetes service IP assigned, before it creates per-network DNS deployment.

Fixes #742

YES
```release-note
when arktos_network_controller is started, it requires --kube-apiserver-ip argument
```

### DIFF
--- a/cmd/arktos-network-controller/app/dns_deployment.go
+++ b/cmd/arktos-network-controller/app/dns_deployment.go
@@ -31,7 +31,7 @@ import (
 	"k8s.io/client-go/kubernetes"
 )
 
-func deployDNSForNetwork(net *v1.Network, client kubernetes.Interface, domainName, kubeAPIServerIP string) error {
+func deployDNSForNetwork(net *v1.Network, client kubernetes.Interface, domainName, kubeAPIServerIP, kubeAPIServerPort string) error {
 	if err := ensureToCreateServiceAccount(net, client); err != nil {
 		return err
 	}
@@ -48,7 +48,7 @@ func deployDNSForNetwork(net *v1.Network, client kubernetes.Interface, domainNam
 		return err
 	}
 
-	if err := ensureToCreateDeployment(net, client, kubeAPIServerIP); err != nil {
+	if err := ensureToCreateDeployment(net, client, kubeAPIServerIP, kubeAPIServerPort); err != nil {
 		return err
 	}
 
@@ -174,7 +174,7 @@ func ensureToCreateConfigMap(net *v1.Network, client kubernetes.Interface, domai
 	return nil
 }
 
-func ensureToCreateDeployment(net *v1.Network, client kubernetes.Interface, kubeAPIServerIP string) error {
+func ensureToCreateDeployment(net *v1.Network, client kubernetes.Interface, kubeAPIServerIP, kubeAPIServerPort string) error {
 	name := dnsBaseName + "-" + net.Name
 	label := dnsServiceDefaultName + "-" + net.Name
 	configmap := dnsBaseName + "-" + net.Name
@@ -267,7 +267,7 @@ func ensureToCreateDeployment(net *v1.Network, client kubernetes.Interface, kube
 								},
 								Env: []corev1.EnvVar{
 									{Name: "KUBERNETES_SERVICE_HOST", Value: kubeAPIServerIP},
-									{Name: "KUBERNETES_SERVICE_PORT", Value: "6443"},
+									{Name: "KUBERNETES_SERVICE_PORT", Value: kubeAPIServerPort},
 								},
 								VolumeMounts: []corev1.VolumeMount{
 									{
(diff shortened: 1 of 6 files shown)


cmd/arktos-network-controller/network-controller.go
```diff
@@ -47,6 +48,10 @@ func main() {
 		workers = defaultWorkers
 	}
 
+	if len(kubeAPIServerIP) == 0 {
```
chenqianfzh, line 51: Nit: As KubeAPIserverIP is a must-have param, shall we change the check to net.ParseIP() to make sure the IP address is valid? Or even more, try whether the 6443 port in that address is reachable to make sure it is an APIServer?

h-w-chen: it makes sense to have a more stringent check; will update as suggested.

chenqianfzh (commented):
Looks good to me as a temporary workaround for an urgent issue. 

Leaving a minor comment, which should not block the check-in.

chenqianfzh (approved):

cmd/arktos-network-controller/app/dns_deployment.go
```diff
@@ -265,6 +265,10 @@ func ensureToCreateDeployment(net *v1.Network, client kubernetes.Interface) erro
 									"-conf",
 									"/etc/coredns/Corefile",
 								},
+								Env: []corev1.EnvVar{
+									{Name: "KUBERNETES_SERVICE_HOST", Value: kubeAPIServerIP},
```
Sindica, line 269: It is strange that api server ip address is saved in deployment spec. Ideally, API server would be accessed via load balancer and dynamically allocated during system run. It could be changing as system maintenance is going on. Put it into spec make the data easily to stale and problematic even in short period.

cmd/arktos-network-controller/network-controller.go
```diff
@@ -90,4 +104,6 @@ func init() {
 	flag.StringVar(&masterURL, "master", "", "The address of the Kubernetes API server. Overrides any value in kubeconfig. Only required if out-of-cluster.")
 	flag.IntVar(&workers, "concurrent-workers", defaultWorkers, "The number of workers that are allowed to process concurrently.")
 	flag.StringVar(&domainName, "cluster-domain", "cluster.local", "the cluster-internal domain name for Services.")
+	flag.StringVar(&kubeAPIServerIP, "kube-apiserver-ip", "", "the ip address kube-apiserver is listening at.")
```
Sindica, line 107: This is not a sustainable design to get kube api server ip from argument.

h-w-chen: agreed; issue #745 has been opened for the long-term fix.

cmd/arktos-network-controller/app/dns_deployment.go
h-w-chen: this workaround is temporary fix to unblock mizar integration; the long-term solution should, as you had guessed out, use kubernetes service ip instead - issue #745 opened to track.

For now, we cannot adopt k8s service ip directly, as it would break flat-type networks.

PR #749, opened 2020-09-18
add switch to set --leader-elect=false for kube-scheduler

sonyafenge:
**What type of PR is this?**

> /kind feature

add switch to set --leader-elect=false for kube-scheduler,  set "ENABLE_SCHEDULER_LEADER_ELECT" to false if you hope to disable kube-scheduler leader election.

Fixes #

```release-note

```


Sindica (approved):
LGTM

yb01:
I suggest not to make this as default settings to disable the leader election. it was turn off just to help us focus on the investigation of issue 621, but keeping the scheduler and controller up during load tests. since it is already an env setting, it should not required for perf testing to enable/disable it on needed basis.

PR #753, opened 2020-09-21
Fix tiny issue found during testing: info should be infof

Hong-Chang:
**What type of PR is this?**
/kind bug


h-w-chen (approved):
looks good.

PR #755, opened 2020-09-22
install cni binary package on arktos worker node

h-w-chen:
**What type of PR is this?**
/kind bug

This PR ensure the common cni binary package to be installed when worker node is being started.

Besides, it documents the security group setting we verified working on AWS for multi-node communication.

NONE


chenqianfzh (approved):
Thanks for the quick update of the doc and fixing the issue we found in troubleshooting.

PR #757, opened 2020-09-23
Added a missing quote in etcd.manifest

jshaofuturewei:
It does not matter. However, it is better to fix it.

### DIFF
--- a/cluster/gce/manifests/etcd.manifest
+++ b/cluster/gce/manifests/etcd.manifest
@@ -30,7 +30,7 @@
         "value": "{{ pillar.get('storage_backend', 'etcd3') }}"
       },
       { "name": "TARGET_VERSION",
-        "value": "{{ pillar.get('etcd_version', '3.4.4-arktos.1-1) }}"
+        "value": "{{ pillar.get('etcd_version', '3.4.4-arktos.1-1') }}"
       },
       { "name": "DATA_DIRECTORY",
         "value": "/var/etcd/data{{ suffix }}"


Sindica (approved):

pdgetrf:
what's the impact of this misspell before this fix?

jshaofuturewei:
> what's the impact of this misspell before this fix?

It is used by migrate-if-needed.sh for an etcd version that can be migrated to. migrate-if-needed.sh is not triggered in arktos since we don't have any version that we need to migrate to.

pdgetrf:
> > what's the impact of this misspell before this fix?
> 
> It is used by migrate-if-needed.sh for an etcd version that can be migrated to. migrate-if-needed.sh is not triggered in arktos since we don't have any version that we need to migrate to.

how did this pass the test done for the previous PR that merged it in?

jshaofuturewei:
> > > what's the impact of this misspell before this fix?
> > 
> > 
> > It is used by migrate-if-needed.sh for an etcd version that can be migrated to. migrate-if-needed.sh is not triggered in arktos since we don't have any version that we need to migrate to.
> 
> how did this pass the test done for the previous PR that merged it in?

There is no migration scenario or test case to use the variable for arktos.

pdgetrf:
> > > > what's the impact of this misspell before this fix?
> > > 
> > > 
> > > It is used by migrate-if-needed.sh for an etcd version that can be migrated to. migrate-if-needed.sh is not triggered in arktos since we don't have any version that we need to migrate to.
> > 
> > 
> > how did this pass the test done for the previous PR that merged it in?
> 
> There is no migration scenario or test case to use the variable for arktos.

was it manually tested on a local environment at least?

jshaofuturewei:
> > > > > what's the impact of this misspell before this fix?
> > > > 
> > > > 
> > > > It is used by migrate-if-needed.sh for an etcd version that can be migrated to. migrate-if-needed.sh is not triggered in arktos since we don't have any version that we need to migrate to.
> > > 
> > > 
> > > how did this pass the test done for the previous PR that merged it in?
> > 
> > 
> > There is no migration scenario or test case to use the variable for arktos.
> 
> was it manually tested on a local environment at least?

In an etcd container, run export and you will see the TARGET_VERSION


declare -x TARGET_STORAGE="etcd3"
declare -x TARGET_VERSION="3.4.3"
declare -x TERM="xterm"

PR #762, opened 2020-09-24
workaround for cicd failures

chenqianfzh:
This once-flaky test turns nasty recently. Its consistent failures in multiples PRs ( including several PRs about doc changes) is annoying. 

A workaround is given in this PR. I extended the waiting time and it unblocked me in my local test runs.


staging/src/k8s.io/apimachinery/pkg/watch/watch_test.go
```diff
@@ -329,7 +329,7 @@ func TestAggregatedWatchOnError(t *testing.T) {
 
 		wg.Wait()
 		agg.Stop()
-		time.Sleep(time.Millisecond)
+		time.Sleep(10 * time.Millisecond)
```
Sindica, line 332: This is not the problem

PR #774, opened 2020-10-06
private change to getClient() in etcd3

chenqianfzh:
trial


staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go
```diff
@@ -518,7 +525,7 @@ func (s *store) GetToList(ctx context.Context, key string, resourceVersion strin
 func (s *store) Count(key string) (int64, error) {
 	key = path.Join(s.pathPrefix, key)
 	startTime := time.Now()
-	getResp, err := s.getClientFromKey(key).KV.Get(context.Background(), key, clientv3.WithRange(clientv3.GetPrefixRangeEnd(key)), clientv3.WithCountOnly())
+	getResp, err := s.client.KV.Get(context.Background(), key, clientv3.WithRange(clientv3.GetPrefixRangeEnd(key)), clientv3.WithCountOnly())
```
Sindica, line 528: client is ETCD client. Here you need to know which ETCD to go

staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go
```diff
@@ -1168,10 +1142,10 @@ func getTenantForKey(segs []string, posToGet int) string {
 
 // getClientsFromKey is used by list to get all related storage clusters
 // For example: get pods from all tenants
-func (s *store) getClientsFromKey(key string) map[uint8]*clientv3.Client {
+func (s *store) getClientsFromKey(key, tenant string) map[uint8]*clientv3.Client {
```
Sindica, line 1145: Sometimes it is a list that need to go to multiple client. Such as list pods across all tenants.

PR #782, opened 2020-10-16
Remove bcast from aggregated watcher. Use watcher stop to exit goroutine

Sindica:


staging/src/k8s.io/apimachinery/pkg/watch/watch.go
```diff
@@ -104,34 +101,43 @@ func NewAggregatedWatcherWithOneWatch(watcher Interface, err error) *AggregatedW
...
 
 func (a *AggregatedWatcher) removeWatcherAndError(watcher Interface) {
+	if watcher == nil {
+		return
+	}
+	watcher.Stop()
 	a.mapLock.Lock()
 	for k, v := range a.watchers {
 		if v == watcher {
 			delete(a.watchers, k)
```
yb01, line 126: should check if key found before delete ?

staging/src/k8s.io/apimachinery/pkg/watch/watch.go
```diff
@@ -140,40 +146,40 @@ func (a *AggregatedWatcher) AddWatchInterface(watcher Interface, err error) {
...
-	a.stopChGrp.Leave(stopCh)
+func (a *AggregatedWatcher) closeWatcher(watcher Interface) {
 	a.removeWatcherAndError(watcher)
 
-	if !a.allowWatcherReset && a.stopChGrp.MemberCount() == 0 {
-		//klog.Infof("Close watcher %v caused aggregated channel %v closed", watcher, a.aggChan)
-		a.stopChGrp.Close()
-		close(a.aggChan)
+	if !a.allowWatcherReset {
+		a.mapLock.RLock()
```
yb01, line 176: if racing condition is a concern here, the Lock, Unlock() should be used instead of Block()

yb01 (approved):
/lgtm

PR #787, opened 2020-10-20
Agg watcher reduce usage pr

Sindica:
Perf run in https://github.com/futurewei-cloud/arktos-perftest/tree/perf-20201019
CI passed in https://github.com/centaurus-cloud/arktos/pull/784
Compacted commits for easiler review

### DIFF
--- a/staging/src/k8s.io/apimachinery/pkg/watch/watch.go
+++ b/staging/src/k8s.io/apimachinery/pkg/watch/watch.go
@@ -95,12 +95,6 @@ func NewAggregatedWatcherWithReset(ctx context.Context) *AggregatedWatcher {
 	return a
 }
 
-func NewAggregatedWatcherWithOneWatch(watcher Interface, err error) *AggregatedWatcher {
-	a := NewAggregatedWatcher()
-	a.AddWatchInterface(watcher, err)
-	return a
-}
-
 func (a *AggregatedWatcher) addWatcherAndError(watcher Interface, err error) (stopchan chan int) {
 	a.mapLock.Lock()
 	a.watchers[a.watcherIndex] = watcher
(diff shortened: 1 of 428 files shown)


staging/src/k8s.io/apimachinery/pkg/watch/watch.go
```diff
@@ -20,12 +20,10 @@ package watch
 import (
 	"context"
 	"fmt"
-	"github.com/grafov/bcast"
```
yb01, line 23: those are in PR 782 already which is merged. can rebase to remove them.

Sindica: Yes, I based this branch from previous commit before it was merged into master and did not rebase to save CI rerun.

staging/src/k8s.io/client-go/tools/watch/retrywatcher.go
```diff
@@ -101,26 +101,12 @@ func (rw *RetryWatcher) send(event watch.Event) bool {
...
-	}
-
-	defer aggWatchers.Stop()
-	if aggWatchers.GetWatchersCount() == 0 {
-		klog.Error("No watcher in aggregated watchers")
-		return false, 0
-	}
-
-	err := aggWatchers.GetErrors()
 	switch err {
```
yb01, line 110: are the errors from the watcher and aggWatcher.GetErrors() match ?

yb01 (approved):
/lgtm

staging/src/k8s.io/client-go/tools/watch/retrywatcher.go
Sindica: This is a revert back to previous implementation. Watch returns watch result and error if any. When I changed the return type of Watch interface, I have to aggregate the watch results and errors. Now the implementation of Watch function will decide how to return error. The auto generated client code calls GetErrors to return error.

PR #791, opened 2020-10-23
etcd grpc proxy

chenqianfzh:
### Why this PR?
This PR allows perf tests to enable etcd-grpc-proxy based on the environement variables. 
If env var START_ETCD_GRPC_PROXY=true, the grpc proxy will be kicked off and the kube-apiserver will connect to the proxy port. Otherwise, kube-apiserver will talk to the etcd core server directly.

During 5k-node perf tests, we see a lot of "took too long" warnings in the core etcd server log and also many etcd delay warnings in Kube-Apiserver logs.  As the etcd grpc proxy offloads the core etcd server by coalescing the watch requests (more details are available in https://etcd.io/docs/v3.1.12/op-guide/grpc_proxy/), we tried to enable the grpc proxy to see whether it solves this issue. 

As the test run results (<em>logs can be found under GCP project: workload-controller-manager on sonya-uswest2: /home/sonyali/logs/perf-test/gce-5000/arktos/1020-qianprivate-1a0w1eshow</em>) , the proxy helps reduce the number of "took too long" warnings ( <em>the logs of the comparing run are under GCP project: workload-controller-manager on sonyadev4: /home/sonyali/logs/perf-test/gce-5000/arktos/1019-daily1019-1a0w1e</em>).

So the team decides to turn grpc proxy off be default in the continuing search for the true reason for the etcd delays, yet at the same time make this feature available within a change of env variable.

### Verification
Tested with perf-test of 100-nodes:
1. running kube-up.sh with START_ETCD_GRPC_PROXY=true. The kube-up script succeeded and the cluster was working. Verified that etcd grpc proxy was started, and kube-apiserver connected to etcd servers via the proxy port (23790）.

1. running kube-up.sh with START_ETCD_GRPC_PROXY not set. The kube-up script succeeded and the cluster was working. Verified that etcd grpc proxy was not started, and kube-apiserver connected to etcd servers via the core server port (2379）.

### DIFF
--- a/cluster/common.sh
+++ b/cluster/common.sh
@@ -413,6 +413,14 @@ function generate-etcd-cert() {
                     "server auth",
                     "client auth"
                 ]
+            },
+            "grpcproxy": {
+                "expiry": "43800h",
+                "usages": [
+                    "signing",
+                    "key encipherment",
+                    "client auth"
+                ]
             }
         }
     }
@@ -466,6 +474,12 @@ EOF
       echo "Generate peer certificates..."
       echo '{"CN":"'${member_ip}'","hosts":[""],"key":{"algo":"ecdsa","size":256}}' \
        | ${CFSSL_BIN} gencert -ca=ca.pem -ca-key=ca-key.pem -config=ca-config.json -profile=peer -hostname="${member_ip},127.0.0.1" - \
+       | ${CFSSLJSON_BIN} -bare "${prefix}"
+      ;;
+    grpcproxy)
+      echo "Generate grpc_proxy certificates..."
+      echo '{"CN":"","hosts":["*"],"key":{"algo":"ecdsa","size":256}}' \
+       | ${CFSSL_BIN} gencert -ca=ca.pem -ca-key=ca-key.pem -config=ca-config.json -profile=grpcproxy - \
        | ${CFSSLJSON_BIN} -bare "${prefix}"
       ;;
     *)
--- a/cluster/gce/gci/configure-helper-common.sh
+++ b/cluster/gce/gci/configure-helper-common.sh
@@ -1146,6 +1146,12 @@ function create-master-etcd-apiserver-auth {
 
      ETCD_APISERVER_CLIENT_CERT_PATH="${auth_dir}/etcd-apiserver-client.crt"
      echo "${ETCD_APISERVER_CLIENT_CERT}" | base64 --decode | gunzip > "${ETCD_APISERVER_CLIENT_CERT_PATH}"
+
+     ETCD_GRPC_PROXY_CLIENT_KEY_PATH="${auth_dir}/etcd-grpc-proxy-client.key"
+     echo "${ETCD_GRPC_PROXY_CLIENT_KEY}" | base64 --decode > "${ETCD_GRPC_PROXY_CLIENT_KEY_PATH}"
+
+     ETCD_GRPC_PROXY_CLIENT_CERT_PATH="${auth_dir}/etcd-grpc-proxy-client.crt"
+     echo "${ETCD_GRPC_PROXY_CLIENT_CERT}" | base64 --decode | gunzip > "${ETCD_GRPC_PROXY_CLIENT_CERT_PATH}"
    fi
 }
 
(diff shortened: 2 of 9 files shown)


cluster/common.sh
```diff
@@ -413,6 +413,14 @@ function generate-etcd-cert() {
                     "server auth",
                     "client auth"
                 ]
+            },
+            "grpcproxy": {
+                "expiry": "43800h",
```
pdgetrf, line 418: 43800h is just a large time so it never expires?

cluster/gce/gci/configure-helper-common.sh
```diff
@@ -1422,6 +1428,7 @@ function prepare-etcd-manifest {
   sed -i -e "s@{{ *etcd_cluster *}}@$etcd_cluster@g" "${temp_file}"
   sed -i -e "s@{{ *liveness_probe_initial_delay *}}@${ETCD_LIVENESS_PROBE_INITIAL_DELAY_SEC:-15}@g" "${temp_file}"
   sed -i -e "s@{{ *listen_client_ip *}}@${ETCD_LISTEN_CLIENT_IP:-127.0.0.1}@g" "${temp_file}"
+
```
pdgetrf, line 1431: nit: extra new line

cluster/gce/gci/configure-helper-common.sh
```diff
@@ -1491,6 +1544,15 @@ function start-etcd-servers {
   prepare-etcd-manifest "-events" "4002" "2381" "100m" "etcd-events.manifest"
 }
 
+function start-etcd-grpc-proxy {
+  echo "Start etcd grpc proxy pod..."
+
+  prepare-log-file /var/log/etcd-grpc-proxy.log
+  prepare-etcd-grpc-proxy-manifest "" "2379" "23790" "200m" "etcd-grpc-proxy.manifest"
```
pdgetrf, line 1551: what's 200m here?

cluster/gce/gci/configure-helper-common.sh
```diff
@@ -1491,6 +1544,15 @@ function start-etcd-servers {
...
+
+  ETCD_SERVERS=${ETCD_GRPC_PROXIES:-http://127.0.0.1:23790}
```
pdgetrf, line 1553: nit: use a string variable for 23790 here and line 1551

cluster/gce/util.sh
```diff
@@ -2720,6 +2732,15 @@ function create-master() {
       --allow "tcp:2380,tcp:2381" &
   fi
 
+  # Create rule for etcd grpc proxy.
+  if [[ "${START_ETCD_GRPC_PROXY:-}" == "true" ]]; then
+    gcloud compute firewall-rules create "${MASTER_NAME}-etcd-grpc-proxy" \
+      --project "${NETWORK_PROJECT}" \
+      --network "${NETWORK}" \
+      --source-ranges "0.0.0.0/0" \
+      --allow tcp:23790 &
```
pdgetrf, line 2741: also 23790 here. it's spread in different files. it'd be great if we could get these in a common place.

pdgetrf (approved):
LGTM

minor comments inline.

cluster/common.sh
chenqianfzh: 43800h = 5 years, which means a long enough time in this context.

I copied this value from the config in the same file.

cluster/gce/gci/configure-helper-common.sh
chenqianfzh: 200m is the cpulimit value of the pod.

cluster/gce/gci/configure-helper-common.sh
chenqianfzh: Actually, I did not use a variable for "23790" as I found there is no variable for "2379", which the port for core etcd server. I guess the reason is that it is a script, which is hard to debug and thus readability is more important.

cluster/gce/gci/configure-helper-common.sh
chenqianfzh: will delete it.

cluster/gce/util.sh
chenqianfzh: Actually, I did not use a variable for "23790" as I found throughout the script there is no variable for "2379", which is the port for core etcd server. Actually, the string "2379" is used much more frequently than "23790". I guess the reason is that it is a script, which is hard to debug and thus readability is more important.

PR #800, opened 2020-11-02
Start replicaset controller and deployment controller in KCM.

Sindica:


cmd/kube-controller-manager/app/controllermanager.go
```diff
@@ -392,8 +392,8 @@ func NewControllerInitializers(loopMode ControllerLoopMode) map[string]InitFunc
 	controllers["garbagecollector"] = startGarbageCollectorController
 	controllers["daemonset"] = startDaemonSetController
 	controllers["job"] = startJobController
-	//controllers["deployment"] = startDeploymentController
-	//controllers["replicaset"] = startReplicaSetController
+	controllers["deployment"] = startDeploymentController
+	controllers["replicaset"] = startReplicaSetController
```
vinaykul, line 396: Can we also add change to disable workload controller that is launched as part of kube-up?

Sindica: I don't know how to do that in kube-up

PR #809, opened 2020-11-09
fix scale-out setup script

pdgetrf:
### Changes
- fix proxy watch timeout
- remove nodeipam from resource partition
- remove changes from hack/common.sh and util to make arktos-up.sh backward compatible

### Verification
- Start up scale-out cluster and verified pod running
- Start up normal arktos-up cluster and verified pod running


chenqianfzh (approved):
lgtm

PR #810, opened 2020-11-10
nginx proxy setting with two TPs

chenqianfzh:
This change add scripts to config the nginx proxy for two TPs & one RP.

Run:

TENANT_PARTITION_ONE_IP=172.31.26.146  TENANT_PARTITION_TWO_IP=172.31.26.99  RESOURCE_PARTITION_IP=172.31.23.48 hack/scale_out_poc/two_TP/setup_nginx_proxy.sh

### DIFF
--- a/hack/scale_out_poc/two_TP/nginx.conf
+++ b/hack/scale_out_poc/two_TP/nginx.conf
@@ -75,22 +75,6 @@ http {
                 }
 
                 #################################################################
-                # forward non-resource requests back to its source partition
-                #################################################################
-
-                location ~ ^/[a-zA-Z0-9_.-]+$ {
-                    proxy_pass http://$remote_addr:8080;
-                }
-
-                location ~ ^/[a-zA-Z0-9_.-]+/[a-zA-Z0-9_.-]+$ {
-                    proxy_pass http://$remote_addr:8080;
-                }
-
-                location ~ ^/apis/[a-zA-Z0-9_.-]+/[a-zA-Z0-9_.-]+$ {
-                    proxy_pass http://$remote_addr:8080;
-                }
-
-                #################################################################
                 # forward the request of system-tenants or no tenants specificed back to its source partition
                 #################################################################
 
--- a/hack/scale_out_poc/two_TP/nginx.conf
+++ b/hack/scale_out_poc/two_TP/nginx.conf
@@ -1,0 +1,144 @@
+user www-data;
+worker_processes auto;
+pid /run/nginx.pid;
+include /etc/nginx/modules-enabled/*.conf;
+
+events {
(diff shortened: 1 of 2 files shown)


hack/scale_out_poc/two_TP/nginx.conf
```diff
@@ -0,0 +1,144 @@
...
+
+                location ~ ^/apis/coordination.k8s.io/v1/leases?(.*) {
+                    proxy_pass $resource_api;
+                }
+
+                location ~ ^/apis/([^/])*/([^/])*/tenants/system/namespaces/kube-node-lease/leases?(.*) {
+                    proxy_pass $resource_api;
+                }
+
+                #################################################################
```
XiaoningDing, line 77: Why do we need to keep the non-resource section and the non-tenant section, are they already covered by the last rule?

hack/scale_out_poc/two_TP/nginx.conf
```diff
@@ -0,0 +1,144 @@
...
+                }
+
+                location ~ ^/apis/[a-zA-Z0-9_.-]+/[a-zA-Z0-9_.-]+$ {
+                    proxy_pass http://$remote_addr:8080;
+                }
+
+                #################################################################
+                # forward the request of system-tenants or no tenants specificed back to its source partition
+                #################################################################
+
```
yb01, line 96: by using the remote_addr, it assume the request is from the IP the same as the cluster Api Server IP, which is not true, i.e. controller, scheduler, kubelet, or other tools/clients etc can all be on different machine than the Api Server.

hack/scale_out_poc/two_TP/nginx.conf
chenqianfzh: You made a point. Updated as suggested.

hack/scale_out_poc/two_TP/nginx.conf
chenqianfzh: I see your point. Yet this script is for POC, where each partition is in one box. More work need to be done when different components reside in different machines. Let's just use this script as is for now.

hack/scale_out_poc/two_TP/nginx.conf
```diff
@@ -0,0 +1,128 @@
...
+    ##
+    # Basic Settings
+    ##
+
+    sendfile on;
+    tcp_nopush on;
+    tcp_nodelay on;
+    keepalive_timeout 65;
+    types_hash_max_size 2048; 
+    proxy_http_version 1.1;
```
pdgetrf, line 22: just realized. this is missing 

    proxy_read_timeout     1200;
    proxy_connect_timeout  1200;

PR #812, opened 2020-11-11
Make sure lease create/renew can use different api server

Sindica:
Pick up perf test changes

### DIFF
--- a/pkg/kubelet/nodelease/controller.go
+++ b/pkg/kubelet/nodelease/controller.go
@@ -27,7 +27,6 @@ import (
 	"k8s.io/apimachinery/pkg/util/clock"
 	"k8s.io/apimachinery/pkg/util/wait"
 	clientset "k8s.io/client-go/kubernetes"
-	coordclientset "k8s.io/client-go/kubernetes/typed/coordination/v1beta1"
 	"k8s.io/utils/pointer"
 
 	"k8s.io/klog"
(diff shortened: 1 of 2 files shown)


pkg/kubelet/nodelease/controller.go
```diff
@@ -123,10 +116,11 @@ func (c *controller) backoffEnsureLease() (*coordv1beta1.Lease, bool) {
 // ensureLease creates the lease if it does not exist. Returns the lease and
 // a bool (true if this call created the lease), or any error that occurs.
 func (c *controller) ensureLease() (*coordv1beta1.Lease, bool, error) {
-	lease, err := c.leaseClient.Get(c.holderIdentity, metav1.GetOptions{})
+	leaseClient := c.client.CoordinationV1beta1().Leases(corev1.NamespaceNodeLease)
```
yb01, line 119: understand the intention is to avoid all node lease OPs come to the same Api Server instance in the api-server- partitioned-clusters, but is this code ( line 119 ) will randomly pick up one Api Server ?

Sindica: Yes it does. The reason leaseClient could not do it is because it was initialized at kubelet startup and won't be able to pick up new api server(s) changes. c.client.CoordinationV1beta1().Leases(ns) returns the standard clientset, which will auotomatically pick up api server updates and randomize the target api server when request is sent. Promethus metrics apiserver_request_count{resource="leases"} proved node lease GET/PUT request from kubelet is now balanced across all api servers.

yb01 (approved):
/lgtm

PR #817, opened 2020-11-13
1.18 improvement api server cacher index and reduce watchers

Sindica:


staging/src/k8s.io/apiserver/pkg/storage/cacher/cacher.go
```diff
@@ -300,15 +309,38 @@ func NewCacherFromConfig(config Config) *Cacher {
...
+		// For now, we don't support multiple trigger functions defined
+		// for a given resource.
+		if len(config.IndexerFuncs) > 1 {
+			// community changed NewCacherFromConfig signature in
+			// https://github.com/kubernetes/kubernetes/commit/ee13be28845e8c55f6e2a2207ca4f46cd67fdb4e
+			// Propagate error from creating cacher and storage decorators up
+			// Jul 15, 2019
+
+			//return nil, fmt.Errorf("cacher %s doesn't support more than one IndexerFuncs: ", reflect.TypeOf(obj).String())
+			return nil
```
yb01, line 323: if we only support 1 indexFunc, this should return error instead of nil.

yb01 (approved):
/lgtm

staging/src/k8s.io/apiserver/pkg/storage/cacher/cacher.go
Sindica: Community changed the interface of NewCacherFromConfig to return surface errors but Arktos does not support it now. Keep the interface now to avoid massive changes.

PR #825, opened 2020-11-19
Fix concurrent map write issue at AggregateWatcher stop

Sindica:


staging/src/k8s.io/apimachinery/pkg/watch/watch.go
```diff
@@ -195,11 +195,16 @@ func (a *AggregatedWatcher) Stop() {
 	if !a.stopped {
 		a.stopped = true
 		a.allowWatcherReset = false
+
+		a.mapLock.RLock()
```
yb01, line 199: not sure why to write to the stop channel needs unlock the watcher map, some comment can help here.

yb01 (approved):
/lgtm

staging/src/k8s.io/apimachinery/pkg/watch/watch.go
Sindica: Updated due to CR feedback.

PR #827, opened 2020-11-19
fix NodeNotReady issue

chenqianfzh:
This PR fixes the issue that the node lifecycle controller cannot get the node-lease from the lister when connecting to nginx proxy.

### Explanation
The cause is due to the following forward logic in nginx.conf:
        location ~ ^/apis/coordination.k8s.io/**v1**/leases?(.*) {
            proxy_pass $RESOURCE_API;
        }

Yet, arktos use version V1beta1 for node-lease requests. As a result, node-leases are not correctly forwarded.

### Verification:

Tested in my dev env, the NodeNotReady issue was gone.

### DIFF
--- a/hack/scale_out_poc/nginx.conf
+++ b/hack/scale_out_poc/nginx.conf
@@ -71,11 +71,11 @@ http {
                     proxy_pass {{ arktos_api_protocol }}://$remote_addr:{{ arktos_api_port }};
                 }
 
-                location ~ ^/api/v1/nodes?(.*) {
+                location ~ ^/api/([^/])*/nodes?(.*) {
                     proxy_pass $resource_api;
                 }
 
-                location ~ ^/apis/coordination.k8s.io/v1/leases?(.*) {
+                location ~ ^/apis/coordination.k8s.io/([^/])*/leases?(.*) {
                     proxy_pass $resource_api;
                 }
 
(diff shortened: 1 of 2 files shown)


pdgetrf (approved):

hack/scale_out_poc/nginx.conf
```diff
@@ -71,11 +71,11 @@ http {
                     proxy_pass {{ arktos_api_protocol }}://$remote_addr:{{ arktos_api_port }};
                 }
 
-                location ~ ^/api/v1/nodes?(.*) {
+                location ~ ^/api/([^/])*/nodes?(.*) {
                     proxy_pass $resource_api;
                 }
 
-                location ~ ^/apis/coordination.k8s.io/v1/leases?(.*) {
```
yb01, line 78: there used to be code to fall back to remote_addr:8080 whenever the location is not found, which I have in my tests env with the PR 810 code. any reason those default location was removed ? the fallback to default should address such issues --- which was the reason I didn't see this in my env.

chenqianfzh: This is for m1. The one you mentioned is a different file, which is for m2 only. 

If we use the logic to remote_addr:8080 whenever the location is not found in m1, the requests for pods/daemonsets from the components in RP will not be forwarded to TP side.

PR #828, opened 2020-11-19
fix secret list and watch

pdgetrf:
## Changes
 This fix sets kubelet to use the correct kube client for secrets list and watch. Without it, pod creation was failing on the 2nd tenant partition during volume mounting. 

## Validation

Validated end to end on local environment. 

On tenant partition 1
```bash
root@goose-t:~/code/src/k8s.io/arktos/test/yaml# kubectl get pods --tenant alex -o wide
NAME              HASHKEY               READY   STATUS    RESTARTS   AGE   IP            NODE      NOMINATED NODE   READINESS GATES
k8s-vanilla-pod   3683182769778565890   1/1     Running   0          32s   10.88.3.108   goose-r   <none>           <none>
```

On tenant partition 2
```bash
root@goose-3:~/code/src/k8s.io/arktos/test/yaml# kubectl get pod --tenant oxygen -o wide
NAME              HASHKEY               READY   STATUS    RESTARTS   AGE   IP            NODE      NOMINATED NODE   READINESS GATES
k8s-vanilla-pod   3367483529165770237   1/1     Running   0          16s   10.88.3.107   goose-r   <none>           <none>
```

## Notes to reviewers
The *getTPClient* function is now being repeated in a few places. Ugly, yes. For now please consider this a makeshift solution for the POC. As more conversion goes in, it will be refactored into a common place.

### DIFF
--- a/pkg/kubelet/secret/secret_manager.go
+++ b/pkg/kubelet/secret/secret_manager.go
@@ -19,6 +19,7 @@ package secret
 
 import (
 	"fmt"
+	"k8s.io/klog"
 	"time"
 
 	"k8s.io/api/core/v1"
@@ -111,6 +112,19 @@ const (
 	defaultTTL = time.Minute
 )
 
+func getTPClient(kubeClients []clientset.Interface, tenant string) clientset.Interface {
+	var client clientset.Interface
+	pick := 0
+	if tenant[0] <= 'm' {
+		client = kubeClients[0]
+	} else {
+		client = kubeClients[1]
+		pick = 1
+	}
+	klog.Infof("tenant %s using client # %d", tenant, pick)
+	return client
+}
+
 // NewCachingSecretManager creates a manager that keeps a cache of all secrets
 // necessary for registered pods.
 // It implements the following logic:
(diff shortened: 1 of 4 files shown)


pkg/kubelet/secret/secret_manager.go
```diff
@@ -111,6 +112,19 @@ const (
 	defaultTTL = time.Minute
 )
 
+func getTPClient(kubeClients []clientset.Interface, tenant string) clientset.Interface {
```
yb01, line 115: I thought we moved this function to a common util function, no ?

yb01 (approved):
/lgtm

pkg/kubelet/secret/secret_manager.go
pdgetrf: will do later as described in the description. this change is fixing a blocking issue so let's have it in first.

XiaoningDing, line 115: Remember to add one line to handle the case of one TP, for the purpose of M1 testing.

pdgetrf: next PR will be for this

PR #830, opened 2020-11-20
add fallback to kubeclient 0 if only one is provided

pdgetrf:
## Changes
if only one kubeclient is provided (most likely for a phase one merge), switch kubelet to use it. 

## Validation
Validate on local environment with
1. 2 tp 1 rp, 2 endpoints in TENANT_SERVERS
2. 1 tp 1 rp, 1 endpoint in TENANT_SERVERS

Pod goes into running into both cases.


pdgetrf:
change list doesn't look right. fixing.

yb01 (approved):
/lgtm

PR #831, opened 2020-11-20
Modify cluster verification to bypass proxy and directly talk to cluster apiserver

vinaykul:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
/kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**: Nginx proxy is flaky. Cluster verification script that talks to the apiserver via proxy intermittently fails and this can cause the cluster bringup script to take a long time to start or intermittently fail. Verifying directly against apiserver is a reliable way to ensure actual cluster is fine. The test will still use proxy as per architecture.

Fixes #

```release-note

```

### DIFF
--- a/cluster/gce/util.sh
+++ b/cluster/gce/util.sh
@@ -2362,7 +2362,8 @@ function kube-up() {
     if [[ "${KUBERNETES_SCALEOUT_PROXY:-false}" == "true" ]]; then
       echo "Scaleout proxy public IP: ${PROXY_RESERVED_IP}:8888"
       echo "Scaleout proxy internal IP: ${PROXY_RESERVED_INTERNAL_IP}:8888"
-      sed -i "s/server: https:\/\/.*/server: http:\/\/${PROXY_RESERVED_IP}:8888/" ${KUBECONFIG}
+      cp -f ${KUBECONFIG} ${LOCAL_KUBECONFIG}
+      sed -i "s/server: https:\/\/.*/server: http:\/\/${PROXY_RESERVED_IP}:8888/" ${LOCAL_KUBECONFIG}
     fi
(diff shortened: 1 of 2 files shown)


cluster/gce/util.sh
```diff
@@ -2362,7 +2362,8 @@ function kube-up() {
     if [[ "${KUBERNETES_SCALEOUT_PROXY:-false}" == "true" ]]; then
       echo "Scaleout proxy public IP: ${PROXY_RESERVED_IP}:8888"
       echo "Scaleout proxy internal IP: ${PROXY_RESERVED_INTERNAL_IP}:8888"
-      sed -i "s/server: https:\/\/.*/server: http:\/\/${PROXY_RESERVED_IP}:8888/" ${KUBECONFIG}
+      cp -f ${KUBECONFIG} ${LOCAL_KUBECONFIG}
+      sed -i "s/server: https:\/\/.*/server: http:\/\/${PROXY_RESERVED_IP}:8888/" ${LOCAL_KUBECONFIG}
```
yb01, line 2366: the local_kubeconfig was updated every time the create_kubemark-Master is called. we probably leave it untouched and if any logic is needed in the upper level function, i.e. in the start Kubemark.sh, they can update as needed.

vinaykul: It is only updated when scaleout_proxy is true (first time kube-up is called in workflow that deploys proxy and RP)
start-kubemark does not have access to the variables in invoked bash functions. There is no way to export it back and handle this in caller.

yb01 (approved):
/lgtm

PR #840, opened 2020-12-02
Pick up performance test changes

Sindica:


cluster/gce/manifests/etcd.manifest
```diff
@@ -45,7 +45,7 @@
         "value": "{{ etcd_protocol }}://{{ hostname }}:{{ server_port }}"
       },
       { "name": "ETCD_SNAPSHOT_COUNT",
-        "value": "10000"
+        "value": "100000"
       }
```
sonyafenge, line 49: do we really need this change? I checked community still using "10000" as value.

cluster/common.sh
```diff
@@ -1132,10 +1132,8 @@ function registry-authentication() {
 function create-and-upload-etcd-image() {
 
   ETCD_IMAGE_REGISTRY="${ETCD_IMAGE_REGISTRY:-${CONTAINER_REGISTRY}/${PROJECT}}"
-
   # Build+push the image through makefile.
   MAKE_DIR="${KUBE_ROOT}/cluster/images/etcd"
-
```
sonyafenge, line 1138: this commit already included in master.

cluster/gce/gci/configure-helper.sh
```diff
@@ -130,7 +130,7 @@ function main() {
     start-kube-controller-manager
     start-kube-scheduler
     wait-till-apiserver-ready
-    start-workload-controller-manager
+   # start-workload-controller-manager
```
yb01, line 133: can we use this variable to control it as the partitionserver-configure-helper.sh does ?
{ENABLE_WORKLOADCONTROLLER}

yb01 (approved):
/lgtm

cluster/gce/manifests/etcd.manifest
Sindica: In that case we should use 10K in arktos run. I noticed recent 1.18 test used 10K. Please check with the team.

cluster/gce/gci/configure-helper.sh
Sindica: The issue is that we need additional work to disable rs controller and deployment controller in KCM. This PR is trying to synchronize perf test and master, would rather leave enhancement in separate changes.

cluster/common.sh
Sindica: Can you specify the commit comment?

cluster/gce/gci/configure-helper.sh
sonyafenge: we do have this parameter, but in this time, it only works for partition servers and it doesn't work for master server. I agree with Ying, for now, we can keep this way.

cluster/common.sh
sonyafenge: included in this PR: https://github.com/CentaurusInfra/arktos/pull/796

PR #842, opened 2020-12-03
backwards compatible node controller

chenqianfzh:
Make node-controller backwards compatible. When "tenant-servers" in the CLI option is empty, use the apiserver specified in the kubeconfig. 

Verified in gce, the issue that nodelifecycle controller crash issue in kube-up.sh is solved.

### DIFF
--- a/cmd/kube-controller-manager/app/core.go
+++ b/cmd/kube-controller-manager/app/core.go
@@ -129,7 +129,16 @@ func startNodeIpamController(ctx ControllerContext) (http.Handler, bool, error) 
 
 func startNodeLifecycleController(ctx ControllerContext) (http.Handler, bool, error) {
 
-	tpAccessors, err := lifecyclecontroller.GetTenantPartitionManagers(ctx.ComponentConfig.NodeLifecycleController.TenantServers, ctx.Stop)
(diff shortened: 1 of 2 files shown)


cmd/kube-controller-manager/app/core.go
```diff
@@ -141,7 +150,7 @@ func startNodeLifecycleController(ctx ControllerContext) (http.Handler, bool, er
 		ctx.InformerFactory.Core().V1().Nodes(),
 		//ctx.InformerFactory.Apps().V1().DaemonSets(),
 		// node lifecycle controller uses existing cluster role from node-controller
-		ctx.ClientBuilder.ClientOrDie("node-controller"),
+		kubeclient,
```
yb01, line 153: this should not  be change to kubeclient. those are original controller clientset we should keep as it. in addition to that, the kubeclient can be empty with the code above ( if - else block )

yb01 (approved):
/lgtm.

except the comment to not use the newly added kubeclient variable.

cmd/kube-controller-manager/app/core.go
chenqianfzh: That is a good suggestion. I will update it.

yb01 (approved):
/lgtm

PR #851, opened 2020-12-08
statefulset fix and kubeclient switch

pdgetrf:
### Changes

1. fix a bug that causes statefulset to fail for non-system tenant
2. add kubeclient selection based on tenant 

### Validation
- statefulset with hostpath pvc created and deleted correctly on both tenant partitions
- no regression with pod and deployment creation and deletion

### Notes
- TP kubeclient is still duplcated
- add this in tenant partition to allow host path pv: export ENABLE_HOSTPATH_PROVISIONER=true
- need further code on node status update

### DIFF
--- a/pkg/kubelet/volumemanager/reconciler/reconciler.go
+++ b/pkg/kubelet/volumemanager/reconciler/reconciler.go
@@ -91,7 +91,7 @@ type Reconciler interface {
 // mounter - mounter passed in from kubelet, passed down unmount path
 // volumePluginMgr - volume plugin manager passed from kubelet
 func NewReconciler(
-	kubeClient clientset.Interface,
+	kubeClients []clientset.Interface,
 	controllerAttachDetachEnabled bool,
 	loopSleepDuration time.Duration,
 	waitForAttachTimeout time.Duration,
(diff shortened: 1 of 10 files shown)


pkg/kubelet/volumemanager/reconciler/reconciler.go
```diff
@@ -560,7 +560,7 @@ func (rc *reconciler) reconstructVolume(volume podVolume) (*reconstructedVolume,
 
 // updateDevicePath gets the node status to retrieve volume device path information.
 func (rc *reconciler) updateDevicePath(volumesNeedUpdate map[v1.UniqueVolumeName]*reconstructedVolume) {
-	node, fetchErr := rc.kubeClient.CoreV1().Nodes().Get(string(rc.nodeName), metav1.GetOptions{})
+	node, fetchErr := rc.kubeClients[0].CoreV1().Nodes().Get(string(rc.nodeName), metav1.GetOptions{})
```
pdgetrf, line 563: operations for node needs further work as it needs to use the rp client. this doesn't affect the current functionality and will be fixed in later PR.

chenqianfzh (approved):
good enough for poc purpose.

PR #852, opened 2020-12-09
Create POC test tenants with kubemark test

yb01:
In testing. currently we can manually perform this in our test env.


XiaoningDing:
Looks good to me.  /approve

PR #853, opened 2020-12-09
multi-tenancy perf-test

chenqianfzh:
This PR allows the perf tests running under a specific non-system tenant, whose name is specified in the env var SCALEOUT_TEST_TENANT.

Verified in gce kubemark cluster as follows:
1. created a 2T1R kubemark test environment
2.  manually create a tenant named aaa in TP-1
2. Running the following commandline:
```
SCALEOUT_TEST_TENANT=aaa perf-tests/clusterloader2/run-e2e.sh --nodes=100 --provider=kubemark --kubeconfig=/home/cloudshare/go/src/k8s.io/arktos/test/kubemark/resources/kubeconfig.kubemark-tp-1 --report-dir=/home/cloudshare/perf-logs --testconfig=testing/density/config.yaml --testoverrides=./testing/experiments/use_simple_latency_query.yaml
```

I checked:
1. the test completed (though I got some timeout issues in pod deletion, the test itself was running to completion)
2. all the workloads are correctly created under the tenant
3. though we hit timeout issue in some pod deletion, the kubelet logs showed the the pod deletion request was received. So it shows that the perf test sends the pod deletion request as expected.

### DIFF
--- a/perf-tests/clusterloader2/pkg/framework/client/objects.go
+++ b/perf-tests/clusterloader2/pkg/framework/client/objects.go
@@ -186,27 +186,27 @@ func ListNodesWithOptions(c clientset.Interface, listOpts metav1.ListOptions) ([
 }
 
 // CreateNamespace creates a single namespace with given name.
-func CreateNamespace(c clientset.Interface, namespace string) error {
+func CreateNamespace(c clientset.Interface, namespace string, tenant string) error {
 	createFunc := func() error {
-		_, err := c.CoreV1().Namespaces().Create(&apiv1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}})
+		_, err := c.CoreV1().NamespacesWithMultiTenancy(tenant).Create(&apiv1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: namespace}})
 		return err
 	}
 	return RetryWithExponentialBackOff(RetryFunction(createFunc, Allow(apierrs.IsAlreadyExists)))
 }
 
 // DeleteNamespace deletes namespace with given name.
-func DeleteNamespace(c clientset.Interface, namespace string) error {
+func DeleteNamespace(c clientset.Interface, namespace string, tenant string) error {
 	deleteFunc := func() error {
-		return c.CoreV1().Namespaces().Delete(namespace, nil)
+		return c.CoreV1().NamespacesWithMultiTenancy(tenant).Delete(namespace, nil)
 	}
 	return RetryWithExponentialBackOff(RetryFunction(deleteFunc, Allow(apierrs.IsNotFound)))
 }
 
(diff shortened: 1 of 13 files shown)


perf-tests/clusterloader2/pkg/util/multi-tenancy.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"os"
+	"strings"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/klog"
+	apimachineryvalidation "k8s.io/apimachinery/pkg/api/validation"
+)
+
+func GetTenant() string {
```
yb01, line 28: fyi, the two initial test tenant for POC is named as "arktos" and "zeta"

perf-tests/clusterloader2/pkg/framework/client/objects.go
```diff
@@ -186,27 +186,27 @@ func ListNodesWithOptions(c clientset.Interface, listOpts metav1.ListOptions) ([
 }
 
 // CreateNamespace creates a single namespace with given name.
-func CreateNamespace(c clientset.Interface, namespace string) error {
+func CreateNamespace(c clientset.Interface, namespace string, tenant string) error {
```
XiaoningDing, line 189: Can we make the extra parameter "tenant string" to "tenants []string" and do a loop in the callee function? Then it would be able to duplicate workloads to multiple tenants within one test run.

perf-tests/clusterloader2/pkg/util/multi-tenancy.go
XiaoningDing: The specific tenant names won't impact this PR, right?  The test script will fetch whatever tenant name set in environment variable from command line.

perf-tests/clusterloader2/pkg/util/multi-tenancy.go
```diff
@@ -0,0 +1,42 @@
...
+	tenantName := os.Getenv("SCALEOUT_TEST_TENANT")
+    if len(tenantName) == 0 {
```
XiaoningDing, line 30: Maybe parse the content of this environment variable to return []tenants? So we can set "EXPORT SCALEOUT_TEST_TENANTS=tenant1;tenant2;tenant3..."

perf-tests/clusterloader2/pkg/framework/client/objects.go
chenqianfzh: per offline sync, we are going to run multiple copies of perf-test for different tenants.

XiaoningDing: Resolve this based on the discussion. We don't do it for now.

PR #856, opened 2020-12-11
fix kubeclient usage for node access

pdgetrf:
### Changes

Use rp kubeclient for Node related access

### Validation

PV, deployment and pod start and delete correctly

### Known issue

- PV resize does not work (presumably not used in perf test)


pkg/kubelet/volumemanager/volume_manager.go
```diff
@@ -199,6 +200,9 @@ func NewVolumeManager(
 
 // volumeManager implements the VolumeManager interface
 type volumeManager struct {
+	// kubeClient is the kube API client used by DesiredStateOfWorldPopulator to
+	// communicate with the API server to fetch PV and PVC objects
+	rpKubeClient clientset.Interface
```
yb01, line 205: let's be consistent with the clients naming in kubelet, heartbeatClient likely a better variable name to align

pdgetrf: I think heartbeatClient is a confusing variable name to start with actually for the resource partition. This client is used to talk to the resource partition apiserver, not just for heartbeat. Naming it heartbeatClient can be misleading.

pdgetrf:
@XiaoningDing this is ready for merge

pkg/kubelet/volumemanager/volume_manager.go
yb01: as long as they are consistent, we should be fine. cross the kubelet, the heartbeat and the other kubeclient ( original one ) is used.

PR #874, opened 2020-12-16
Disable APIServerConfigManager and run update-gofmt

XiaoningDing:
**What type of PR is this?**
> /kind feature

Disable the APIServerConfigManager of API Server partition.  This feature isn't used in PoC and it might cause side effects.
Comment it out instead of deleting in case we need it later.  (One of the occurrences is deletion because it's a merge error from code sync.)


yb01 (approved):
/lgtm

PR #878, opened 2020-12-16
NO-THROTTLE-APISERVER in kubemark

chenqianfzh:
For the convenience of the scalability test, when ${NO-THROTTLE-APISERVER}==true, the constraints on max-requests-inflightmax and mutating-requests-inflight are removed.


cluster/gce/gci/configure-helper-common.sh
```diff
@@ -1704,17 +1704,22 @@ function start-kube-apiserver {
   if [[ -n "${ENABLE_GARBAGE_COLLECTOR:-}" ]]; then
     params+=" --enable-garbage-collector=${ENABLE_GARBAGE_COLLECTOR}"
   fi
-  if [[ -n "${NUM_NODES:-}" ]]; then
-    # If the cluster is large, increase max-requests-inflight limit in apiserver.
-    if [[ "${NUM_NODES}" -gt 3000 ]]; then
-      params+=" --max-requests-inflight=3000 --max-mutating-requests-inflight=1000"
-    elif [[ "${NUM_NODES}" -gt 500 ]]; then
-      params+=" --max-requests-inflight=1500 --max-mutating-requests-inflight=500"
+  if [[ -n "${NO_THROTTLE_APISERVER:-}" == "true" ]]; then
```
XiaoningDing, line 1707: Why do we need this change?  I suggest keeping the original code so that we have a comparable configuration with regular Arktos tests

yb01 (approved):
/lgtm

PR #879, opened 2020-12-16
enable prometheus for scale-out 

pdgetrf:
### Changes 
add changes from the following commits to enable prometheus in scale-out cluster

https://github.com/futurewei-cloud/arktos-perftest/commit/e5f6915b1603e29fc76f6ec6af3739573165ddd8
https://github.com/futurewei-cloud/arktos-perftest/commit/c808cf317b2a0d6b16a672f1cfe81709c3c54a2b

### DIFF
--- a/cluster/gce/gci/configure-helper.sh
+++ b/cluster/gce/gci/configure-helper.sh
@@ -156,6 +156,7 @@ function main() {
   reset-motd
   prepare-mounter-rootfs
   modprobe configs
+  start-prometheus &  #####start prometheus
   echo "Done for the configuration for kubernetes"
 }
 
--- a/cluster/gce/gci/configure-helper-common.sh
+++ b/cluster/gce/gci/configure-helper-common.sh
@@ -1401,6 +1401,27 @@ function start-kube-proxy {
   cp "${src_file}" /etc/kubernetes/manifests
 }
 
+# Starts prometheus.
+function start-prometheus {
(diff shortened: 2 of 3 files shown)


pdgetrf:
@sonyafenge other than the changes here, is there anything else we need to do to enable prometheus when starting the cluster?

sonyafenge:
> @sonyafenge other than the changes here, is there anything else we need to do to enable prometheus when starting the cluster? for example, env variables like PROMETHEUS_SCRAPE_ETCD=true ENABLE_PROMETHEUS_SERVER=true

no, we don't need any env variables because this is a temporally solution without switch variables.  ENABLE_PROMETHEUS_SERVER=true is the one supported by community and arktos code, I have a quick check and cannot get this one access outside of cluster.

cluster/gce/gci/configure-helper.sh
```diff
@@ -156,6 +156,7 @@ function main() {
   reset-motd
   prepare-mounter-rootfs
   modprobe configs
+  start-prometheus &  #####start prometheus
```
sonyafenge, line 159: this one only for main master, if you have partitioned server, you need double check if your partitioned server also has this started.

sonyafenge (approved):
LGTM

XiaoningDing:
Remove the "do not merge" label as Qian confirmed the verification is done.

PR #885, opened 2020-12-23
fix controller QPS issue

chenqianfzh:
Fixing the issue that replica-set controller was set to 8000. It should be caused by shared config which was multiplied by 20 by the tenant controller and namespace controller.

### DIFF
--- a/cmd/kube-controller-manager/app/core.go
+++ b/cmd/kube-controller-manager/app/core.go
@@ -353,16 +353,33 @@ func startResourceQuotaController(ctx ControllerContext) (http.Handler, bool, er
 	return nil, true, nil
 }
 
+func increaseControllerRateLimit(config *restclient.Config, QPSCoefficient float32, BurstCoefficient float32) *restclient.Config {
+	newConfig := restclient.NewAggregatedConfig()
+	for _, kubeConfig := range config.GetAllConfigs() {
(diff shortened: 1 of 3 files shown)


cmd/kube-controller-manager/app/core.go
```diff
@@ -371,13 +388,16 @@ func startTenantController(ctx ControllerContext) (http.Handler, bool, error) {
...
-		tnKubeConfig.Burst *= 100
-	}
+	tnKubeConfigs := increaseControllerRateLimit(ctx.ClientBuilder.ConfigOrDie("tenant-controller"), 20, 100)
 	tenantKubeClient := clientset.NewForConfigOrDie(tnKubeConfigs)
 
+	//add some logs to track the rate limit setting 
+	if tenantKubeClient != nil && tenantKubeClient.CoreV1().RESTClient().GetRateLimiter() != nil {
+		klog.V(2).Infof("starting tenant controller: RateLimiter QPS: %v ", tenantKubeClient.CoreV1().RESTClient().GetRateLimiter().QPS())
+	} else {
+		klog.V(2).Infof("starting tenant controller: RateLimiter is nil ")
```
yb01, line 398: Ratelimiter nil should be treated as an error case or warning at least.

yb01 (approved):
/lgtm

PR #896, opened 2020-12-28
fix haproxy cfg & script to work in one-box

chenqianfzh:
Recent changes to haproxy.cfg to add Prometheus monitoring to proxy breaks the one-box haproxy setup. This PR fixes it by removing those lines when generating the haproxy.cfg.

Also included in this PR:
1. increase the number of processes of haproxy from 32 to 64 ( the max value allowed) as we have very powerful machines in tests.
2. code refactoring

Verification:
1. verified in my local env that it works.
2. Verified that the clusters were built after integrating with the Haproxy Prometheus fix (https://github.com/CentaurusInfra/arktos/pull/895)

### DIFF
--- a/hack/scale_out_poc/haproxy_2T1R/haproxy.cfg
+++ b/hack/scale_out_poc/haproxy_2T1R/haproxy.cfg
@@ -20,40 +20,9 @@ global
         #  https://mozilla.github.io/server-side-tls/ssl-config-generator/?server=haproxy
         ssl-default-bind-ciphers ECDH+AESGCM:DH+AESGCM:ECDH+AES256:DH+AES256:ECDH+AES128:DH+AES:RSA+AESGCM:RSA+AES:!aNULL:!MD5:!DSS
         ssl-default-bind-options no-sslv3
-        nbproc 32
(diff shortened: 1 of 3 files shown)


hack/scale_out_poc/haproxy_2T1R/haproxy.cfg
```diff
@@ -74,14 +43,13 @@ defaults
...
 
-# the following setup enable the prometheus export
-frontend stats
-    bind *:8404
-    option http-use-htx
-    http-request use-service prometheus-exporter if { path /metrics }
-    stats enable
-    stats uri /stats
-    stats refresh 10s
+KUBEMARK_ONLY:frontend stats
```
pdgetrf, line 46: I will later have a change to enable this even for one-box environment.

pdgetrf: here actually https://github.com/pdgetrf/arktos/commit/9956e2c7a00abad918745da5bb875eaaf8cb782e

but okay to check in with your change for now.

PR #897, opened 2020-12-29
Extend kubelet to use kubeclient from more than 2 tenant partitions

pdgetrf:
### Changes

- kubelet supports using kube clients from more than 2 tenant partitions
- remove duplication of GetTPClient (finally!)

### Validation

cannot test more than 2 tenants though without proxy change.

- one box test with 2 tenants (done)
- 100-node test with arktos tenant (done)

### DIFF
--- a/pkg/kubelet/config/config.go
+++ b/pkg/kubelet/config/config.go
@@ -19,7 +19,10 @@ package config
 
 import (
 	"fmt"
+	"k8s.io/kubernetes/pkg/kubelet/util"
 	"reflect"
+	"strconv"
+	"strings"
 	"sync"
 
 	"k8s.io/api/core/v1"
@@ -260,6 +263,14 @@ func (s *podStorage) merge(source string, change interface{}) (adds, updates, de
 			if ref.Annotations == nil {
 				ref.Annotations = make(map[string]string)
 			}
+
+			if strings.HasPrefix(source, kubetypes.ApiserverSource) {
+				if _, ok := util.Tenant2api[ref.Tenant]; !ok {
+					clientId, _ := strconv.Atoi(source[(len(source) - 1):])
+					util.Tenant2api[ref.Tenant] = clientId
+				}
+			}
+
 			ref.Annotations[kubetypes.ConfigSourceAnnotationKey] = source
 			if existing, found := oldPods[ref.UID]; found {
 				pods[ref.UID] = existing
--- a/pkg/kubelet/config/config.go
+++ b/pkg/kubelet/config/config.go
@@ -21,8 +21,6 @@ import (
 	"fmt"
 	"k8s.io/kubernetes/pkg/kubelet/util"
 	"reflect"
-	"strconv"
-	"strings"
 	"sync"
 
 	"k8s.io/api/core/v1"
@@ -264,12 +262,8 @@ func (s *podStorage) merge(source string, change interface{}) (adds, updates, de
 				ref.Annotations = make(map[string]string)
 			}
 
-			if strings.HasPrefix(source, kubetypes.ApiserverSource) {
-				if _, ok := util.Tenant2api[ref.Tenant]; !ok {
-					clientId, _ := strconv.Atoi(source[(len(source) - 1):])
-					util.Tenant2api[ref.Tenant] = clientId
-				}
-			}
+			// Bookkeeping mapping between tenant and its origin apiserver
+			util.RegisterTenantSourceServer(source, ref)
 
 			ref.Annotations[kubetypes.ConfigSourceAnnotationKey] = source
 			if existing, found := oldPods[ref.UID]; found {
--- a/pkg/kubelet/util/util.go
+++ b/pkg/kubelet/util/util.go
@@ -1,5 +1,6 @@
 /*
 Copyright 2017 The Kubernetes Authors.
+Copyright 2020 Authors of Arktos - file modified.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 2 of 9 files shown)


pdgetrf:
@XiaoningDing please take a preliminary look and see if this matches what you have in mind. I'm doing a few more tests and let's check this in tomorrow.

pkg/kubelet/config/config.go
```diff
@@ -260,6 +263,14 @@ func (s *podStorage) merge(source string, change interface{}) (adds, updates, de
 			if ref.Annotations == nil {
 				ref.Annotations = make(map[string]string)
 			}
+
+			if strings.HasPrefix(source, kubetypes.ApiserverSource) {
+				if _, ok := util.Tenant2api[ref.Tenant]; !ok {
+					clientId, _ := strconv.Atoi(source[(len(source) - 1):])
+					util.Tenant2api[ref.Tenant] = clientId
```
XiaoningDing, line 270: I think you still need a lock (maybe a RWMutex?) even you have only one writing thread. If I remember correctly Golang runtime will complain if it detects read/write from different goroutines.

pkg/kubelet/util/util.go
```diff
@@ -18,8 +19,25 @@ package util
...
+	"k8s.io/klog"
 )
 
+// poc, tenant name -> api server id
+var Tenant2api map[string]int
+
+func GetTPClient(kubeClients []clientset.Interface, tenant string) clientset.Interface {
+	var client clientset.Interface
+	pick, ok := Tenant2api[tenant]
+	if len(kubeClients) == 1 || tenant[0] <= 'm' || !ok { // !ok to force a retry
```
XiaoningDing, line 32: Why do we still need "tenant[0] <= 'm'" here?

pdgetrf: removed

pkg/kubelet/config/config.go
XiaoningDing: Also:  1) Not sure how we handle system tenants in kubelet now. Do we need to skip empty string or "system" here?  2) Probably use "toLower" both here and the lookup function to ensure case insensitive?

pdgetrf: based on this logic I wonder if there's a single-thread place where this map can be filled instead of here to avoid the need of a lock

yb01: regardless, the delete of map is unavoidable in the future, having the lock in place might be eventually needed anyway so I suggest to have it.

pkg/kubelet/util/util.go
```diff
@@ -18,8 +19,25 @@ package util
...
 import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/klog"
 )
 
+// poc, tenant name -> api server id
+var Tenant2api map[string]int
+
+func GetTPClient(kubeClients []clientset.Interface, tenant string) clientset.Interface {
```
yb01, line 29: it might be get the code more organized to provide the add, get( which is here already), delete OP functions to the map here, so the other functions just call them to manipulate the map

yb01 (approved):
/lgtm. overall looks good.

pkg/kubelet/config/config.go
pdgetrf: > I think you still need a lock (maybe a RWMutex?) even you have only one writing thread. If I remember correctly Golang runtime will complain if it detects read/write from different goroutines.

it seems the lock here is already doing what is needed for this

https://github.com/CentaurusInfra/arktos/blob/0167ab88f8503f1359845a8ef68a88e8a16b755f/pkg/kubelet/config/config.go#L173

pdgetrf: > Also: 1) Not sure how we handle system tenants in kubelet now. Do we need to skip empty string or "system" here? 2) Probably use "toLower" both here and the lookup function to ensure case insensitive?

1. "system" tenant should still follow the proxy rules. I don't think we allow empty tenant but I'm double checking this.

2. Added the conversion. Although I don't think it has been defined whether tenant name is case sensitive or not.

pdgetrf: Updates: 

- Add RW lock
- "system" tenant from a certain tenant partition should communicate with the origin tp apiserver
- Tenant by definition should be lower case already. Add toLower to be safe
- Skip empty tenant
- Refactor

pkg/kubelet/util/util.go
pdgetrf: good point. did some refactor.

pdgetrf:
this is good for final review now.

pkg/kubelet/util/util.go
```diff
@@ -17,9 +18,52 @@ limitations under the License.
...
+		return
+	}
+	key := strings.ToLower(ref.Tenant)
+	clientId, err := strconv.Atoi(source[(len(source) - 1):])
+	if err != nil {
+		klog.Errorf("Failed to get apiserver id. Err: %s", err)
+		return
+	}
+	Tenant2apiLock.Lock()
+	Tenant2apiLock.Unlock()
```
XiaoningDing, line 47: What's this? Did you miss "defer"?

pkg/kubelet/util/util.go
```diff
@@ -17,9 +18,52 @@ limitations under the License.
...
+	if _, ok := Tenant2api[key]; !ok {
+		Tenant2api[key] = clientId
+	}
+}
+
+func GetTPClient(kubeClients []clientset.Interface, tenant string) clientset.Interface {
+	Tenant2apiLock.RLock()
+	Tenant2apiLock.RUnlock()
```
XiaoningDing, line 55: Same here. defer RUnlock()?

PR #899, opened 2020-12-29
enlarge net.netfilter.nf_conntrack_max when building scale-out masters

chenqianfzh:
Verified.

With this change, "cat /proc/sys/net/netfilter/nf_conntrack_max" returned 26214400 in Tp-1/Tp-2/RP machines.


yb01 (approved):
/lgtm

PR #907, opened 2021-01-05
Debug logging in kubelet and scheduler

yb01:
Update some key logging entries to level 2 so that they can be visible in perf run with default logging level settings.


chenqianfzh (approved):

vinaykul:
Looks good. Let's add a couple of more statements in 
func (cache *schedulerCache) AddPod(pod *v1.Pod) and RemovePod(pod *v1.Pod) functions. I believe this is when the watch adds a new pod to scheduler cache, and can help determine how long it takes a pod to be scheduled after being seem by scheduler.

PR #908, opened 2021-01-05
fix onebox haproxy cfg script

chenqianfzh:
Fix onebox cfg script.

Verifcation done in my dev box.

### DIFF
--- a/hack/scale_out_poc/config_haproxy/setup_haproxy.sh
+++ b/hack/scale_out_poc/config_haproxy/setup_haproxy.sh
@@ -72,7 +72,11 @@ config_haproxy() {
 
         template_source=${repo_root}/hack/scale_out_poc/config_haproxy/haproxy.cfg.template
 
-        go build -o /tmp/haproxy_cfg_generator "${script_root}/cfg_generator/"
+        pushd .
+
+        cd ${script_root} && go build -o /tmp/haproxy_cfg_generator "./cfg_generator/"
+
+        popd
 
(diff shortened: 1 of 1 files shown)


hack/scale_out_poc/config_haproxy/setup_haproxy.sh
```diff
@@ -72,7 +72,11 @@ config_haproxy() {
 
         template_source=${repo_root}/hack/scale_out_poc/config_haproxy/haproxy.cfg.template
 
-        go build -o /tmp/haproxy_cfg_generator "${script_root}/cfg_generator/"
+        pushd .
```
pdgetrf, line 75: don't you mean pushd ${script_root} and remove the cd command from line 77?

chenqianfzh: I mean to push the working directory of the user and go back to that, for a better user experience. This "go build" command will hit some dependency issue if the current directory is not ${script_root}.

pdgetrf: understood the intent. 

pushd will go to the target folder and popd will return to the original directory


```bash
pushd ${script_root}  # this save the current directory to the stack and go to ${script_root}
go build -o /tmp/haproxy_cfg_generator "./cfg_generator/" # this is executed in ${script_root}
popd # this leaves ${script_root} and returns to the original folder
```

@XiaoningDing this is probably different ways that will achieve the same goal. I'm okay this is merged as is.

PR #906, opened 2021-01-05
separate and direct haproxy and add scheduler & controller prometheus

pdgetrf:
### Changes: 
- make haproxy emit log only to /var/log/haproxy.log in **one-box environment**. 
- build the same haproxy with prometheus in one-box environment
- this changes apply to 2-tp 1-rp setup script. a second PR will follow to make the same changes in the N tp and 1 rp script that is also used for GCE setup
- enable prometheus scraping to scheduler 
- enable prometheus scraping to controller

### Why this change?
- prior to this change, haproxy log is either flooding /var/log/syslog, or writing to both /var/log/syslog and /var/log/haproxy.log
- consistent haproxy version with the one used in GCE
- help debugging and analysis in one-box

### Notes
- change is based on https://pc-freak.net/blog/enable-haproxy-logging-separate-log-varloghaproxylog-prevent-haproxy-duplicate-messages-varlogmessages/
- metrics info in https://www.sumologic.com/blog/kubernetes-monitoring/

### How tested?
- local one box (2tp-1rp) envirment. verified haproxy log only shows up in /var/log/haproxy.log, not in /var/log/syslog
- scheduler metrics in one-box testing:
image

- metrics shows up in GCE 100-node env
image

PR #911, opened 2021-01-06
make haproxy dump logs only to /var/log/haproxy.log (part 2)

pdgetrf:
### Changes
This is the 2nd part of changes for haproxy logging after https://github.com/CentaurusInfra/arktos/pull/906.

### How Validated?
- one box: cluster start and sanity test runs correctly. haproxy log shows up only in /var/log/haproxy.log
- GCE: 100-node density test passes. logs correct.
image

### DIFF
--- a/cluster/gce/util.sh
+++ b/cluster/gce/util.sh
@@ -2846,6 +2846,7 @@ function setup-proxy() {
   ssh-to-node ${PROXY_NAME} "sudo apt install -y ${KUBERNETES_SCALEOUT_PROXY_APP}"
 
   patch-haproxy-prometheus
+  direct-haproxy-logging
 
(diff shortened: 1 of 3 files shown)


cluster/gce/util.sh
```diff
@@ -2867,6 +2868,38 @@ function patch-haproxy-prometheus {
   ssh-to-node ${PROXY_NAME} "haproxy -vv|grep Prometheus"
 }
 
+function direct-haproxy-logging {
+  set -x
```
pdgetrf, line 2872: this line will be removed once GCE validation is done

hack/scale_out_poc/config_haproxy/setup_haproxy.sh
```diff
@@ -96,6 +138,10 @@ fi
 
 install_haproxy_if_needed
 
+patch-haproxy-prometheus
+
+direct-haproxy-logging
```
pdgetrf, line 143: having this two make one-box uses exactly the same haproxy as GCE does

PR #912, opened 2021-01-07
Settings of max-request-inflight and max-mutating-request-inflight, cap api server mem usage for 600G

yb01:
In Testing.

without parameters set, the default is taken in the 100 node test
```
--max-requests-inflight=1500 --max-mutating-requests-inflight=500
```

with parameters set, check api server process with parameters:
```
--max-requests-inflight=1200 --max-mutating-requests-inflight=200
```


cluster/gce/gci/configure-helper-common.sh
```diff
@@ -1729,17 +1729,44 @@ function start-kube-apiserver {
...
   if [[ -n "${NUM_NODES:-}" ]]; then
+    local max_request_inflight
+    local max_mutating_request_inflight
+
     # If the cluster is large, increase max-requests-inflight limit in apiserver.
-    if [[ "${NUM_NODES}" -gt 3000 ]]; then
-      params+=" --max-requests-inflight=3000 --max-mutating-requests-inflight=1000"
-    elif [[ "${NUM_NODES}" -gt 500 ]]; then
-      params+=" --max-requests-inflight=1500 --max-mutating-requests-inflight=500"
+    if [[ "${NUM_NODES}" -gt 10000 ]]; then
```
XiaoningDing, line 1738: Nit: this structure is not as clear as the original one. It can potentially miss a case between different "if" combinations, while it won't happen in the original chained "elif" style.  But given we are on a PoC branch I guess we don't want to spend too much time on this :)

yb01: previously the logic doesn't set the parameters if nodes is less than 500. i changed the code to keep the previous logic and update the overwrite code with ENV VAR below.

yb01: pushed the changes.

PR #919, opened 2021-01-11
fix static pod's invalid uuid issue

chenqianfzh:
We saw a lot of messages like "Got invalid uuid [XXXXXXXXXXXXXXXXXXXXXXXXX]. Assign a new one [XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXX]" in the kube-apiserver.log in TP master. This PR fixes this issue.

#### Root cause:
The root cause is the static pod's UID is in the formaet of XXXXXXXXXXXXXXXXXXXXXXXXX, while the api server expects a UID in the format of XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXX.

#### Verification:
Created a 100-node cluster and finished a density test successfully. No message of "Got invalid uuid" was seen in the kube-apiserver.log in TP-1 master.

#### more
this change also fixed the issue that static pods restarts frequently (once per minute or more). As show below,  the static pods were not restarted since being created.

image

### DIFF
--- a/pkg/kubelet/config/common.go
+++ b/pkg/kubelet/config/common.go
@@ -20,7 +20,6 @@ package config
 
 import (
 	"crypto/md5"
-	"encoding/hex"
 	"fmt"
 	"strings"
 
@@ -59,7 +58,8 @@ func applyDefaults(pod *api.Pod, source string, isFile bool, nodeName types.Node
 			fmt.Fprintf(hasher, "url:%s", source)
 		}
 		hash.DeepHashObject(hasher, pod)
-		pod.UID = types.UID(hex.EncodeToString(hasher.Sum(nil)[0:]))
+		hexVlaue := hasher.Sum(nil)[0:]
+		pod.UID = types.UID(fmt.Sprintf("%x-%x-%x-%x-%x", hexVlaue[0:4], hexVlaue[4:6], hexVlaue[6:8], hexVlaue[8:10], hexVlaue[10:]))
 		klog.V(5).Infof("Generated UID %q pod %q from %s", pod.UID, pod.Name, source)
 	}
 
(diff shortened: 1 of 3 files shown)


pkg/kubelet/config/common.go
```diff
@@ -59,7 +58,8 @@ func applyDefaults(pod *api.Pod, source string, isFile bool, nodeName types.Node
 			fmt.Fprintf(hasher, "url:%s", source)
 		}
 		hash.DeepHashObject(hasher, pod)
-		pod.UID = types.UID(hex.EncodeToString(hasher.Sum(nil)[0:]))
+		hexVlaue := hasher.Sum(nil)[0:]
```
yb01, line 61: this function is commonly used. any reason this only occurs to Arktos ?

Sindica: Arktos made changes to ./staging/src/k8s.io/apiserver/pkg/registry/rest/meta.go and reused meta.uid instead of generate a new one (Alkaid always generate new uid)

chenqianfzh: I think Sindica's comment makes sense.

yb01 (approved):
/lgtm

pkg/kubelet/config/common.go
XiaoningDing: According to the code logics in this function, multiple static pods will have same UID if they are on the same node.  Will this be a problem?  Shall we just call GUID generation function to generate a unique UID?

(Previous this probably isn't a problem because API server will replace the UID, as Ying mentioned in the comment)

XiaoningDing: I think more discussion is needed here:  1) Shall we keep the original behavior that API Server re-generates UID?  2) If we don't keep the original behavior and rely on client side to generate guid, shall we just call GUID generation function here?  I can see there is a problem here, multiple static pods will have same UID if they are on the same node.

PR #920, opened 2021-01-12
Partial fix for kube-up & start-kubemark for non-scaleout cluster

vinaykul:
**What type of PR is this?**
>
> /kind api-change
kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**: This helps verify that max open files configuration for haproxy is correctly configured for all processes.

**Which issue(s) this PR fixes**: Perf cluster deployment (kube-up + start-kubemark) has regressed in scale-out-poc branch for non scale-out architecture (scale-up). This partially fixes the regression. Currently, the non scaleout cluster fails in start-kubemark due to kubemark binary unable start if tenant & resource server params are missing. The binary should start in scale-up mode if that were the case. This will be a separate PR.

I tested this by doing kube-up and start-kubemark for both scaleout & non-scaleout arch. It comes up successfully for scaleout, so it is not regressing current work.

Fixes #

```release-note

```

### DIFF
--- a/test/kubemark/start-kubemark.sh
+++ b/test/kubemark/start-kubemark.sh
@@ -43,14 +43,19 @@ LOCAL_KUBECONFIG_TMP="${RESOURCE_DIRECTORY}/kubeconfig.kubemark.tmp"
 
 export SCALEOUT_TP_COUNT="${SCALEOUT_TP_COUNT:-1}"
 
-TP_KUBECONFIG="${RESOURCE_DIRECTORY}/kubeconfig.kubemark-tp"
-
 ### files to explicitly save the kubeconfig to different cluster or proxy
 ### those are used for targeted operations such as tenant creation etc on
 ### desired cluster directly
 ###
+TP_KUBECONFIG="${RESOURCE_DIRECTORY}/kubeconfig.kubemark-tp"
 PROXY_KUBECONFIG_SAVED="${RESOURCE_DIRECTORY}/kubeconfig.kubemark.proxy.saved"
 RP_KUBECONFIG_SAVED="${RESOURCE_DIRECTORY}/kubeconfig.kubemark.rp.saved"
+if [[ "${SCALEOUT_CLUSTER:-false}" == "false" ]]; then
+  TP_KUBECONFIG="${LOCAL_KUBECONFIG}"
+  LOCAL_KUBECONFIG_TMP="${LOCAL_KUBECONFIG}"
+  RP_KUBECONFIG_SAVED="${LOCAL_KUBECONFIG}"
+fi
+
 
 ### POC tenant yaml files
 ###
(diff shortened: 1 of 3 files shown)


pdgetrf:
please add details in PR description with the following info:

1. what was the problem that's being fixed
2. how was the fix verified

vinaykul:
> please add details in PR description with the following info:
> 
> 1. what was the problem that's being fixed
> 2. how was the fix verified

Updated the first section.

pdgetrf (approved):

test/kubemark/resources/hollow-node_template.yaml
```diff
@@ -50,20 +50,10 @@ spec:
...
-        - name: RESOURCE_SERVER
-          valueFrom:
-            configMapKeyRef:
-              name: node-configmap
-              key: resource.server
         command:
         - /bin/sh
         - -c
-        - /kubemark --morph=kubelet --tenant-servers={{tenant_servers}} --resource-server={{resource_server}} --name=$(NODE_NAME) {{hollow_kubelet_params}} --kubeconfig=/kubeconfig/kubelet.kubeconfig $(CONTENT_TYPE) --alsologtostderr 1>>/var/log/kubelet-$(NODE_NAME).log 2>&1
+        - /kubemark --morph=kubelet --name=$(NODE_NAME) {{hollow_kubelet_params}} --kubeconfig=/kubeconfig/kubelet.kubeconfig $(CONTENT_TYPE) --alsologtostderr 1>>/var/log/kubelet-$(NODE_NAME).log 2>&1
```
yb01, line 56: i think the --tenant-servers arg should be OK to keep it as it for both scale out and non-sccale-out clusters. the Arktos kubelet is changed to handle backcompt as well for this arg.

test/kubemark/start-kubemark.sh
```diff
@@ -184,7 +189,11 @@ function create-kube-hollow-node-resources {
   # Create the replication controller for hollow-nodes.
   # We allow to override the NUM_REPLICAS when running Cluster Autoscaler.
   NUM_REPLICAS=${NUM_REPLICAS:-${NUM_NODES}}
-  sed "s@{{numreplicas}}@${NUM_REPLICAS}@g" "${RESOURCE_DIRECTORY}/hollow-node_template.yaml" > "${RESOURCE_DIRECTORY}/hollow-node.yaml"
+  if [[ "${SCALEOUT_CLUSTER:-false}" == "true" ]]; then
+    sed "s@{{numreplicas}}@${NUM_REPLICAS}@g" "${RESOURCE_DIRECTORY}/hollow-node_template_scaleout.yaml" > "${RESOURCE_DIRECTORY}/hollow-node.yaml"
```
yb01, line 193: it looks the diffs from the yaml files for scaleout an non-scaleout is not much, should we maintain one yaml file for them instead of creating two? consider this will be eventually into the main branch ? i am ok with two, just something to make sure we evaluated the long term pros-cons

yb01 (approved):
/lgtm

chenqianfzh (approved):

test/kubemark/resources/hollow-node_template.yaml
vinaykul: ok, i'll try adding it.

vinaykul: i looked into the kubeconfig for scaleup, and the server address is correct. something change did in the kubemark in poc branch regressed it. let's fix it there so that we don't have to pass the tenant parameter for scale up kubemark, where we use defaults.

test/kubemark/start-kubemark.sh
vinaykul: let's do it when we fix the kubemark regression.

vinaykul:
Needs /approve

PR #927, opened 2021-01-15
migrate haproxy config tool/script to master branch

chenqianfzh:
This PR migrates the HaProxy config tool/script from PoC branch to master.

It is verified in one-box.

Verfication with kubemark clusters will be done after the kubemakr scale-out change is also migrated to master.


cmd/haproxy-cfg-generator/cfg_generator.go
```diff
@@ -0,0 +1,228 @@
...
+		os.Exit(1)
+	}
+
+	if net.ParseIP(resourcePartitionIP) == nil {
+		LogError("\nIP address (%v) is invalid.\n", resourcePartitionIP)
+		os.Exit(1)
+	}
+}
+
+func get_tp_request_acl() string {
```
pdgetrf, line 98: some functions in this file with "fancy" logic might need unit tests

chenqianfzh: As discussed, this tool is now test-protected.

Also the "patch haproxy for prometheus" will be activated only when a specific env var is "true".

PR #931, opened 2021-01-19
1.18 improvement kubelet

Sindica:
kubelet: Avoid sending no-op patches
Ensure Kubelet always reports terminating pod container status
This change was perf tested on 1/21 & 1/25


pdgetrf (approved):
/lgtm

PR #933, opened 2021-01-20
migrate the code change about network CRD to master

chenqianfzh:
It was found that turning off the multi-tenancy network CRD feature ( as it has not completed) helps improve the scalability. This PR turns off this feature in master branch. 

This feature was turned on/off per the setting of feature gate value of "MandatoryArktosNetwork".

Verification:
It is verified that by default setting, the kubelet does not send out request like "tenant/aaa/networks/default" to apiserver.

### DIFF
--- a/pkg/kubelet/network/dns/dns.go
+++ b/pkg/kubelet/network/dns/dns.go
@@ -88,6 +88,7 @@ func NewConfigurer(recorder record.EventRecorder, nodeRef *v1.ObjectReference, n
 
 func (c *Configurer) getClusterDNS(pod *v1.Pod) ([]net.IP, error) {
 	// fallback to default cluster DNS for pods on the default network if the feature gate is not enabled
+	// remove the following block of if when feature MandatoryArktosNetwork is set by default.
 	if utilfeature.DefaultFeatureGate.Enabled(features.MandatoryArktosNetwork) == false {
 		return c.clusterDNS, nil
 	}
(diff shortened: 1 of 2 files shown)


chenqianfzh:
@assign @pdgetrf

pdgetrf (commented):
I have little knowledge of how this network CRD works. Please get approval from Hong Wei and whoever is more familiar.

pkg/kubelet/network/dns/dns.go
```diff
@@ -105,16 +110,7 @@ func (c *Configurer) getClusterDNS(pod *v1.Pod) ([]net.IP, error) {
 		klog.Errorf("failed to get network %s/%s: %v", pod.Tenant, networkName, err)
 	}
 
-	if utilfeature.DefaultFeatureGate.Enabled(features.MandatoryArktosNetwork) {
-		return nil, fmt.Errorf("network %s/%s not found", pod.Tenant, networkName)
-	}
-
-	// fallback to default cluster DNS for pods on the default network
```
h-w-chen, line 112: nit: suggest to keep this default network behavior (though it does not matter when the feature gate is disabled by default).

h-w-chen (approved):
/approve

pkg/kubelet/network/dns/dns.go
chenqianfzh: updated per comment.

Issue #923, opened 2021-01-13
update k8s.io/kube-openapi to v0.0.0-20200403204345-e1beb1bd0f35


sonyafenge:
**What happened**:
update k8s.io/kube-openapi to v0.0.0-20200403204345-e1beb1bd0f35 
**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:
https://github.com/kubernetes/kubernetes/pull/80123
https://github.com/kubernetes/kubernetes/pull/89833
**Environment**:
- Arktos version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

sonyafenge:
this is part of upgrade Golang to 1.13.9: https://github.com/CentaurusInfra/arktos/issues/891

sonyafenge:
Update K8s.io/kube-openapi to V3 (v0.0.0-20200403204345) need these main change:
- [ ] Add v1 admission registration types under pkg/apis. Arktos only support v1beta1 now.
- [ ] Add CRD V1types under apiextensions-apiserver/pkg/apis/apiextensions
- [ ] Import metadata client under client-go

after all these changes, there are also need:
- [ ] multi-tenancy change in metadata client.

Sindica:
Import metadata client under client-go is part of https://github.com/CentaurusInfra/arktos/pull/941

PR #935, opened 2021-01-20
1.18 improvement kcm metadataclient

Sindica:
Ported the following changes from 1.18:
. Add a metadata client to client-go that can read PartialObjectMetadata
. Update namespace controller & garbage controller to use the metadata client

Arktos changes:
. Add multi-tenancy changes to metadata client, fix UT & integration test
. Copyright fix

Reason:
. Performance improvement from 1.18: "the metadata client uses protobuf and returns only a subset of object
data (the metadata) which allows operations that act only on objects generically to work much faster. "
. Vendor dependency changes required in https://github.com/CentaurusInfra/arktos/issues/923


chenqianfzh (approved):
I verified the code change with the multi-tenancy e2e test and it passed.

sonyafenge:
This PR is also part of kube-openapi upgrade: #923

sonyafenge (approved):
/LGTM

PR #936, opened 2021-01-20
fix backcompat issue in hollow-node

yb01:
hollow node backcompat support. this follows the same approach as kubelet cmd/app/server.go does.

Tested with latest POC branch with Vinay's PR


vinaykul:
I've verified that this works. @XiaoningDing Please approve and merge it. Thanks.

Issue #790, opened 2020-10-22
Apiserver log: meta.go:43] Got invalid uuid. Assign a new one


sonyafenge:
**What happened**:
run Kube-up and start-kubemark to start kubemark cluster and run perf-test using 100 hollow-node, found there are bunch of logs in kube-apiserver.log
```
I1022 19:29:50.686726       1 meta.go:43] Got invalid uuid [e93772b5f98b2876a87ed4d778f2c0d8]. Assign a new one [f7411db5-236a-44fd-bfad-66d36cff630c]
I1022 19:29:56.684593       1 meta.go:43] Got invalid uuid [af7a600762ecfd06e9c9b2f4017b671d]. Assign a new one [4cb4842c-5ab1-4fbf-a970-9267092fe518]
I1022 19:30:01.684744       1 meta.go:43] Got invalid uuid [073ea5ab787afbf3a8aa6f2479891e32]. Assign a new one [dc2c2017-45c8-4ffd-b634-6921aa25f3f8]
I1022 19:30:07.684730       1 meta.go:43] Got invalid uuid [c14601db3e97c3d138b593756c397f2a]. Assign a new one [cd69d01c-5f11-456f-badf-4f424159bb13]
I1022 19:30:36.687000       1 meta.go:43] Got invalid uuid [35c49dbf2b8b84c31d9d6c992b4d0cc9]. Assign a new one [6c9573d1-73d4-4b8a-b92e-dfba6f54aac6]
I1022 19:30:54.684835       1 meta.go:43] Got invalid uuid [e93772b5f98b2876a87ed4d778f2c0d8]. Assign a new one [7f1f3434-34a8-4935-864d-82594dd94997]
I1022 19:31:16.686634       1 meta.go:43] Got invalid uuid [073ea5ab787afbf3a8aa6f2479891e32]. Assign a new one [e83d2f07-cb58-4823-bb48-a0e88c407e45]
I1022 19:31:17.684864       1 meta.go:43] Got invalid uuid [4dbf627684585dc631b45d2948d750f1]. Assign a new one [397911a8-7127-4b07-8d9f-d42e05fe6ac4]
I1022 19:32:23.686551       1 meta.go:43] Got invalid uuid [af7a600762ecfd06e9c9b2f4017b671d]. Assign a new one [c1aed30c-2a80-49b7-af9f-0342e5bbb251]
[...]

PR #940, opened 2021-01-21
reduce the reset-uuid message log level to 3

chenqianfzh:
Due to https://github.com/CentaurusInfra/arktos/issues/790, there are excessive messages of "invalid uuid" in the apiserver logs of scalability tests. Yet we know this reset-uid operation is not blocking the test. So we lower the log level to have a cleaner log.

THis change has been verified in previous a 10K*2 test run in PoC branch.

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/registry/rest/meta.go
+++ b/staging/src/k8s.io/apiserver/pkg/registry/rest/meta.go
@@ -40,7 +40,7 @@ func FillObjectMetaSystemFields(meta metav1.Object) {
 		if err != nil {
 			uid := uuid.NewUUID()
 			meta.SetUID(uid)
-			klog.Infof("Got invalid uuid [%s]. Assign a new one [%v]", uidStr, uid)
+			klog.V(3).Infof("Got invalid uuid [%s]. Assign a new one [%v]", uidStr, uid)
 		}
 	}
 	uid := meta.GetUID()


Sindica (approved):
LGTM

pdgetrf (approved):
/lgtm

staging/src/k8s.io/apiserver/pkg/registry/rest/meta.go
```diff
@@ -40,7 +40,7 @@ func FillObjectMetaSystemFields(meta metav1.Object) {
 		if err != nil {
 			uid := uuid.NewUUID()
 			meta.SetUID(uid)
-			klog.Infof("Got invalid uuid [%s]. Assign a new one [%v]", uidStr, uid)
+			klog.V(3).Infof("Got invalid uuid [%s]. Assign a new one [%v]", uidStr, uid)
```
yb01, line 43: just to follow up, what is fix to avoid this invalid UUID error?

chenqianfzh: Actually, we have not found the solution yet. 

I tried the following two, none of them working
1. let each static pod create a real uid. Then the hollow nodes fail to get ready.
2.  change the format of static format to the expected. The density test failed as some static pods are not in the right states.

PR #932, opened 2021-01-19
Scaleout node controller to master branch

chenqianfzh:
This PR migrate the node-controller changes from PoC branch to master branch.

With the change, the node controller will talk to RP master about nodes & node - lease, while talking to RP master about the other resources. 

Verification:
It is verified that this code change is backwards compatible. More verification in the scale-out clusters will be done after the other scale-out changes are migrated to master branch.

PR #944, opened 2021-01-22
Migrate node controller

chenqianfzh:
the code change of node controller was migrated to master in https://github.com/CentaurusInfra/arktos/pull/932. Yet it was found that the node will be unschedulable when the onebox cluster is started by arktos-up.sh. So this PR 932 was reverted.

This PR has the issue fixed and try to migrate the node-controller change to master again.

Verfication:
Manual tests: the nodes is schedulable and deployments can be created successfully.
The code change is also verified with e2e test test/e2e/arktos/multi_tenancy/run-e2e-test.sh.


pdgetrf (approved):
/lgtm

PR #953, opened 2021-01-26
fix QPS issue in insecure mode

chenqianfzh:
We hit the issue in PoC branch where the QPS of some controllers is abnormally high, which is caused by different controllers share the same KubeConfig object. A workaround was provided in https://github.com/CentaurusInfra/arktos/pull/885.

Yet later it was found this QPS issue only happens when the controller-manager is started with "--use-serviceaccount-credentials" option turned off. The root cause is that all the controllers share the  same kubeconfig object in insecure mode. 

The workaround we had in PoC branch is an overkill as it also touches the code path of secure mode. This PR gives a clean fix by letting all the controllers, in **insecure** mode, own its own KubeConfig object. 

### Verification:
In PoC branch, I reverted the previous workaround and applied this fix, then printed out the KubeConfig objects used by tenant-controller/namespace-controller/dynamic-informer. As the screenshot show, they point to different objects.

image

### DIFF
--- a/pkg/cloudfabric-controller/client_builder.go
+++ b/pkg/cloudfabric-controller/client_builder.go
@@ -58,8 +58,8 @@ type SimpleControllerClientBuilder struct {
 }
 
 func (b SimpleControllerClientBuilder) Config(name string) (*restclient.Config, error) {
-	clientConfig := *b.ClientConfig
-	return restclient.AddUserAgent(&clientConfig, name), nil
+	clientConfig := restclient.CopyConfigs(b.ClientConfig)
+	return restclient.AddUserAgent(clientConfig, name), nil
 }
 
 func (b SimpleControllerClientBuilder) ConfigOrDie(name string) *restclient.Config {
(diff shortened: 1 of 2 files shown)


pkg/cloudfabric-controller/client_builder.go
```diff
@@ -58,8 +58,8 @@ type SimpleControllerClientBuilder struct {
 }
 
 func (b SimpleControllerClientBuilder) Config(name string) (*restclient.Config, error) {
-	clientConfig := *b.ClientConfig
-	return restclient.AddUserAgent(&clientConfig, name), nil
+	clientConfig := restclient.CopyConfigs(b.ClientConfig)
```
yb01, line 61: will this only be used by the insecure mode ? if not, likely it will end up some unnecessary copies of config  objects.

chenqianfzh: Actually, SimpleControllerClientBuilder is a struct that will be used by insecure mode only. Secure mode uses SAControllerClientBuilder. See func Run() in cmd/kube-controller-manager/app/controllermanager.go.

yb01 (approved):
/lgtm

pkg/cloudfabric-controller/client_builder.go
yb01: good to know.

PR #957, opened 2021-01-27
Use insecure port in configs only for scaleout architecture

vinaykul:
**What type of PR is this?**
>
> /kind api-change
/kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

**What this PR does / why we need it**: Currently in master branch, ENABLE_APISERVER_SECURE_PORT is broken, kube-up does not work. This fixes the issue for scaleup architecture by passing the ENABLE_APISERVER_SECURE_PORT to master vm configuration script, and uses ports 8080/8888 appropriately for scaleout architecture while retaining normal operation for scaleup.

Tested this by deploying 1TP-1RP cluster in scale-out-poc branch, and deploying ENABLE_INSECURE_PORT set to true as well as false and successfully bringing up scaleup cluster in both cases in master branch.

Fixes #

```release-note

```

### DIFF
--- a/cluster/gce/gci/configure-helper-common.sh
+++ b/cluster/gce/gci/configure-helper-common.sh
@@ -1423,11 +1423,7 @@ scrape_configs:
   - job_name: prometheus-metrics
     static_configs:
     - targets: ['127.0.0.1:2379','127.0.0.1:8080','127.0.0.1:10251','127.0.0.1:10252']
-  - job_name: 'haproxy'
-    static_configs:
-    - targets: ['scale_out_proxy_ip:8404']
 EOF
-  sed -i -e "s/scale_out_proxy_ip/$PROXY_RESERVED_IP/g" /tmp/prometheus-metrics.yaml
   ./prometheus --config.file="/tmp/prometheus-metrics.yaml" --web.listen-address=":9090" --web.enable-admin-api &
 }
 
--- a/cluster/gce/gci/configure-helper-common.sh
+++ b/cluster/gce/gci/configure-helper-common.sh
@@ -617,12 +617,10 @@ function create-master-auth {
     for extra_component in "${extra_components[@]}"; do
       local token="$(secure_random 32)"
       append_or_replace_prefixed_line "${known_tokens_csv}" "${token}," "system:${extra_component},uid:system:${extra_component}"
-      if [[ "${ENABLE_APISERVER_INSECURE_PORT:-false}" == "true" ]]; then
-        if [[ "${KUBERNETES_TENANT_PARTITION:-false}" == "true" ]]; then
-          create-kubeconfig "${extra_component}" "${token}" "${PROXY_RESERVED_IP}" "8888" "http"
-        else
-          create-kubeconfig "${extra_component}" "${token}" "localhost" "8080" "http"
-        fi
+      if [[ "${KUBERNETES_TENANT_PARTITION:-false}" == "true" ]]; then
+        create-kubeconfig "${extra_component}" "${token}" "${PROXY_RESERVED_IP}" "8888" "http"
+      elif [[ "${KUBERNETES_RESOURCE_PARTITION:-false}" == "true" ]]; then
+        create-kubeconfig "${extra_component}" "${token}" "localhost" "8080" "http"
       else
         create-kubeconfig "${extra_component}" "${token}"
       fi
@@ -1074,7 +1072,7 @@ function create-master-kubelet-auth {
   # set in the environment.
   if [[ -n "${KUBELET_APISERVER:-}" && -n "${KUBELET_CERT:-}" && -n "${KUBELET_KEY:-}" ]]; then
     REGISTER_MASTER_KUBELET="true"
-    if [[ "${ENABLE_APISERVER_INSECURE_PORT:-false}" == "true" ]]; then
+    if [[ "${KUBERNETES_RESOURCE_PARTITION:-false}" == "true" ]] || [[ "${KUBERNETES_TENANT_PARTITION:-false}" == "true" ]]; then
       create-kubelet-kubeconfig ${KUBELET_APISERVER} "8080" "http"
     else
       create-kubelet-kubeconfig ${KUBELET_APISERVER}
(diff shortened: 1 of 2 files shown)


cluster/gce/gci/configure-helper.sh
```diff
@@ -95,10 +95,10 @@ function main() {
     gke-master-start
   else
     create-node-pki
-    if [[ "${ENABLE_APISERVER_INSECURE_PORT:-false}" != "true" ]]; then
-      create-kubelet-kubeconfig ${KUBERNETES_MASTER_NAME}
-    else
+    if [[ "${KUBERNETES_RESOURCE_PARTITION:-false}" == "true" ]] || [[ "${KUBERNETES_TENANT_PARTITION:-false}" == "true" ]]; then
```
yb01, line 98: the logic flow is the same if not considering typos for the true/false. if considering typos, the previous one seems better.

cluster/gce/gci/configure-helper-common.sh
```diff
@@ -3197,12 +3176,10 @@ function ensure-master-bootstrap-kubectl-auth {
   # If the insecure port is disabled, kubectl will need to use an admin-authenticated kubeconfig.
   local master_ip=${1:-localhost}
   if [[ -n "${KUBE_BOOTSTRAP_TOKEN:-}" ]]; then
-    if [[ "${ENABLE_APISERVER_INSECURE_PORT:-false}" == "true" ]]; then
-      if [[ "${KUBERNETES_TENANT_PARTITION:-false}" == "true" ]]; then
-        create-kubeconfig "kube-bootstrap" "${KUBE_BOOTSTRAP_TOKEN}" "${PROXY_RESERVED_IP}" "8888" "http"
-      else
-        create-kubeconfig "kube-bootstrap" "${KUBE_BOOTSTRAP_TOKEN}" "${master_ip}" "8080" "http"
-      fi
+    if [[ "${KUBERNETES_TENANT_PARTITION:-false}" == "true" ]]; then
```
yb01, line 3179: this assumes if the RP or TP, it is always using the insecure port --- which is incorrect assumption. some hard coded TRUE for RP or TP in the config default have been removed. so we need to keep the ENABLE secure port check here.

cluster/gce/gci/configure-helper-common.sh
```diff
@@ -1691,11 +1689,12 @@ function start-kube-apiserver {
     params+=" --etcd-keyfile=${ETCD_APISERVER_CLIENT_KEY_PATH}"
   fi
   params+=" --secure-port=443"
-  if [[ "${ENABLE_APISERVER_INSECURE_PORT:-false}" != "true" ]]; then
+  if [[ "${ENABLE_APISERVER_INSECURE_PORT:-false}" == "false" ]]; then
     # Default is :8080
     params+=" --insecure-port=0"
-  else
-    params+=" --insecure-bind-address=0.0.0.0 --insecure-port=8080"
+  fi
```
yb01, line 1695: comparing with the default of this to-be-deprecated arg,  it is preferable to explicitly set.

cluster/gce/gci/configure-helper-common.sh
```diff
@@ -617,12 +617,10 @@ function create-master-auth {
     for extra_component in "${extra_components[@]}"; do
       local token="$(secure_random 32)"
       append_or_replace_prefixed_line "${known_tokens_csv}" "${token}," "system:${extra_component},uid:system:${extra_component}"
-      if [[ "${ENABLE_APISERVER_INSECURE_PORT:-false}" == "true" ]]; then
-        if [[ "${KUBERNETES_TENANT_PARTITION:-false}" == "true" ]]; then
-          create-kubeconfig "${extra_component}" "${token}" "${PROXY_RESERVED_IP}" "8888" "http"
-        else
-          create-kubeconfig "${extra_component}" "${token}" "localhost" "8080" "http"
-        fi
+      if [[ "${KUBERNETES_TENANT_PARTITION:-false}" == "true" ]]; then
```
yb01, line 620: again, this falls back to assumption that RP or TP always use insecure ports connection

vinaykul: In this version, yes - this is what we currently have today. When we have secure port support, we will be modifying this path.

cluster/gce/gci/configure-helper-common.sh
vinaykul: It defaults to 8080. No need to set it explicitly. I tested it and it works.

cluster/gce/gci/configure-helper-common.sh
vinaykul: At this time the assumption is correct. When we add support for secure in scaleout, we will change this to use https by default.

cluster/gce/gci/configure-helper.sh
vinaykul: what typos?

yb01: ENABLE_APISERVER_INSECURE_PORT=True, or ENABLE_APISERVER_INSECURE_PORT=treu,
previously it will leave to secure mode for kubeconfigs. with this change, it is left not controlled

cluster/gce/gci/configure-helper-common.sh
yb01: if this is the case, we probably need to reset this flag to TRUE in the default config if the RP or TP flags is set --- which is in the previous PR. so we can keep the previous flow here.

yb01 (approved):
/lgtm

approve since this is to fixing a blocking issue. we can address the implementation details later on in future PRs.

cluster/gce/gci/configure-helper.sh
vinaykul: 'true' seems to be the correct usage. i=I tested it for scaleout as well. we should only depend on _PARTITION vars. when we enable secure, then insecure port should be used only for debug as it is done today.

yb01 (approved):
/lgtm

sonyafenge (approved):
/LGTM

PR #966, opened 2021-01-29
fix volume usage for kube clients

pdgetrf:
### Changes

Fix usage of kubeClient in volume. 

### Validation

1. local box sanity functions passed
2. 100-node scale-up density test passed
3. 100-node scale-out 3-tp density test passed with both arktos and zeta tenant


pkg/volume/util/operationexecutor/operation_generator.go
```diff
@@ -73,14 +79,24 @@ type operationGenerator struct {
...
 // NewOperationGenerator is returns instance of operationGenerator
-func NewOperationGenerator(kubeClient clientset.Interface,
+func NewOperationGenerator(
+	kubeResourcePartitionClient clientset.Interface,
 	volumePluginMgr *volume.VolumePluginMgr,
 	recorder record.EventRecorder,
 	checkNodeCapabilitiesBeforeMount bool,
-	blkUtil volumepathhandler.BlockVolumePathHandler) OperationGenerator {
+	blkUtil volumepathhandler.BlockVolumePathHandler,
+	kubeTenantPartitionClients ...[]clientset.Interface) OperationGenerator {
```
pdgetrf, line 88: this code is also used in controller. hence the "optional" variable kubeTenantPartitionClients which is only provided by the kubelet, not the controller

pkg/kubelet/volumemanager/volume_manager.go
```diff
@@ -158,7 +158,7 @@ func NewVolumeManager(
 	checkNodeCapabilitiesBeforeMount bool,
 	keepTerminatedPodVolumes bool) VolumeManager {
 
-	vm := &volumeManager{
+	manager := &volumeManager{
```
pdgetrf, line 161: rename because "vm" is such a confusing name

cmd/kubelet/app/server.go
```diff
@@ -649,13 +649,13 @@ func run(s *options.KubeletServer, kubeDeps *kubelet.Dependencies, stopCh <-chan
 	// If the kubelet config controller is available, and dynamic config is enabled, start the config and status sync loops
 	if utilfeature.DefaultFeatureGate.Enabled(features.DynamicKubeletConfig) && len(s.DynamicConfigDir.Value()) > 0 &&
 		kubeDeps.KubeletConfigController != nil && !standaloneMode && !s.RunOnce {
-		if err := kubeDeps.KubeletConfigController.StartSync(kubeDeps.KubeClient, kubeDeps.EventClient, string(nodeName)); err != nil {
+		if err := kubeDeps.KubeletConfigController.StartSync(kubeDeps.KubeTPClients[0], kubeDeps.EventClient, string(nodeName)); err != nil {
```
chenqianfzh, line 652: kubelet config controller sounds like a component belonging to RP, instead of TP1. We may need to have a better understanding of this component first.

cmd/kubelet/app/server.go
```diff
@@ -649,13 +649,13 @@ func run(s *options.KubeletServer, kubeDeps *kubelet.Dependencies, stopCh <-chan
...
 			return err
 		}
 	}
 
 	if kubeDeps.Auth == nil {
-		auth, err := BuildAuth(nodeName, kubeDeps.KubeClient, s.KubeletConfiguration)
+		auth, err := BuildAuth(nodeName, kubeDeps.KubeTPClients[0], s.KubeletConfiguration)
```
chenqianfzh, line 658: Seems to me that the kubelet should authenticate itself to RP, instead of TP1. TPs are like consumers of the node resources,  while RP is the mananger of the node resources.

pkg/kubelet/pod_actions.go
```diff
@@ -105,7 +106,8 @@ func (kl *Kubelet) DoPodAction(action *v1.Action, pod *v1.Pod) {
 	}
 	actionStatus.Error = errStr
 	action.Status = actionStatus
-	if _, err := kl.kubeClient.CoreV1().Actions(action.Namespace).UpdateStatus(action); err != nil {
+	tenantPartitionClient := kubeclientmanager.ClientManager.GetTPClient(kl.kubeTPClients, action.Tenant)
+	if _, err := tenantPartitionClient.CoreV1().Actions(action.Namespace).UpdateStatus(action); err != nil {
```
chenqianfzh, line 110: Actions() -> ActionsWithMultiTenancy(ns, tenant)

pkg/kubelet/volume_host.go
```diff
@@ -69,8 +69,8 @@ func NewInitializedVolumePluginMgr(
 	const resyncPeriod = 0
 	if utilfeature.DefaultFeatureGate.Enabled(features.CSIDriverRegistry) {
 		// Don't initialize if kubeClient is nil
```
chenqianfzh, line 71: Nit: the comment is not updated to heartbeatclient

pkg/volume/util/operationexecutor/operation_generator.go
```diff
@@ -73,14 +78,25 @@ type operationGenerator struct {
...
+
+	// kubeTenantPartitionClients is for the case when NewOperationGenerator is called from kubelet
+	var tenantClients []clientset.Interface
+	if len(kubeTenantPartitionClients) > 0 {
+		tenantClients = kubeTenantPartitionClients[0] // call is from kubelet
```
chenqianfzh, line 92: Guess we need to calculate the corresponding TP client from the tenant, instead of using TP1?

cmd/kubelet/app/server.go
pdgetrf: agree. this will be fixed together when we switch to secure mode.

pkg/kubelet/pod_actions.go
pdgetrf: fixed

pkg/volume/util/operationexecutor/operation_generator.go
pdgetrf: this we need further discussion on the design. will be address later.

pkg/kubelet/volume_host.go
pdgetrf: this comment is removed as what's being discussed in it is self-explanatory from code

cmd/kubelet/app/server.go
pdgetrf: this will need further discussion on design. Will be fixed later.

chenqianfzh: OK. Please consider it resolved.

cmd/kubelet/app/server.go
chenqianfzh: OK. Please consider it resolved.

pkg/kubelet/pod_actions.go
chenqianfzh: OK. Please consider it resolved.

pkg/volume/util/operationexecutor/operation_generator.go
chenqianfzh: OK. Please consider it resolved.

pkg/kubelet/volumemanager/volume_manager.go
pdgetrf: this is a comment. resolving.

pkg/volume/util/operationexecutor/operation_generator.go
pdgetrf: also just a comment here. resolving.

Issue #783, opened 2020-10-16
[Perf-tests: arktos]object daemonset-priorityclass-0 creation error


sonyafenge:
**What happened**:
when run perf-tests - load testing, test failed with error:
```
E1015 14:02:06.830967   14966 clusterloader.go:217]   Errors: [namespace 0270w4-testns object daemonset-priorityclass-0 creation error: the server could not find the requested resource
```
**What you expected to happen**:
daemonset-priorityclass created successfully
**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Arktos version (use `kubectl version`):
- Cloud provider or hardware configuration:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Network plugin and version (if this is a network-related bug):
- Others:

sonyafenge:
This is 100% repro in perf-test load testing, even 100 hollow-nodes can repro. you can find detailed logs in any load testing result. here is one example: under GCP project: workload-controller-manager on sonyadev4: 
/home/sonyali/logs/perf-test/gce-100/arktos/1204-daily-100-1a0w1e

PR #976, opened 2021-02-08
Fix priorityclass missing in loadtest

chenqianfzh:
This PR fix https://github.com/CentaurusInfra/arktos/issues/783

The root cause of this bug is that, priorityclass, a cluster-scope obejct, was created with namespace/tenant set in the dynamic client. The code change fixed it by setting the namespace/tenant correctly according to the scope of the resource.

### Verification:
1. Scale-up 100-node load test finished without daemonset-priorityclass-0 creation error.
2. Scale-out 100-node load test finished without daemonset-priorityclass-0 creation error.
3.  Scale-up 100-node density test succeeded
4.  Scale-out 100-node density test succeeded


sonyafenge (approved):
Verified this PR on top of #977 and both density/load with 100 nodes Success.
/LGTM

Issue #491, opened 2020-07-16
perf-tests: perf-tests (500 nodes) failed with objects timed out error


sonyafenge:
**What happened**:
run perf-tests (500 hollow-node) and check test result get bunch of timeout error including DaemonSets, Statefulsets, deployments
**What you expected to happen**:
500 nodes perf-tests run successfully without timeout errors
**How to reproduce it (as minimally and precisely as possible)**:
1. following instruction to run perf-test: https://github.com/sonyafenge/arktos-tool/blob/master/perftools/Howtorunperf-tests.md
2. check perf-tests output. and get error
```
measurement call WaitForControlledPodsRunning - WaitForRunningDaemonSets error: 5 objects timed out: DaemonSets: mfqsdt-testns/daemonset-0, yqdubf-testns/daemonset-0, 3or4ob-testns/daemonset-0, iyobnn-testns/daemonset-0, nhpf7c-testns/daemonset-0
measurement call WaitForControlledPodsRunning - WaitForRunningStatefulSets error: 10 objects timed out: StatefulSets: 3or4ob-testns/medium-statefulset-0, nhpf7c-testns/medium-statefulset-0, nhpf7c-testns/small-statefulset-0, mfqsdt-testns/medium-statefulset-0, mfqsdt-testns/small-statefulset-0, iyobnn-testns/medium-statefulset-0, yqdubf-testns/small-statefulset-0, 3or4ob-testns/small-statefulset-0, yqdubf-testns/medium-statefulset-0, iyobnn-testns/small-statefulset-0
[...]

PR #977, opened 2021-02-08
add a switch to disable PV in load test

chenqianfzh:
Scale-out and Scale-up load tests both failed due ot PV/PVC processing of deployments&statefulsets. The deployments/statefulset which uses volumes will never get the pods in "Running" state.  As we are continuing to solve issue (tracked in https://github.com/CentaurusInfra/arktos/issues/978), this PR provides a switch to skip this issue temporarily.

When the env var  CL2_ENABLE_PVS is set to false, the test deployments&statefulsets will be created without attaching to PVs.

### Verification:
This PR was tested on top of https://github.com/CentaurusInfra/arktos/pull/976, which fixed another error in load-test, and I got the following result:
1. Scale-up 100-node load test succeeded with commandline test-configs "--testconfig=testing/load/config.yaml --testoverrides=./testing/experiments/disable_pvs.yaml".
1. Scale-out 100-node load test succeeded with commandline test-configs "--testconfig=testing/load/config.yaml --testoverrides=./testing/experiments/disable_pvs.yaml".
3. Scale-up 100-node density test succeeded
4. Scale-out 100-node density test succeeded

### DIFF
--- a/perf-tests/clusterloader2/testing/experiments/disable_pvs.yaml
+++ b/perf-tests/clusterloader2/testing/experiments/disable_pvs.yaml
@@ -1,0 +1,1 @@
+CL2_ENABLE_PVS: false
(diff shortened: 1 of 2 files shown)


sonyafenge (approved):
Verified this PR on top of https://github.com/CentaurusInfra/arktos/pull/976 and both density and load with 100 nodes Success. 
This PR can be a temporally workaround for https://github.com/CentaurusInfra/arktos/issues/491

/LGTM

PR #982, opened 2021-02-09
Fix pod/service deleting issue

Hong-Chang:
Currently in mizar pod and service controller, when deleting the object, the code will try to retrieve the object. Since the object has already been deleted, the deleting operation will always trigger an error.

The issue happens only when mizar is running because it happens in mizar controllers.

The fix: when deleting object, don't retrieve it at that time. Instead, just send the name, namespace, tenant info to mizar backend for deleting.

**What type of PR is this?**
/kind bug

Current pod and service deleting will always throw error for mizar controllers.

### DIFF
--- a/pkg/controller/mizar/mizar-pod-controller.go
+++ b/pkg/controller/mizar/mizar-pod-controller.go
@@ -18,6 +18,7 @@ import (
 	"time"
 
 	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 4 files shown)


h-w-chen (commented):
better to have an arktos engineer to review the code, as each project/team has different code bar.

chenqianfzh (approved):

pkg/controller/mizar/mizar-service-controller.go
```diff
@@ -184,7 +185,13 @@ func (c *MizarServiceController) syncService(eventKeyWithType KeyWithEventType)
 
 	svc, err := c.serviceLister.ServicesWithMultiTenancy(namespace, tenant).Get(name)
 	if err != nil {
-		return err
+		if event == EventType_Delete {
+			if !apierrors.IsNotFound(err) {
+				klog.Errorf("Should get NotFound error when retrieving deleted object %s/%s/%s but got error: %v", tenant, namespace, name, err)
+			}
```
chenqianfzh, line 191: Shall we "return err" here?

pkg/controller/mizar/mizar-pod-controller.go
```diff
@@ -180,10 +181,23 @@ func (c *MizarPodController) handle(keyWithEventType KeyWithEventType) error {
...
+		if eventType == EventType_Delete {
+			obj = &v1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      name,
+					Namespace: namespace,
+					Tenant:    tenant,
+				},
+			}
+			if !errors.IsNotFound(err) {
+				klog.Errorf("Should get NotFound error when retrieving deleted object %s/%s/%s but got error: %v", tenant, namespace, name, err)
```
chenqianfzh, line 193: Shall we "return err" here?

PR #987, opened 2021-02-12
Use HTTPS as etcd-apiserver protocol when mTLS is enabled

sonyafenge:
refer to kubernetes PR https://github.com/kubernetes/kubernetes/pull/77561 and https://github.com/kubernetes/kubernetes/pull/77657

**What type of PR is this?**

> /kind bug

when run arktos perf-test with these two env var: PROMETHEUS_SCRAPE_ETCD=true ENABLE_PROMETHEUS_SERVER=true to enable apiresponsiveness. But Prometheus failed to start with the error below when run clusterloader2
```
0/1 targets are ready, example not ready target: {map[endpoint:etcd-2382 instance:10.40.0.9:2382 job:master namespace:monitoring service:master] down}
```
arktos are missing the insecure http port number:2382 to access ETCD. This port 2382 is used by perf-test to access etcd matrix.

Fixes #

Tested on GCE 100 nodes for both density and load, both got finished with success. 

```release-note

```

### DIFF
--- a/cluster/gce/gci/configure-helper-common.sh
+++ b/cluster/gce/gci/configure-helper-common.sh
@@ -1546,9 +1546,12 @@ function prepare-etcd-manifest {
   local etcd_cluster=""
   local cluster_state="new"
   local etcd_protocol="http"
+  local etcd_apiserver_protocol="http"
   local etcd_creds=""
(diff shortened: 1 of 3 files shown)


Sindica (approved):
/lgtm

Issue #339, opened 2020-06-02
Consisten etcd key paths for the system tenants and regular tenants.


chenqianfzh:
Consisten etcd key paths for the system tenants and regular tenants. It gives etcd partition more systematic and convenient.

Issue #435, opened 2020-07-01
"kubectl cp" fails


chenqianfzh:
"kubectl scale" fails

The following is test based on

test server https://18.236.170.164:6443
context tenant-qian-5-admin-context

test:
```
qianchen@qianchen-VirtualBox:~$ kubectl get pod nginx --namespace default
NAME    HASHKEY               READY   STATUS    RESTARTS   AGE
nginx   2334343887965160474   1/1     Running   0          9h

qianchen@qianchen-VirtualBox:~$ kubectl cp /home/qianchen/go113/go/src/archive/tar/testdata/hardlink.tar default/nginx:/tmp/
error: unable to upgrade connection: Forbidden (user=kube-apiserver, verb=create, resource=nodes, subresource=proxy)
```

chenqianfzh:
Fixed in https://github.com/CentaurusInfra/arktos/pull/989

PR #989, opened 2021-02-16
Consistent etcd key path

chenqianfzh:
This PR lets the resources under the system-tenant space and non-system-tenant space have the same format of etcd keys. 

Previously, for the convenience of development, the resource of system tenant has a backward-compatible etcd key format: 
```
/registry/[resource_type]/[ns_name]/[resource_name]
```
while those of non-system tenant have a different format:
```
/registry/[resource_type]/[tenant-Name]/[ns_name]/[resource_name]
```

However, as the system-tenant etcd key prefix is the same as that of all-tenant etcd key prefix, it caused the problem that list the resources under system tenants actually will return the resources of every tenant.  This leads to problems like https://github.com/CentaurusInfra/arktos/issues/339, which hinders the development of other multi-tenancy features.

With this change, the etcd key of the resource of system tenant  is now in the following format: 
```
/registry/[resource_type]/system/[ns_name]/[resource_name]
```

#### 1. Check the etcd key path generated:
Check the etcd keys with the following command,
```
ETCDCTL_API=3 etcdctl get "" --prefix=true --keys-only
```

 and the tenant of "system" is shown in the etcd key paths as expected
image

#### 2. Verify that issue 339 is resolved
a. create a tenant 
```
kubectl create tenant x
```

b. Run Command to show the namespaces of system-tenant
```
 kubectl get namepsace --tenant system
```

As show below, only the namespaces under system-tenant will be shown:
image

Yet, before this change, the output will be as follows, where the namespaces under all the tenants will be returned:

image

c. The code change was verified by the multi-tenancy e2e test.


staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go
```diff
@@ -331,9 +331,7 @@ type Initializer struct {
 
 const (
 	// TenantAll is the default argument to specify on a context when you want to list or filter resources across all tenants
-	TenantAll string = ""
-	// Sometimes we need to explicitly indicate the operation targets at all the tenants, not the empty string of the above TenantAll
-	TenantAllExplicit string = "all"
+	TenantAll string = "all"
```
pdgetrf, line 334: with this, "all" would be a reserved phrase and it should be checked against when regular user creates tenant

staging/src/k8s.io/apiserver/pkg/registry/generic/registry/store.go
```diff
@@ -253,15 +251,12 @@ func TenantKeyRootFunc(ctx context.Context, prefix string) string {
 // namespace-scope resource relative to the given prefix enforcing tenant/namespace rules. If the
 // context does not contain a tenant/namespace, it errors.
 func NamespaceKeyFunc(ctx context.Context, prefix string, name string) (string, error) {
-	var te, ns string
+	var tenant, ns string
```
pdgetrf, line 254: god yes.

staging/src/k8s.io/client-go/dynamic/fake/simple.go
```diff
@@ -80,11 +80,15 @@ type dynamicResourceClient struct {
...
-	return &dynamicResourceClient{client: c, resource: resource}
+	return &dynamicResourceClient{client: c, resource: resource, namespace: metav1.NamespaceAll, tenant: metav1.TenantAll}
 }
 
 func (c *dynamicResourceClient) Namespace(ns string) dynamic.ResourceInterface {
-	return c.NamespaceWithMultiTenancy(ns, metav1.TenantSystem)
+	if ns != metav1.NamespaceAll {
+		return c.NamespaceWithMultiTenancy(ns, metav1.TenantSystem)
+	} else {
+		return c.NamespaceWithMultiTenancy(ns, metav1.TenantAll)
```
pdgetrf, line 90: nit: else { not really needed here.

pdgetrf (approved):
/lgtm

a few minor comments.

staging/src/k8s.io/apimachinery/pkg/apis/meta/v1/types.go
chenqianfzh: Correct. It is already checked here:

image

chenqianfzh:
This PR resolves issues https://github.com/CentaurusInfra/arktos/issues/339 and https://github.com/CentaurusInfra/arktos/issues/435.

PR #1000, opened 2021-02-23
Short path fix

chenqianfzh:


Sindica:
Why do we need this? What was intended for System Tenant and why revert back to TenantNone?

chenqianfzh:
> Why do we need this? What was intended for System Tenant and why revert back to TenantNone?
> 
We cannot assume the default tenant is "System tenant", otherwise a regular tenant will get system-level access by default. The tenant info needs to be extract from the user information, a design referred as "short path".

pdgetrf:
wow, hit #1000 mark!

PR #1003, opened 2021-02-24
Scheduler talks to TPs directly with SSL enabled in dev environment

Sindica:
. Pick up yunwen's scheduler changes.
. Generate RP kubeconfig for scheduler in TP - with SSL enabled
. Works in 2TP/1RP

### DIFF
--- a/docs/setup-guide/scale-out-local-dev-setup.md
+++ b/docs/setup-guide/scale-out-local-dev-setup.md
@@ -34,13 +34,28 @@ export RESOURCE_PARTITION_IP=[RP_IP]
 1. Set up environment variables
 
 ```
-export SCALE_OUT_PROXY_ENDPOINT=http://[PROXY_IP]:8888
+export SCALE_OUT_PROXY_IP=[PROXY_IP]
+export SCALE_OUT_PROXY_PORT=8888
 export IS_RESOURCE_PARTITION=false
+export RESOURCE_SERVER=[RP_IP]
+export REUSE_CERTS=true
 ```
 
 1. Run ./hack/arktos-up-scale-out-poc.sh
 
 1. Expected last line of output: "Tenant Partition Cluster is Running ..."
+
+Note:
+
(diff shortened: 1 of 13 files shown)


yb01:
please rebase with master to get the scheduler code changes

docs/setup-guide/scale-out-local-dev-setup.md
```diff
@@ -34,23 +34,40 @@ export RESOURCE_PARTITION_IP=[RP_IP]
...
 
 1. Expected last line of output: "Tenant Partition Cluster is Running ..."
 
+Note:
+
+1. As we start to picking up secure mode in scale out, api server certificates will be shared across all partitions in 
+development environment. The first TP that started needs to generate api server certificates and be copied over to other
+TP/RP before they start.
+
+1. To generate first set of certs, run `REUSE_CERTS=false; ./hack/arktos-up-scale-out-poc.sh`
```
yb01, line 54: the IPs ( or master name ) for the other cluster's master will needed to set the proper SANs or the certs.

hack/lib/util.sh
```diff
@@ -507,22 +507,15 @@ function kube::util::write_client_kubeconfig {
...
+    local protocol=${8:-"https"}
 
     cat <<EOF | ${sudo} tee "${dest_dir}"/"${client_id}".kubeconfig > /dev/null
 apiVersion: v1
 kind: Config
 clusters:
   - cluster:
       certificate-authority: ${ca_file}
-      server: ${MASTER_ENDPOINT}
+      server: ${protocol}://${api_host}:${api_port}/
```
yb01, line 518: do we need the tailing "/"

Sindica: This is the original code

Sindica: I added protocol

docs/setup-guide/scale-out-local-dev-setup.md
Sindica: This is user manual. It's better let the tool handle the terminology. I would avoid add technical terms here.

PR #1009, opened 2021-03-03
Take multiple RPs in Scheduler

Sindica:
Tested with 1TP/1RP, 1TP/2RP, 2TP/2RP.

. In both TP,  %cat /tmp/kube-scheduler.log | grep "Add node"
I0302 23:35:20.921581    5965 eventhandlers.go:98] Add node to cache. node [ip-172-30-0-122], rpId [rp0]
I0302 23:36:12.868404    5965 eventhandlers.go:98] Add node to cache. node [ip-172-30-0-172], rpId [rp1]

. kubectl create pods for tenant a, z both works
. As discussed, use insecure mode for local dev test environment now
. Might needs further changes in scheduler to make sure rpId to be properly used in node evaluation

### DIFF
--- a/cmd/kube-apiserver/app/aggregator.go
+++ b/cmd/kube-apiserver/app/aggregator.go
@@ -211,8 +211,8 @@ func makeAPIServiceAvailableHealthCheck(name string, apiServices []*apiregistrat
 
 	// Watch add/update events for APIServices
 	apiServiceInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc:    func(obj interface{}) { handleAPIServiceChange(obj.(*apiregistration.APIService)) },
-		UpdateFunc: func(old, new interface{}) { handleAPIServiceChange(new.(*apiregistration.APIService)) },
+		AddFunc:    func(obj interface{}, rpId string) { handleAPIServiceChange(obj.(*apiregistration.APIService)) },
+		UpdateFunc: func(old, new interface{}, rpId string) { handleAPIServiceChange(new.(*apiregistration.APIService)) },
 	})
 
 	// Don't return healthy until the pending list is empty
--- a/cmd/kube-scheduler/app/options/deprecated.go
+++ b/cmd/kube-scheduler/app/options/deprecated.go
@@ -20,7 +20,6 @@ package options
 import (
 	"fmt"
 	"github.com/spf13/pflag"
-
 	"k8s.io/apimachinery/pkg/util/validation/field"
 	kubeschedulerconfig "k8s.io/kubernetes/pkg/scheduler/apis/config"
 	"k8s.io/kubernetes/pkg/scheduler/factory"
@@ -66,7 +65,7 @@ func (o *DeprecatedOptions) AddFlags(fs *pflag.FlagSet, cfg *kubeschedulerconfig
 		"RequiredDuringScheduling affinity is not symmetric, but there is an implicit PreferredDuringScheduling affinity rule corresponding "+
 			"to every RequiredDuringScheduling affinity rule. --hard-pod-affinity-symmetric-weight represents the weight of implicit PreferredDuringScheduling affinity rule. Must be in the range 0-100.")
 	fs.MarkDeprecated("hard-pod-affinity-symmetric-weight", "This option was moved to the policy configuration file")
-	fs.StringVar(&cfg.ResourceProviderClientConnection.Kubeconfig, "resource-providers", cfg.ResourceProviderClientConnection.Kubeconfig, "string representing resource provider kubeconfig files")
+	fs.StringVar(&cfg.ResourceProviderKubeConfig, "resource-providers", cfg.ResourceProviderKubeConfig, "string representing resource provider kubeconfig files")
 }
 
 // Validate validates the deprecated scheduler options.
--- a/cmd/kube-scheduler/app/options/options.go
+++ b/cmd/kube-scheduler/app/options/options.go
@@ -19,9 +19,11 @@ package options
 
 import (
 	"fmt"
+	coreinformers "k8s.io/client-go/informers/core/v1"
 	"net"
 	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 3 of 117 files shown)


cmd/kube-apiserver/app/aggregator.go
```diff
@@ -211,8 +211,8 @@ func makeAPIServiceAvailableHealthCheck(name string, apiServices []*apiregistrat
 
 	// Watch add/update events for APIServices
 	apiServiceInformer.Informer().AddEventHandler(cache.ResourceEventHandlerFuncs{
-		AddFunc:    func(obj interface{}) { handleAPIServiceChange(obj.(*apiregistration.APIService)) },
-		UpdateFunc: func(old, new interface{}) { handleAPIServiceChange(new.(*apiregistration.APIService)) },
+		AddFunc:    func(obj interface{}, rpId string) { handleAPIServiceChange(obj.(*apiregistration.APIService)) },
```
yb01, line 214: what is the name format of the rpId, they need to be unique so a guid string probably is needed.

cmd/kube-scheduler/app/options/deprecated.go
```diff
@@ -66,7 +65,7 @@ func (o *DeprecatedOptions) AddFlags(fs *pflag.FlagSet, cfg *kubeschedulerconfig
 		"RequiredDuringScheduling affinity is not symmetric, but there is an implicit PreferredDuringScheduling affinity rule corresponding "+
 			"to every RequiredDuringScheduling affinity rule. --hard-pod-affinity-symmetric-weight represents the weight of implicit PreferredDuringScheduling affinity rule. Must be in the range 0-100.")
 	fs.MarkDeprecated("hard-pod-affinity-symmetric-weight", "This option was moved to the policy configuration file")
-	fs.StringVar(&cfg.ResourceProviderClientConnection.Kubeconfig, "resource-providers", cfg.ResourceProviderClientConnection.Kubeconfig, "string representing resource provider kubeconfig files")
+	fs.StringVar(&cfg.ResourceProviderKubeConfig, "resource-providers", cfg.ResourceProviderKubeConfig, "string representing resource provider kubeconfig files")
```
yb01, line 68: so it seem there is no simple way to expand the client connection to be an array of componentbaseconfig.ClientConnectionConfiguration, right ?

cmd/kube-scheduler/app/options/options.go
```diff
@@ -280,9 +291,15 @@ func (o *Options) Config() (*schedulerappconfig.Config, error) {
 	return c, nil
 }
 
-func kubeconfigFileExists(name string) bool {
-	_, err := os.Stat(name)
-	return err == nil
+func kubeconfigFileExists(kubeConfigFilenames string) ([]string, bool) {
```
yb01, line 294: Nit. probably rename it as parseKubeconfigList ?

hack/lib/common.sh
```diff
@@ -222,7 +222,7 @@ function kube::common::generate_certs {
       kube::util::create_signing_certkey "${CONTROLPLANE_SUDO}" "${CERT_DIR}" request-header '"client auth"'
 
       # serving cert for kube-apiserver
-      kube::util::create_serving_certkey "${CONTROLPLANE_SUDO}" "${CERT_DIR}" "server-ca" kube-apiserver kubernetes.default kubernetes.default.svc "localhost" "${API_HOST_IP}" "${API_HOST}" "${FIRST_SERVICE_CLUSTER_IP}" "${API_HOST_IP_EXTERNAL}" "${RESOURCE_SERVER:-}" "${APISERVERS_EXTRA:-}" "${PUBLIC_IP:-}"
+      kube::util::create_serving_certkey "${CONTROLPLANE_SUDO}" "${CERT_DIR}" "server-ca" kube-apiserver kubernetes.default kubernetes.default.svc "localhost" "${API_HOST_IP}" "${API_HOST}" "${FIRST_SERVICE_CLUSTER_IP}" "${API_HOST_IP_EXTERNAL}" "${APISERVERS_EXTRA:-}" "${PUBLIC_IP:-}"
```
yb01, line 225: any reason the RESOURCE_SERVER is removed ?

hack/lib/common.sh
```diff
@@ -385,29 +385,39 @@ EOF
...
         ADMIN_CONFIG_API_HOST=${PUBLIC_IP:-${API_HOST}}
-        kube::util::write_client_kubeconfig "${CONTROLPLANE_SUDO}" "${CERT_DIR}" "${ROOT_CA_FILE}" "${ADMIN_CONFIG_API_HOST}" "$secureport" admin
+
         ${CONTROLPLANE_SUDO} chown "${USER}" "${CERT_DIR}/client-admin.key" # make readable for kubectl
 
         if [[ "${IS_SCALE_OUT}" == "true" ]]; then
           # in scale out poc, point api server to proxy
-          kube::util::write_client_kubeconfig "${CONTROLPLANE_SUDO}" "${CERT_DIR}" "${ROOT_CA_FILE}" "${SCALE_OUT_PROXY_IP}" "${SCALE_OUT_PROXY_PORT}" controller "" "http"
-          kube::util::write_client_kubeconfig "${CONTROLPLANE_SUDO}" "${CERT_DIR}" "${ROOT_CA_FILE}" "${SCALE_OUT_PROXY_IP}" "${SCALE_OUT_PROXY_PORT}" scheduler "" "http"
+          kube::util::write_client_kubeconfig "${CONTROLPLANE_SUDO}" "${CERT_DIR}" "${ROOT_CA_FILE}" "${ADMIN_CONFIG_API_HOST}" "${API_PORT}" admin "" "http"
```
yb01, line 397: if the superbox env supports the HAProxy, this will be needed.

hack/lib/common.sh
```diff
@@ -559,7 +569,14 @@ function kube::common::start_kubescheduler {
     SCHEDULER_LOG=${LOG_DIR}/kube-scheduler.log
 
     if [[ "${IS_SCALE_OUT}" == "true" ]]; then
-      RESOURCE_PROVIDER_KUBECONFIG_FLAGS="--resource-providers=${CERT_DIR}/resource-provider-scheduler.kubeconfig"
+      RESOURCE_PROVIDER_KUBECONFIG_FLAGS="--resource-providers="
```
yb01, line 572: Nice!

pkg/cloudfabric-controller/controllerframework/controllerinstancemanager.go
```diff
@@ -120,10 +120,10 @@ func GetInstanceId() types.UID {
 	}
```
yb01, line 120: Do we really need this changes here in this file ?

pkg/controller/volume/scheduling/scheduler_binder.go
```diff
@@ -469,9 +469,14 @@ func (b *volumeBinder) checkBindings(pod *v1.Pod, bindings []*bindingInfo, claim
 		return false, fmt.Errorf("failed to get cached claims to provision for pod %q", podName)
 	}
 
-	node, err := b.nodeInformer.Lister().Get(pod.Spec.NodeName)
-	if err != nil {
-		return false, fmt.Errorf("failed to get node %q: %v", pod.Spec.NodeName, err)
+	var node *v1.Node
```
yb01, line 472: I recall we will have some means to associate the node and the RP, so we won't need to end up looping through them to find the node by names.

pkg/controller/volume/scheduling/scheduler_binder.go
```diff
@@ -469,9 +469,14 @@ func (b *volumeBinder) checkBindings(pod *v1.Pod, bindings []*bindingInfo, claim
...
+	var err error
+	for _, nodeInformer := range b.nodeInformers {
+		node, err = nodeInformer.Lister().Get(pod.Spec.NodeName)
+		if err != nil { // TODO - check error type, continue to search next RP if no found
```
yb01, line 476: likely the TODO is needed in this PR instead of later. e.g. if the node is on the second RP, this function currently will return error.

pkg/kubelet/kubelet.go
```diff
@@ -454,7 +454,10 @@ func NewMainKubelet(kubeCfg *kubeletconfiginternal.KubeletConfiguration,
 		r := cache.NewReflector(nodeLW, &v1.Node{}, nodeIndexer, 0)
 		go r.Run(wait.NeverStop)
 	}
-	nodeInfo := &predicates.CachedNodeInfo{NodeLister: corelisters.NewNodeLister(nodeIndexer)}
+
+	cacheNodeInfos := make(map[string]corelisters.NodeLister, 1)
```
yb01, line 458: per the current design, each node ONLY belongs to one RP. so kubelet will not need this array here.

cmd/kube-apiserver/app/aggregator.go
Sindica: Possible. Right now, we will use 0,1,2 as the sequence they come from KubeConfigFile arg. They only need to be unique to each other inside a scheduler.

cmd/kube-scheduler/app/options/deprecated.go
Sindica: I tried to make ClientConnectionConfiguration an array. But it did not work for me when assign resource-providers to ClientConnectionConfiguration[0].Kubeconfig

cmd/kube-scheduler/app/options/options.go
Sindica: Yeah, it's a nit as it was old code :) Will do

hack/lib/common.sh
Sindica: admin.kubeconfig is used local. In arktos-up and in scale-out-local-cluster up, one of its usage is to check whether node is up and running from kubectl. It should go to local apiserver, not HA proxy.

hack/lib/common.sh
Sindica: RESOURCE_SERVER can be an array now. It does not work here anymore. Also we are using insecure mode so not needed for now.

pkg/cloudfabric-controller/controllerframework/controllerinstancemanager.go
Sindica: > Do we really need this changes here in this file ?

Necessary for compilation

pkg/controller/volume/scheduling/scheduler_binder.go
Sindica: The association is in NodeInfo and podState. Neither of them were passed into this function. Keep the looping now to pass the compilation w/o further changes.

> I recall we will have some means to associate the node and the RP, so we won't need to end up looping through them to find the node by names.

pkg/kubelet/kubelet.go
Sindica: > per the current design, each node ONLY belongs to one RP. so kubelet will not need this array here.

As we discussed over zoom, this change is due to implementation changes of NewCachedNodeInfo() in scheduler

pkg/controller/volume/scheduling/scheduler_binder.go
Sindica: This is in volumeBinder checkBindings. Currently I don't have a use case to test what will be the error message so don't have a way to handle. I will add an error log for now.

yb01:
Overall functionality wise it is OK for our private branch. I am bit concerned on all the interface changes to support the rpId as new parameters, the affect code is just too much ( and probably not all the test code has been updated yet). probably we need to do more investigation before merging to master.
we can get this PR in to the private branch for 430 code merging, integration of code changes.

PR #1001, opened 2021-02-23
Enable ssl transmission among Arktos components and enhance kubemark cluster setup flow

yb01:
**What type of PR is this?**
> /kind feature

**What this PR does / why we need it**:
enables ssl transmition among arktos components.
**Which issue(s) this PR fixes**:
Fixes #

**Special notes for your reviewer**:
```
1.       Env variable/flag to control if the env enables the insecure ports 
2.       Each cluster has its own kubeconfig files, and pushing to minions so hollow-nodes can use them.   
3.       Controllers take array of kubeconfigs instead of currently tenantServers, resourceServers args and construct the clientsets accordingly.  
4.       Kubelet take array of kubeconfigs instead of currently tenantServers, resourceServers args and construct the clientsets accordingly. 
5.       Scheduler to communicate to secure port on proxy server.
6.       Proxy to communicate with TPs, RPs via the secure ports ( hopefully this is just config change as token in headers should be supported already.  
7.       Proxy uses secure port to accept client connection from kubemark and scheduler.
```
**Testing**:
```
I0224 20:15:15.410786   29745 probes.go:130] Probe DnsLookupLatency wasn't started, skipping the Dispose() step
I0224 20:15:35.427025   29745 simple_test_executor.go:385] Resources cleanup time: 20.016249245s
I0224 20:15:35.427078   29745 clusterloader.go:212] --------------------------------------------------------------------------------
I0224 20:15:35.427084   29745 clusterloader.go:213] Test Finished
I0224 20:15:35.427089   29745 clusterloader.go:214]   Test: testing/density/config.yaml
I0224 20:15:35.427094   29745 clusterloader.go:215]   Status: Success
I0224 20:15:35.427098   29745 clusterloader.go:219] --------------------------------------------------------------------------------

JUnit report was created: /home/cloudshare/logs/test/junit.xml
```

Scale-up perf run:
```
I0225 21:49:53.542759   21175 simple_test_executor.go:385] Resources cleanup time: 15.016946284s
I0225 21:49:53.542831   21175 clusterloader.go:212] --------------------------------------------------------------------------------
I0225 21:49:53.542841   21175 clusterloader.go:213] Test Finished
I0225 21:49:53.542848   21175 clusterloader.go:214]   Test: testing/density/config.yaml
I0225 21:49:53.542855   21175 clusterloader.go:215]   Status: Success
I0225 21:49:53.542861   21175 clusterloader.go:219] --------------------------------------------------------------------------------

[...]

PR #1011, opened 2021-03-05
Use kubeconfig for TPs in nodelifecycle controller

Sindica:
This change picks up Yunwen's arktos code change in https://github.com/CentaurusInfra/arktos/pull/1001

Additionally, changed command line arg tenant-server-kubeconfigs from array to string tenant-server-kubeconfig to be consistent with scheduler changes. This will prevent further interface changes when we picking up auto discovery of TP/RPs.

Made script changes to local cluster up to accommodate the code change.


cmd/genutils/genutils.go
```diff
@@ -41,3 +43,17 @@ func OutDir(path string) (string, error) {
 	outDir = outDir + "/"
 	return outDir, nil
 }
+
+// ParseKubeConfigFiles gets a string that contains one or multiple kubeconfig files.
+// If there are more than one file, separated by comma
```
yb01, line 48: should the stringSliceVar from command line args flag take care of this parsing already ?

docs/setup-guide/scale-out-local-dev-setup.md
```diff
@@ -57,7 +57,7 @@ Note:
 export SCALE_OUT_PROXY_IP=[PROXY_IP]
 export SCALE_OUT_PROXY_PORT=8888
 export IS_RESOURCE_PARTITION=true
-export TENANT_SERVERS=http://[TP1_IP]:8080,http://[TP2_IP]:8080
+export TENANT_SERVER=[TP1_IP]<,[TP2_IP]>
```
yb01, line 60: should change to kubeconfigs

hack/arktos-up-scale-out-poc.sh
```diff
@@ -55,12 +55,7 @@ fi
 KUBE_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 
 # for POC, the kubelet_flags is used for the new temporary kubelet commandline args
-KUBELET_FLAGS="--tenant-servers="${TENANT_SERVERS}
-# --tenant-servers set the tenant server urls for node-lifecycle controller
-CONTROLLER_MANAGER_EXTRA_FLAGS="--tenant-servers="${TENANT_SERVERS}
-
-echo KUBELET_FLAGS for new kubelet commandline --tenant-servers
-echo ${KUBELET_FLAGS}
+KUBELET_FLAGS="--tenant-servers=http://172.30.0.148:8080,http://172.30.0.24:8080"
```
yb01, line 58: with hardcoded IPs for tenant servers ?

cmd/genutils/genutils.go
Sindica: As I explained earlier, I prefer let internal logic decide treat the argument as a string or array. This logic applies to both single or multiple kubeconfigs without option parsing changes.

docs/setup-guide/scale-out-local-dev-setup.md
Sindica: Kubeconfig is generated from TENANT_SERVER automatically. Otherwise, dev needs to provide their own kubeconfig in local cluster up. This is not ideal.

hack/arktos-up-scale-out-poc.sh
Sindica: I had to hardcode it in my local test as I wanted to let kubelet and KCM share same environment variable TENANT_SERVERS. But kubelet code was not updated to use kubeconfig so I did not update the corresponding script.

Will fix the hardcode servers in this CR

hack/arktos-up-scale-out-poc.sh
```diff
@@ -54,13 +54,17 @@ fi
...
 KUBE_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 
-# for POC, the kubelet_flags is used for the new temporary kubelet commandline args
-KUBELET_FLAGS="--tenant-servers="${TENANT_SERVERS}
-# --tenant-servers set the tenant server urls for node-lifecycle controller
-CONTROLLER_MANAGER_EXTRA_FLAGS="--tenant-servers="${TENANT_SERVERS}
 
-echo KUBELET_FLAGS for new kubelet commandline --tenant-servers
-echo ${KUBELET_FLAGS}
+TENANT_SERVERS_KUBELET=${TENANT_SERVERS_KUBELET:-}
```
yb01, line 58: TENANT_SERVERS_KUBELET seems a bit confusing , from the error code it seem it is the server, not the kubelet.

hack/lib/common.sh
```diff
@@ -510,51 +512,80 @@ function kube::common::start_controller_manager {
...
-        --pvclaimbinder-sync-period="${CLAIM_BINDER_SYNC_PERIOD}" \
-        --feature-gates="${FEATURE_GATES}" \
-        "${cloud_config_arg[@]}" \
-        --kubeconfig "${kubeconfigfilepaths}" \
-        --controllers="${KUBE_CONTROLLERS}" \
-        --leader-elect=false \
-        --cert-dir="${CERT_DIR}" \
-        --default-network-template-path="${ARKTOS_NETWORK_TEMPLATE}" >"${CTLRMGR_LOG}" 2>&1 &
+    KUBE_CONTROLLERS="*"
+    if [[ "${IS_SCALE_OUT}" == "true" ]]; then
```
yb01, line 516: since it only one line difference from scale-up and scale-out command lines, probably it will be clearer to use the if-else block to make the desired args for the --controllers arg to avoid such dup for most the command line

yb01:
what tests has been done with this PR ?

hack/arktos-up-scale-out-poc.sh
Sindica: This is the --tenant-servers arg used in kubelet. The code was not ready to pick up kubeconfig so I had to use a workaround. This is explained in the manual.

hack/lib/common.sh
Sindica: Will remove line 515

Sindica:
> what tests has been done with this PR ?

In TP/RP, start cluster with this command and created pods.
./hack/arktos-up-scale-out-poc.sh

hack/lib/common.sh
Sindica: Removed 515 and tested in TP/RP

PR #1015, opened 2021-03-09
Multiple RPs support in KCM

Sindica:
Please ignore the first 2 commits. They are reviewed in https://github.com/CentaurusInfra/arktos/pull/1011.
-- merged and removed.

Tests done:
In TP1:
ubuntu@ip-172-30-0-148:~/go/src/arktos$ cat /tmp/kube-controller-manager.log | grep "Caches are synced for " | grep node
I0309 22:12:36.434773     979 controller_utils.go:1089] Caches are synced for attach detach (w/o node) controller
I0309 22:12:36.504194     979 controller_utils.go:1089] Caches are synced for persistent volume (w/o node) controller
I0309 22:12:56.435835     979 nodecache_utils.go:90] Caches are synced for attach detach (node) controller
I0309 22:12:56.505039     979 nodecache_utils.go:90] Caches are synced for persistent volume (node) controller

. Note service controller is not started in local cluster up.

In RPs:
ubuntu@ip-172-30-0-122:/var/run/kubernetes$ cat /tmp/kube-controller-manager.log | grep "Caches are synced for"
I0309 22:12:54.374956   30486 controller_utils.go:1089] Caches are synced for daemon sets controller
I0309 22:12:54.378205   30486 controller_utils.go:1089] Caches are synced for TTL controller
I0309 22:12:54.381359   30486 apiserverconfigmanager.go:166] Caches are synced for api server end points. [map[1:{Addresses:[{IP:172.30.0.122 Hostname: NodeName:<nil> TargetRef:nil}] NotReadyAddresses:[] Ports:[{Name:https Port:6443 Protocol:TCP}] ServiceGroupId:1}]]

### DIFF
--- a/cmd/kube-controller-manager/app/controllermanager.go
+++ b/cmd/kube-controller-manager/app/controllermanager.go
@@ -254,6 +254,7 @@ func Run(c *config.CompletedConfig, stopCh <-chan struct{}) error {
 				go controllerContext.ResourceProviderNodeInformers[rpId].Informer().Run(controllerContext.Stop)
 			}
 		} else {
+			// There is no additional resource provider, fall back to single cluster case
 			controllerContext.ResourceProviderNodeInformers = make(map[string]coreinformers.NodeInformer, 1)
 			controllerContext.ResourceProviderNodeInformers["0"] = controllerContext.InformerFactory.Core().V1().Nodes()
 		}
--- a/cmd/kube-controller-manager/app/controllermanager.go
+++ b/cmd/kube-controller-manager/app/controllermanager.go
@@ -28,6 +28,7 @@ import (
 	"math/rand"
 	"net/http"
 	"os"
+	"strconv"
 	"time"
 
 	"k8s.io/client-go/datapartition"
@@ -47,6 +48,7 @@ import (
 	"k8s.io/apiserver/pkg/util/term"
 	cacheddiscovery "k8s.io/client-go/discovery/cached"
 	"k8s.io/client-go/informers"
+	coreinformers "k8s.io/client-go/informers/core/v1"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/metadata"
 	"k8s.io/client-go/metadata/metadatainformer"
@@ -237,6 +239,25 @@ func Run(c *config.CompletedConfig, stopCh <-chan struct{}) error {
 		}
 		saTokenControllerInitFunc := serviceAccountTokenControllerStarter{rootClientBuilder: rootClientBuilder}.startServiceAccountTokenController
 
+		// start client to resource providers
+		if len(c.ResourceProviderClients) > 0 {
+			controllerContext.ResourceProviderClients = make(map[string]clientset.Interface)
+			controllerContext.ResourceProviderNodeInformers = make(map[string]coreinformers.NodeInformer)
+			for i, rpClient := range c.ResourceProviderClients {
+				rpId := "rp" + strconv.Itoa(i)
+				resourceInformerFactory := informers.NewSharedInformerFactory(rpClient, 0)
+				resourceInformerFactory.Start(controllerContext.Stop)
+				controllerContext.ResourceProviderClients[rpId] = rpClient
+				controllerContext.ResourceProviderNodeInformers[rpId] = resourceInformerFactory.Core().V1().Nodes()
+				klog.V(2).Infof("Created the node informer %p from resource provider %s",
+					controllerContext.ResourceProviderNodeInformers[rpId].Informer(), rpId)
+				go controllerContext.ResourceProviderNodeInformers[rpId].Informer().Run(controllerContext.Stop)
+			}
+		} else {
+			controllerContext.ResourceProviderNodeInformers = make(map[string]coreinformers.NodeInformer, 1)
+			controllerContext.ResourceProviderNodeInformers["0"] = controllerContext.InformerFactory.Core().V1().Nodes()
+		}
+
 		if err := StartControllers(controllerContext, saTokenControllerInitFunc, NewControllerInitializers(controllerContext.LoopMode), unsecuredMux); err != nil {
 			klog.Fatalf("error starting controllers: %v", err)
 		}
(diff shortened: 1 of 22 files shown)


cmd/kube-controller-manager/app/controllermanager.go
```diff
@@ -237,6 +239,25 @@ func Run(c *config.CompletedConfig, stopCh <-chan struct{}) error {
...
+				resourceInformerFactory.Start(controllerContext.Stop)
+				controllerContext.ResourceProviderClients[rpId] = rpClient
+				controllerContext.ResourceProviderNodeInformers[rpId] = resourceInformerFactory.Core().V1().Nodes()
+				klog.V(2).Infof("Created the node informer %p from resource provider %s",
+					controllerContext.ResourceProviderNodeInformers[rpId].Informer(), rpId)
+				go controllerContext.ResourceProviderNodeInformers[rpId].Informer().Run(controllerContext.Stop)
+			}
+		} else {
+			controllerContext.ResourceProviderNodeInformers = make(map[string]coreinformers.NodeInformer, 1)
+			controllerContext.ResourceProviderNodeInformers["0"] = controllerContext.InformerFactory.Core().V1().Nodes()
```
h-w-chen, line 258: is there a reason go informers["0"].Informer().Run(...) not to start in this branch?

cmd/kube-controller-manager/app/options/options.go
```diff
@@ -412,11 +418,30 @@ func (s KubeControllerManagerOptions) Config(allControllers []string, disabledBy
 
 	eventRecorder := createRecorder(client, KubeControllerManagerUserAgent)
 
+	// get resource provider kube configs
+	var resourceProviderClients []clientset.Interface
+	if len(s.ResourceProviderKubeConfig) > 0 {
```
h-w-chen, line 423: what if 0? is this case be treated already somewhere else?

pkg/controller/service/service_controller.go
```diff
@@ -374,21 +381,27 @@ func (s *ServiceController) syncLoadBalancerIfNeeded(service *v1.Service, key st
 	return op, nil
 }
 
+// Will this work for multiple resource partition clusters?
```
h-w-chen, line 384: this comment seems temporary dev purpose; may be able to remove?

pkg/controller/service/service_controller.go
```diff
@@ -660,30 +673,36 @@ func getNodeConditionPredicate() corelisters.NodeConditionPredicate {
...
-	newHosts, err := s.nodeLister.ListWithPredicate(getNodeConditionPredicate())
-	if err != nil {
-		runtime.HandleError(fmt.Errorf("Failed to retrieve current set of nodes from node lister: %v", err))
-		return
+	allNewHosts := make([]*v1.Node, 0)
+	for _, nodeLister := range s.nodeListers {
+		newHosts, err := nodeLister.ListWithPredicate(getNodeConditionPredicate())
+		if err != nil {
+			// TODO - check error for RP not available, skip
+			runtime.HandleError(fmt.Errorf("Failed to retrieve current set of nodes from node lister: %v", err))
```
h-w-chen, line 681: nit: if it can hint out the culprit node lister/RP, the logging would be more useful.

hack/lib/common.sh
```diff
@@ -539,7 +544,18 @@ function kube::common::start_controller_manager {
           --cert-dir="${CERT_DIR}" \
           --default-network-template-path="${ARKTOS_NETWORK_TEMPLATE}" >"${CTLRMGR_LOG}" 2>&1 &
       else
-        KUBE_CONTROLLERS="*,-nodelifecycle,-nodeipam"
+        KUBE_CONTROLLERS="*,-daemonset,-nodelifecycle,-nodeipam,-ttl"
```
h-w-chen, line 547: are we going to start daemonset controller now?

h-w-chen (commented):
PR is cleaner. if it could split into two, with one including the 2 commits not to be reviewed and hence be able to merge directly, it will help reviewer to look at the other only.

cmd/kube-controller-manager/app/controllermanager.go
Sindica: The else branch is used for single cluster (TP/RP has the same api server). controllerContext.InformerFactory started somewhere else.

> is there a reason go informers["0"].Informer().Run(...) not to start in this branch?

cmd/kube-controller-manager/app/options/options.go
Sindica: If ResourceProviderKubeConfig is not provided, fallback to single cluster case.

> what if 0? is this case be treated already somewhere else?

pkg/controller/service/service_controller.go
Sindica: > this comment seems temporary dev purpose; may be able to remove?

I prefer leave it here since local cluster up does not start service controller. We don't have a way to test it locally. Will remove once it proves working

pkg/controller/service/service_controller.go
Sindica: > nit: if it can hint out the culprit node lister/RP, the logging would be more useful.

NodeLister is internally generated, it does not talk to RP directly so error will be internal only.

hack/lib/common.sh
Sindica: > are we going to start daemonset controller now?

Yes, we can start the controller. But we will disable daemonset test in 430.

pkg/util/node/nodecache_utils.go
```diff
@@ -0,0 +1,92 @@
...
+	coreinformers "k8s.io/client-go/informers/core/v1"
+	corelisters "k8s.io/client-go/listers/core/v1"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog"
+	"sync"
+	"time"
+)
+
+func GetNodeFromNodelisters(nodeListers map[string]corelisters.NodeLister, nodeName string) (*v1.Node, string, error) {
+	for rpId, nodeLister := range nodeListers {
```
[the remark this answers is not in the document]
Sindica: > this loop seems discontinue after the first iteration.

I think when node is not find in the lister, it will return NotFoundError (in TODO), this case, we will continue to search next node lister. Otherwise, the node is found and hence return

pkg/controller/service/service_controller.go
Sindica: Added TODO

h-w-chen (approved):
/lgtm
considering the loop premature return will be addressed properly

pkg/util/node/nodecache_utils.go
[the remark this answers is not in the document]
Sindica: Update to check NotFoundError.

PR #1017, opened 2021-03-11
Remove rp id from event handler arg

Sindica:
As we discussed, making event handler signature changes to get resource provider id is not idea. This change has two major changes:
1. Remove rpId from event handler
2. Change array of node informers/listers to map

Test done:
1TP/2RP, able to scheduler to both RPs. But pods for system tenant keeps getting restarted. It is out of scope for this PR.

```
ubuntu@ip-172-30-0-148:~$ kubectl get pods -A -T
TENANT   NAMESPACE     NAME                                  HASHKEY               READY   STATUS              RESTARTS   AGE
aaa      namespace-a   nginx-deployment-a-676f445fdf-jr8w8   8198298326532790695   1/1     Running             0          28m
system   default       nginx-deployment-676f445fdf-q8xmt     2990775166439868131   1/1     Running             0          17s
system   kube-system   kube-dns-6fdcc959b4-qsdpc             7036721288620706801   0/3     ContainerCreating   0          17s
```

### DIFF
--- a/pkg/scheduler/internal/cache/cache.go
+++ b/pkg/scheduler/internal/cache/cache.go
@@ -19,6 +19,8 @@ package cache
 
 import (
 	"fmt"
+	corelisters "k8s.io/client-go/listers/core/v1"
+	nodeutil "k8s.io/kubernetes/pkg/util/node"
 	"sync"
 	"time"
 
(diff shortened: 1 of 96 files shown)


yb01:
Probably it will be more cleaner to revert the previous PR that introduced such rpId interface changes first? and make changes on the original code base ?

pkg/scheduler/internal/cache/cache.go
```diff
@@ -554,13 +556,17 @@ func (cache *schedulerCache) AddNode(node *v1.Node, resourceProviderId string) e
...
 
-func (cache *schedulerCache) UpdateNode(oldNode, newNode *v1.Node, resourceProviderId string) error {
+func (cache *schedulerCache) UpdateNode(oldNode, newNode *v1.Node, nodeListers map[string]corelisters.NodeLister) error {
 	cache.mu.Lock()
 	defer cache.mu.Unlock()
 
 	n, ok := cache.nodes[newNode.Name]
 	if !ok {
 		nodeInfo := schedulernodeinfo.NewNodeInfo()
+		_, resourceProviderId, err := nodeutil.GetNodeFromNodelisters(nodeListers, newNode.Name)
```
h-w-chen, line 566: under what situation could the old.rpid be not the the same as new.rpid? if they might be different, this lookup is critical.

Sindica: This case is when node not found in our cache. Since the cache was flat, not found by node name means the node is not in the cache. Therefore, we won't get old rpId.

I think this case happen if we missed AddNode event. Perhaps just a safety guard.

PR #1021, opened 2021-03-12
Use kubeconfig to replace TP api server endpoints in kubelet

Sindica:
> 1. Changed Kubelet TP config from ip:port pairs to kubeconfig files

Previous kubelet starts as 

root      1179  4.5  0.3 1261568 105304 pts/9  Sl+  00:48   0:00 /home/ubuntu/go/src/arktos/_output/bin/hyperkube kubelet --v=3 --vmodule= --chaos-chance=0.0 --container-runtime=remote --hostname-override=ip-172-30-0-122 --cloud-provider= --cloud-config= --address=0.0.0.0 --kubeconfig /var/run/kubernetes/kubelet.kubeconfig --feature-gates=AllAlpha=false,WorkloadInfoDefaulting=true --cpu-cfs-quota=true --enable-controller-attach-detach=true --cgroups-per-qos=true --cgroup-driver= --cgroup-root= --eviction-hard=memory.available<100Mi,nodefs.available<10%,nodefs.inodesFree<5% --eviction-soft= --eviction-pressure-transition-period=1m --pod-manifest-path=/var/run/kubernetes/static-pods --fail-swap-on=false --authorization-mode=Webhook --authentication-token-webhook --client-ca-file=/var/run/kubernetes/client-ca.crt --cluster-dns=10.0.0.10 --cluster-domain=cluster.local --container-runtime-endpoint=containerRuntime,container,/run/containerd/containerd.sock;vmRuntime,vm,/run/virtlet.sock --runtime-request-timeout=2m --port=10250 **--tenant-servers=http://172.30.0.148:8080,http://172.30.0.24:8080**

Now kubelet starts as

root     29605  4.5  0.3 1392476 120020 pts/8  Sl+  03:18   0:04 /home/ubuntu/go/src/arktos/_output/local/bin/linux/amd64/hyperkube kubelet --v=3 --vmodule= --chaos-chance=0.0 --container-runtime=remote --hostname-override=ip-172-30-0-122 --cloud-provider= --cloud-config= --address=0.0.0.0 --kubeconfig /var/run/kubernetes/kubelet.kubeconfig --feature-gates=AllAlpha=false,WorkloadInfoDefaulting=true --cpu-cfs-quota=true --enable-controller-attach-detach=true --cgroups-per-qos=true --cgroup-driver= --cgroup-root= --eviction-hard=memory.available<100Mi,nodefs.available<10%,nodefs.inodesFree<5% --eviction-soft= --eviction-pressure-transition-period=1m --pod-manifest-path=/var/run/kubernetes/static-pods --fail-swap-on=false --authorization-mode=Webhook --authentication-token-webhook --client-ca-file=/var/run/kubernetes/client-ca.crt --cluster-dns=10.0.0.10 --cluster-domain=cluster.local --container-runtime-endpoint=containerRuntime,container,/run/containerd/containerd.sock;vmRuntime,vm,/run/virtlet.sock --runtime-request-timeout=2m --port=10250 **--tenant-server-kubeconfig=/var/run/kubernetes/tenant-server-kubelet0.kubeconfig,/var/run/kubernetes/tenant-server-kubelet1.kubeconfig**

> 2. Tested in 1TP/2RP, 2TP/2RP. Tenant deployment created running pods successfully. System tenant pods keep restarting. This is existing syndrome.

> 3. All local generated kubeconfig files no longer points to HA proxy.

### DIFF
--- a/cmd/kubelet/app/options/options.go
+++ b/cmd/kubelet/app/options/options.go
@@ -55,7 +55,7 @@ const defaultRootDir = "/var/lib/kubelet"
 // In general, please try to avoid adding flags or configuration fields,
 // we already have a confusingly large amount of them.
 type KubeletFlags struct {
-	TenantPartitionApiservers []string
+	TenantPartitionKubeConfig string
 	KubeConfig                string
 	BootstrapKubeconfig       string
 
@@ -362,7 +362,7 @@ func (f *KubeletFlags) AddFlags(mainfs *pflag.FlagSet) {
 
 	//TODO: post POC, move this to a separated kubeconfig file dedicated for tenant servers
 	//
-	fs.StringSliceVar(&f.TenantPartitionApiservers, "tenant-servers", f.TenantPartitionApiservers, "Comma separated string representing tenant api-server URLs.")
+	fs.StringVar(&f.TenantPartitionKubeConfig, "tenant-server-kubeconfig", f.TenantPartitionKubeConfig, "Comma separated string representing kubeconfig files point to tenant api servers.")
 	fs.StringVar(&f.KubeletConfigFile, "config", f.KubeletConfigFile, "The Kubelet will load its initial configuration from this file. The path may be absolute or relative; relative paths start at the Kubelet's current working directory. Omit this flag to use the built-in default configuration values. Command-line flags override configuration from this file.")
 	fs.StringVar(&f.KubeConfig, "kubeconfig", f.KubeConfig, "Path to kubeconfig files, specifying how to connect to  API servers. Providing --kubeconfig enables API server mode, omitting --kubeconfig enables standalone mode.")
 
--- a/cmd/kubelet/app/server.go
+++ b/cmd/kubelet/app/server.go
@@ -24,6 +24,8 @@ import (
 	"errors"
 	"fmt"
 	arktos "k8s.io/arktos-ext/pkg/generated/clientset/versioned"
+	"k8s.io/client-go/util/clientutil"
+	"k8s.io/kubernetes/cmd/genutils"
 	"k8s.io/kubernetes/pkg/kubelet/kubeclientmanager"
 	"math/rand"
 	"net"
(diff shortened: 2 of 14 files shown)


cmd/kubelet/app/options/options.go
```diff
@@ -55,7 +55,7 @@ const defaultRootDir = "/var/lib/kubelet"
 // In general, please try to avoid adding flags or configuration fields,
 // we already have a confusingly large amount of them.
 type KubeletFlags struct {
-	TenantPartitionApiservers []string
+	TenantPartitionKubeConfig string
```
yb01, line 58: will this still be the comma separated strings for multiple kubeconfig files? if, so, using plural might be more clear.

cmd/kubelet/app/server.go
```diff
@@ -580,25 +582,27 @@ func run(s *options.KubeletServer, kubeDeps *kubelet.Dependencies, stopCh <-chan
...
-
-		for i, tenantServer := range s.TenantPartitionApiservers {
-			for _, cfg := range clientConfigCopy.GetAllConfigs() {
-				cfg.Host = tenantServer
-				klog.V(6).Infof("clientConfigCopy.Host: %v", cfg.Host)
-			}
-
-			kubeDeps.KubeTPClients[i], err = clientset.NewForConfig(&clientConfigCopy)
+		klog.V(3).Infof("make kubeDeps.KubeTPClients based on TenantPartitionKubeConfig arg: %v", s.TenantPartitionKubeConfig)
+		if len(s.TenantPartitionKubeConfig) == 0 {
```
yb01, line 586: if the stringVar is used for the kubeconfig parameter, better just check if it is empty "" string as default. functionality the same. directly compare string empty is clearer -- in contrast the len(s) to check array size as it was before as TenantPartitionApiServers.

cmd/kubelet/app/options/options.go
Sindica: Yes, currently it take comma separated strings. Since in scheduler and KCM, all kubeconfig for RP/TP are singular, keep it consistent.

cmd/kubelet/app/server.go
Sindica: https://stackoverflow.com/questions/18594330/what-is-the-best-way-to-test-for-an-empty-string-in-go#:~:text=As%20pointed%20out%20in%20Timmmm's,identical%20code%20in%20both%20cases.&text=because%20the%20semantics%20is%20clearly%20equal.&text=Checking%20for%20length%20is%20a,that%20is%20also%20only%20whitespace.

Historically, check length is faster than compare with string. But golang has optimization to generate same code in this case. So it becomes a matter of taste.

Scheduler check "" and KCM checks length. I will keep this implementation.

hack/lib/common.sh
```diff
@@ -706,6 +707,22 @@ function kube::common::start_kubelet {
...
     fi
 
+    KUBELET_FLAGS="--tenant-server-kubeconfig="
+    if [[ "${IS_SCALE_OUT}" == "true" ]] && [ "${IS_RESOURCE_PARTITION}" == "true" ]; then
+      serverCount=${#TENANT_SERVERS[@]}
+      kubeconfig_filename="tenant-server-kubelet"
+      for (( pos=0; pos<${serverCount}; pos++ ));
+      do
+        # here generate kubeconfig for remote API server. Only work in non secure mode for now
+        kube::util::write_client_kubeconfig "${CONTROLPLANE_SUDO}" "${CERT_DIR}" "" "${TENANT_SERVERS[${pos}]}" "${API_PORT}" tenant-server-kubelet "" "http"
```
yb01, line 717: to clarify: how this will be updated if secure port is used , with cert and token needs to be set ?

cmd/kubelet/app/server.go
yb01: that is fine. i was not worried about this micro optimization here. it was due to the before this change, the parameter is slice so check length of slice indicates how many kubeconfigs are passed in. 
we can resolve. no sweat.

yb01 (approved):
/lgtm

hack/lib/common.sh
Sindica: Not sure how certs can be used in local cluster up and also allowing either TP or RP starts arbitrarily. We might need to use token.

> to clarify: how this will be updated if secure port is used , with cert and token needs to be set ?

Issue #1020, opened 2021-03-11
system tenant pods get restarted constantly


h-w-chen:
This issue is in 430 PC branch; not in master yet.

**What happened**:
(based on YinhH's testing) 1TP/2RP, is able to scheduler to both RPs. But pods for system tenant keeps getting restarted.
Actually it was reproducible at 1TP/1RP env

```
ubuntu@ip-172-30-0-148:~$ kubectl get pods -A -T
TENANT   NAMESPACE     NAME                                  HASHKEY               READY   STATUS              RESTARTS   AGE
aaa      namespace-a   nginx-deployment-a-676f445fdf-jr8w8   8198298326532790695   1/1     Running             0          28m
system   default       nginx-deployment-676f445fdf-q8xmt     2990775166439868131   1/1     Running             0          17s
```

**What you expected to happen**:
system/default/nginx pod in stably running state
 
**How to reproduce it (as minimally and precisely as possible)**:
1. start 1TP/1RP test env, after the RP node was registered itself as the sole node
```
$ kubectl --kubeconfig /var/run/kubernetes/scheduler.kubeconfig get nodes -o wide
NAME               STATUS   ROLES    AGE     VERSION   INTERNAL-IP     EXTERNAL-IP   OS-IMAGE             KERNEL-VERSION   CONTAINER-RUNTIME
ip-172-31-48-165   Ready    <none>   2m18s   v0.7.0    172.31.48.165   <none>        Ubuntu 20.04.2 LTS   5.4.0-1038-aws   containerd://1.3.3-0ubuntu2.2
```
2. start system tenant nginx pod
```
$ kubectl --kubeconfig /var/run/kubernetes/scheduler.kubeconfig run nn1 --image=nginx
```
3. start test1 tenant nginx pod 
```
$ kubectl --kubeconfig /var/run/kubernetes/scheduler.kubeconfig create tenant test1
$ kubectl --kubeconfig /var/run/kubernetes/scheduler.kubeconfig run nn2 --image=nginx --tenant test1
```
4. after 2-3 minutes,
```
$ kubectl --kubeconfig /var/run/kubernetes/scheduler.kubeconfig get pod -A -T -o wide
TENANT   NAMESPACE     NAME                        HASHKEY               READY   STATUS    RESTARTS   AGE     IP           NODE               NOMINATED NODE   READINESS GATES
[...]

PR #1025, opened 2021-03-13
Fix pod GC issue in scale out case

Sindica:
This change fix this issue: https://github.com/CentaurusInfra/arktos/issues/1020

Before code fix, GC list nodes every 20s and could not find node, hence recycle pods every 20s.
```
ubuntu@ip-172-30-0-148:~$ kubectl get pods -A
NAMESPACE     NAME                                HASHKEY               READY   STATUS    RESTARTS   AGE
default       nginx-deployment-676f445fdf-t2lnw   5880710705033174106   1/1     Running   0          55s
kube-system   kube-dns-6fdcc959b4-66ssw           1357571739321344899   2/3     Running   1          110s
```

After stopped node where pod t2lnw was created, it got recreated in a different host:
```
ubuntu@ip-172-30-0-148:~$ kubectl get pods -A
NAMESPACE     NAME                                HASHKEY               READY   STATUS    RESTARTS   AGE
default       nginx-deployment-676f445fdf-8h257   8574656549163554292   1/1     Running   0          8s
kube-system   kube-dns-6fdcc959b4-66ssw           1357571739321344899   2/3     Running   1          2m28s
```

Log:
I0313 00:59:24.509731    1412 gc_controller.go:199] Forced deletion of orphaned Pod system/default/nginx-deployment-676f445fdf-t2lnw succeeded
I0313 00:59:24.512072    1412 event.go:259] Event(v1.ObjectReference{Kind:"ReplicaSet", Namespace:"default", Name:"nginx-deployment-676f445fdf", UID:"03d8692b-2996-45b5-8606-96d6e6c7faf3", APIVersion:"apps/v1", ResourceVersion:"847038162319966209", FieldPath:"", Tenant:"system"}): type: 'Normal' reason: 'SuccessfulCreate' Created pod: nginx-deployment-676f445fdf-8h257

Also works for pod for non system tenant:
```
ubuntu@ip-172-30-0-148:~$ kubectl get pods -A -T
TENANT   NAMESPACE     NAME                                  HASHKEY               READY   STATUS             RESTARTS   AGE
aaa      namespace-a   nginx-deployment-a-676f445fdf-nk5ft   3243237195972317040   1/1     Running            0          61s
system   default       nginx-deployment-676f445fdf-8h257     8574656549163554292   1/1     Running            0          5m2s
system   kube-system   kube-dns-6fdcc959b4-66ssw             1357571739321344899   1/3     CrashLoopBackOff   9          7m22s
```

Manually stopped RP that hosts the pods, pods got recreated automatically in another RP:
```
ubuntu@ip-172-30-0-148:~$ kubectl get pods -A -T
TENANT   NAMESPACE     NAME                                  HASHKEY               READY   STATUS    RESTARTS   AGE
aaa      namespace-a   nginx-deployment-a-676f445fdf-szqx6   5738940787212938146   1/1     Running   0          60s
system   default       nginx-deployment-676f445fdf-tbw9f     9007118871540703210   1/1     Running   0          20s
system   kube-system   kube-dns-6fdcc959b4-gnf9t             6951555085537816387   2/3     Running   0          60s
```

### DIFF
--- a/hack/arktos-up-scale-out-poc.sh
+++ b/hack/arktos-up-scale-out-poc.sh
@@ -23,6 +23,10 @@ IS_RESOURCE_PARTITION=${IS_RESOURCE_PARTITION:-"false"}
 
 # proxy is still used to start cloud KCM. Also useful for system tenant requests.
 # However, don't use proxy to query node list as there is no aggregator for multiple RPs
+# As we are tring to remove HA proxy, SCALE_OUT_PROXY_IP and SCALE_OUT_PROXY_PORT are both no longer
+#  required in local cluster up. When they are not provided, they will be default to API server host
+#  ip and port. If you need proxy to be running, please set environment variable SCALE_OUT_PROXY_IP
+#  and SCALE_OUT_PROXY_PORT explicitly.
 SCALE_OUT_PROXY_IP=${SCALE_OUT_PROXY_IP:-}
 SCALE_OUT_PROXY_PORT=${SCALE_OUT_PROXY_PORT:-}
 TENANT_SERVER=${TENANT_SERVER:-}
--- a/hack/arktos-up-scale-out-poc.sh
+++ b/hack/arktos-up-scale-out-poc.sh
@@ -15,12 +15,16 @@
 # limitations under the License.
 
 # set up variables
+KUBE_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
+echo KUBE_ROOT ${KUBE_ROOT}
+source "${KUBE_ROOT}/hack/lib/common-var-init.sh"
+
 IS_RESOURCE_PARTITION=${IS_RESOURCE_PARTITION:-"false"}
 
 # proxy is still used to start cloud KCM. Also useful for system tenant requests.
 # However, don't use proxy to query node list as there is no aggregator for multiple RPs
 SCALE_OUT_PROXY_IP=${SCALE_OUT_PROXY_IP:-}
-SCALE_OUT_PROXY_PORT=${SCALE_OUT_PROXY_PORT:-"8888"}
+SCALE_OUT_PROXY_PORT=${SCALE_OUT_PROXY_PORT:-}
 TENANT_SERVER=${TENANT_SERVER:-}
 RESOURCE_SERVER=${RESOURCE_SERVER:-}
 IS_SCALE_OUT=${IS_SCALE_OUT:-"true"}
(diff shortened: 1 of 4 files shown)


hack/arktos-up-scale-out-poc.sh
```diff
@@ -15,12 +15,16 @@
...
+echo KUBE_ROOT ${KUBE_ROOT}
+source "${KUBE_ROOT}/hack/lib/common-var-init.sh"
+
 IS_RESOURCE_PARTITION=${IS_RESOURCE_PARTITION:-"false"}
 
 # proxy is still used to start cloud KCM. Also useful for system tenant requests.
 # However, don't use proxy to query node list as there is no aggregator for multiple RPs
 SCALE_OUT_PROXY_IP=${SCALE_OUT_PROXY_IP:-}
-SCALE_OUT_PROXY_PORT=${SCALE_OUT_PROXY_PORT:-"8888"}
+SCALE_OUT_PROXY_PORT=${SCALE_OUT_PROXY_PORT:-}
```
yb01, line 27: any reason the default 8888 port is removed ?

pkg/controller/podgc/gc_controller.go
```diff
@@ -143,13 +146,43 @@ func (gcc *PodGCController) gcTerminated(pods []*v1.Pod) {
 func (gcc *PodGCController) gcOrphaned(pods []*v1.Pod) {
 	klog.V(4).Infof("GC'ing orphaned")
 	// We want to get list of Nodes from the etcd, to make sure that it's as fresh as possible.
-	nodes, err := gcc.kubeClient.CoreV1().Nodes().List(metav1.ListOptions{})
-	if err != nil {
+
+	// get nodes from resource provider clients
+	allRpNodes := make(map[string]*v1.NodeList, len(gcc.resourceProviderClients))
```
yb01, line 151: do we have some analysis on how often this function to be be called in our test ( or some other env ) ? if this is called frequently it would be impact GC perf.

yb01 (approved):
/lgtm to 430 branch

hack/arktos-up-scale-out-poc.sh
Sindica: In the case SCALE_OUT_PROXY_IP was not provided and default to local host ip, 8888 won't work

pkg/controller/podgc/gc_controller.go
Sindica: > do we have some analysis on how often this function to be be called in our test ( or some other env ) ? if this is called frequently it would be impact GC perf.

It's called every 20s.

yb01: good to know. in this case, then it probably cannot be implemented as such. likely a list/watch approach is needed  here.

hack/arktos-up-scale-out-poc.sh
yb01: I saw the code below in line 44. it seems the reason is that the proxy port and Api Server port is mixed. hence is won't work. probably we should not mix them so each port severs for its own purpose.

pkg/controller/podgc/gc_controller.go
Sindica: > good to know. in this case, then it probably cannot be implemented as such. likely a list/watch approach is needed here.

This is a K8s existing implementation. I agree list/watch might have better performance but this code has comment specifically said it wanted real time data. I would rather keep it the original logic before we found there are performance issue.

hack/arktos-up-scale-out-poc.sh
Sindica: > I saw the code below in line 44. it seems the reason is that the proxy port and Api Server port is mixed. hence is won't work. probably we should not mix them so each port severs for its own purpose.

We were trying to remove Proxy here totally. The only reason that we haven't totally removed SCALE_OUT_PROXY_IP and SCALE_OUT_PROXY_PORT here is because somehow cloud KCM is still using it, which might be blindly copied from local cluster up/arktos up. Will remove cloud KCM once we found it is not needed in perf test. Make proxy port and api server port separate meaning in our local test we must set SCALE_OUT_PROXY_PORT explicitly. This is not current intention.

pkg/controller/podgc/gc_controller.go
yb01: that should be OK for the 430 we can always revisit.

yb01 (approved):
/lgtm

hack/arktos-up-scale-out-poc.sh
Sindica: Added comment

PR #1027, opened 2021-03-13
Fix issue 1023

chenqianfzh:
Fixing https://github.com/CentaurusInfra/arktos/issues

Verified with a successful 100-node scale-out 1TP-1RP density test run, where the maximal number of pods should be 3000+. Here is the Prometheus graph for metric etcd_object_count of pods, where the peak number reaches 3430. 

image


sonyafenge:
tried 500 nodes perf-test, etcd_obejct_count for pods displayed correctly. density test run successfully.
image

Sindica:
Thanks for fixing this Qian. Can you please verify in scale up case as well?

chenqianfzh:
> Thanks for fixing this Qian. Can you please verify in scale up case as well?

I believe scale-up was not impacted by this bug, which I checked during the bug fix. The cause is that scale-up is about system tenant only.

sonyafenge:
also verified scale-up with 500 nodes, etcd_obejct_count for pods displayed correctly. Density test run successfully.
image

sonyafenge (approved):
/lgtm

PR #1028, opened 2021-03-16
Enable serviceaccount & serviceaccount-token controller in RP local cluster

Sindica:
Tested in 2TP/2RPs, tenant pods created.


yb01 (approved):
/lgtm

PR #1029, opened 2021-03-16
Fix kube-proxy stop running issue in local scale up cluster

Sindica:
Fix kube-proxy stop running issue in local scale up cluster

### DIFF
--- a/hack/lib/common.sh
+++ b/hack/lib/common.sh
@@ -798,12 +798,18 @@ EOF
     fi >>/tmp/kube-proxy.yaml
 
     kube::common::generate_kubeproxy_certs
+    local port=${API_SECURE_PORT}
+    local protocol="https"
+    if [[ "${IS_SCALE_OUT}" == "true" ]]; then
+      port=${API_PORT}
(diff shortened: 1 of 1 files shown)


hack/lib/common.sh
```diff
@@ -798,12 +798,18 @@ EOF
     fi >>/tmp/kube-proxy.yaml
 
     kube::common::generate_kubeproxy_certs
+    local port=${API_SECURE_PORT}
+    local protocol="https"
+    if [[ "${IS_SCALE_OUT}" == "true" ]]; then
```
yb01, line 803: probably a comment with TODO can help future work to enable the secure connection on scale-out env in local-up.

yb01 (approved):
/lgtm

hack/lib/common.sh
Sindica: I think whoever reenable SSL mode in local cluster up will check all the kube config and override values. There is no plan to do so in the near future, so prefer not leave a TODO here and there.

PR #1030, opened 2021-03-16
Pickup critical bugfix from master to 430 branch

yb01:
**What type of PR is this?**

> /kind bug

kube-up.sh failed in 430 branch due to missing a critical fix for a regression in master

Fixes #

### DIFF
--- a/pkg/kubelet/kubelet.go
+++ b/pkg/kubelet/kubelet.go
@@ -450,7 +450,7 @@ func NewMainKubelet(kubeCfg *kubeletconfiginternal.KubeletConfiguration,
 	nodeIndexer := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{})
 	if kubeDeps.HeartbeatClient != nil {
 		fieldSelector := fields.Set{api.ObjectNameField: string(nodeName)}.AsSelector()
-		nodeLW := cache.NewListWatchFromClient(kubeDeps.HeartbeatClient.CoreV1(), "nodes", metav1.NamespaceAll, fieldSelector)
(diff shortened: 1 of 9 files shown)


pkg/kubelet/kubelet.go
```diff
@@ -450,7 +450,7 @@ func NewMainKubelet(kubeCfg *kubeletconfiginternal.KubeletConfiguration,
 	nodeIndexer := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{})
 	if kubeDeps.HeartbeatClient != nil {
 		fieldSelector := fields.Set{api.ObjectNameField: string(nodeName)}.AsSelector()
-		nodeLW := cache.NewListWatchFromClient(kubeDeps.HeartbeatClient.CoreV1(), "nodes", metav1.NamespaceAll, fieldSelector)
+		nodeLW := cache.NewListWatchFromClientWithMultiTenancy(kubeDeps.HeartbeatClient.CoreV1(), "nodes", metav1.NamespaceAll, fieldSelector, metav1.TenantNone)
```
Sindica, line 453: NamespaceAll but TenantNone. This combination is very weird

yb01: this is cherry-picked from the master branch. Qian, can you comment on this ?

chenqianfzh: I agree. It is better to use "metav1.NamespaceNone" instead of "metav1.NamespaceAll". They are both "" under the hood, but "metav1.NamespaceNone" makes more sense. 

Feel free to make the changes.

chenqianfzh (approved):

pkg/kubelet/kubelet.go
yb01: good. in this, i would prefer to fix. 
just pushed a name change to use NamespaceNone here.

PR #1032, opened 2021-03-17
perftest tool: bypass accessing RPs

h-w-chen:
this PR targets 430 POC.

**What type of PR is this?**
/kind feature

It introduces to perf command tool (clusterloader) a new env var controlling whether to access RPs to list nodes, or not.

In default mode (no chaos monkey), clusterloader does not really care much about node details, other than the number of nodes. This PR makes option for it not to access RPs, so to access TP directly, w/o proxy.

When no to access RPs, chaos monkey must not be enabled (it is disabled by default). The change to support enabling chaos monkey will be considered later (as chaos monley is not being used in current stage of Arktos scalability test)

Yes
```release-note
run-e2e.sh could use TP-only kubeconfig (--kubeconfig) parameter; 
run-e2e.sh must provide --nodes parameter with proper value (unless RP_ACCESS set to true and --kubeconfig points to proxy kubeconfig)
```

### DIFF
--- a/perf-tests/clusterloader2/cmd/clusterloader.go
+++ b/perf-tests/clusterloader2/cmd/clusterloader.go
@@ -1,5 +1,6 @@
 /*
 Copyright 2018 The Kubernetes Authors.
+Copyright 2020 Authors of Arktos - file modified.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


perf-tests/clusterloader2/cmd/clusterloader.go
```diff
@@ -244,8 +250,11 @@ func main() {
 		klog.Exitf("Cannot create report directory: %v", err)
 	}
 
-	if err = util.LogClusterNodes(mclient.GetClient()); err != nil {
-		klog.Errorf("Nodes info logging error: %v", err)
+	if clusterLoaderConfig.ClusterConfig.RPAccess {
+		// todo: use direct access to RPs w/o proxy
```
yb01, line 254: if, assuming this direct access to the RP won't needed, do we still need this changes in this PR ?

yb01 (approved):
/lgtm to 430 branch

perf-tests/clusterloader2/cmd/clusterloader.go
h-w-chen: in case the proxy kubeconfig is used, it prints out the nodes (for logging purpose) which is the current behavior it tries to keep.

sonyafenge (approved):
/lgtm

PR #1033, opened 2021-03-17
Add some log to perf-test tool.

Sindica:


yb01:
what is the purpose of adding this log ? I mean, to provide info to catch the possible last execution before the perf tool exits/beingKilled/crashed, the log is too high.

to just log the test tenant, it is fine.

h-w-chen (approved):
/lgtm

PR #1034, opened 2021-03-18
Double GC/RS controller QPS. Guarded with feature gate.

Sindica:
Double replicaset & garbage controller QPS. Guarded with feature gates.


yb01 (approved):
/lgtm

PR #1037, opened 2021-03-19
Double GC/RS controller QPS. Guarded with feature gate.

Sindica:


yb01 (approved):
/lgtm.
cherry-picked reviewed and tested code from 430 branch

PR #1051, opened 2021-03-25
bug fix 1048: api server tsl handshake error with proxy server

### LINKED ISSUE #1048: haproxy ssl check causes api server "TSL handshake error"
**What happened**:
currently we configured haproxy to do service health check on backend with "ssl check", which, per this doc, enforced TSL on the check traffic, but the ports to the haproxy is not the SSL enabled ports, hence the error on api server side.

https://cbonte.github.io/haproxy-dconv/2.4/configuration.html#check-ssl

```
I0325 20:52:41.549556       1 log.go:172] http: TLS handshake error from 34.70.13.189:60912: write tcp 10.40.0.6:443->34.70.13.189:60912: write: connection reset by peer
I0325 20:52:41.549696       1 log.go:172] http: TLS handshake error from 34.70.13.189:60914: write tcp 10.40.0.6:443->34.70.13.189:60914: write: connection reset by peer
I0325 20:52:41.551603       1 log.go:172] http: TLS handshake error from 34.70.13.189:60918: write tcp 10.40.0.6:443->34.70.13.189:60918: write: connection reset by peer
```

**Environment**:
- OS (e.g: `cat /etc/os-release`):
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

yb01:
**What type of PR is this?**
> /kind bug

the haproxy service periodically performs health check with the backend API servers. the current configuration for this backend health check is check-ssl, which enforces full TSL traffic for the check. this causes a lot log entries in the api server due to the source where the health check originated is not the 443 port of the haporxy.

currently one can either specify the "source <proxy server Ip>:443" or just perform TCP checks to backend servers. this PR takes the later one to not occupy the 443 port too much just in case to affect testing throughputs.

Fixes #1048 

Tested with 100 node under  POC 430 branch.

### DIFF
--- a/cmd/haproxy-cfg-generator/cfg_generator.go
+++ b/cmd/haproxy-cfg-generator/cfg_generator.go
@@ -212,8 +212,9 @@ func get_backends(config *Config) string {
 	}
 
 	if config.tlsMode == BRIDGING {
-		tp_backend = "backend tenant_api_%v\n    server tp_%v %v:%v maxconn %v ssl check ca-file /etc/haproxy/ca.crt\n\n"
-		rp_backend = "backend resource_api\n    server rp %v:%v maxconn %v ssl check ca-file /etc/haproxy/ca.crt\n\n"
+		// perform a TCP check for backend health
+		tp_backend = "backend tenant_api_%v\n    option tcp-check\n    tcp-check connect\n    server tp_%v %v:%v maxconn %v ssl check ca-file /etc/haproxy/ca.crt\n\n"
+		rp_backend = "backend resource_api\n    option tcp-check\n    tcp-check connect\n    server rp %v:%v maxconn %v ssl check ca-file /etc/haproxy/ca.crt\n\n"
 	} else {
 		tp_backend = "backend tenant_api_%v\n    server tp_%v %v:%v maxconn %v\n\n"
 		rp_backend = "backend resource_api\n    server rp %v:%v maxconn %v\n\n"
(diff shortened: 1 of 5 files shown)


cmd/haproxy-cfg-generator/cfg_generator.go
```diff
@@ -212,8 +212,9 @@ func get_backends(config *Config) string {
 	}
 
 	if config.tlsMode == BRIDGING {
-		tp_backend = "backend tenant_api_%v\n    server tp_%v %v:%v maxconn %v ssl check ca-file /etc/haproxy/ca.crt\n\n"
-		rp_backend = "backend resource_api\n    server rp %v:%v maxconn %v ssl check ca-file /etc/haproxy/ca.crt\n\n"
+		// perform a TCP check for backend health
+		tp_backend = "backend tenant_api_%v\n    option tcp-check\n    tcp-check connect\n    server tp_%v %v:%v maxconn %v ssl check ca-file /etc/haproxy/ca.crt\n\n"
+		rp_backend = "backend resource_api\n    option tcp-check\n    tcp-check connect\n    server rp %v:%v maxconn %v ssl check ca-file /etc/haproxy/ca.crt\n\n"
 	} else {
```
XiaoningDing, line 218: For the configs in the "else" clause, do we need health check or by default it's already tcp-check?

XiaoningDing (approved):

cmd/haproxy-cfg-generator/cfg_generator.go
yb01: per the HAPROXY config doc,  if not set, it will not perform health check and assumed the backend service is always available. we probably will need add if we continue to use the insecure mode

"- when not set, no health checking is performed, and the server is always
    considered available."

yb01: we will add when needed. from now on, we will be testing secure mode as main stream of testing.

PR #1053, opened 2021-03-26
add logging to grpc message sends to mizar

clu2xlu:
**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Add more logging for Mizar Arktos integration project to track the messages that was sent to Mizar through grpc. 

Fixes #
N/A

```release-note

```

### DIFF
--- a/pkg/controller/mizar/util.go
+++ b/pkg/controller/mizar/util.go
@@ -16,6 +16,7 @@ package mizar
 import (
 	"encoding/json"
 	"strconv"
+	"k8s.io/klog"
 
 	v1 "k8s.io/api/core/v1"
 )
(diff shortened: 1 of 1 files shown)


pkg/controller/mizar/util.go
```diff
@@ -80,6 +84,9 @@ func ConvertToPodContract(pod *v1.Pod) *BuiltinsPodMessage {
 		network = ""
 	}
 
+	klog.Infof("Pod Name: %s, HostIP: %s, Namespace: %s, Tenant: %s, Arktos network: %s",
```
XiaoningDing, line 87: Do we need to set a log level for these logs?

XiaoningDing (approved):

pkg/controller/mizar/util.go
clu2xlu: Good call, I will add log level

vinaykul: @clu2xlu Are we still planning to fix this and update the PR?

PR #1070, opened 2021-04-08
Add log for percentageOfNodesToScore vs. evaluated node

Sindica:
K8s document said percentageOfNodesToScore can be set as lower than 5%. But the code seems only accept 5% as lowest threshold. Adding log to confirm.

### DIFF
--- a/pkg/scheduler/core/generic_scheduler.go
+++ b/pkg/scheduler/core/generic_scheduler.go
@@ -456,6 +456,8 @@ func (g *genericScheduler) numFeasibleNodesToFind(numAllNodes int32) (numNodes i
 		return minFeasibleNodesToFind
 	}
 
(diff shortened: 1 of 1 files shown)


pkg/scheduler/core/generic_scheduler.go
```diff
@@ -456,6 +456,8 @@ func (g *genericScheduler) numFeasibleNodesToFind(numAllNodes int32) (numNodes i
 		return minFeasibleNodesToFind
 	}
 
+	klog.V(2).Infof("Get # of evaluated node. Total nodes %v, percentageOfNodesToScore %v, adaptivePercentage %v, node to evaluate %v",
```
yb01, line 459: for small cluster this log will not shown. but for our current investigation, it is OK.

yb01 (approved):
/lgtm

PR #1077, opened 2021-04-12
add trace around scheduler queue operatios

yb01:

last week per our testing, it indicates some the pod stays in the queue for more than 1 seconds, which indicates either scheuler logic ( predicate and proritize) delays or the queue OP delays. add more trace to future identify them.

also, add one log to ensure affinity check is NOT in our case.

### DIFF
--- a/pkg/scheduler/algorithm/predicates/metadata.go
+++ b/pkg/scheduler/algorithm/predicates/metadata.go
@@ -399,7 +399,7 @@ func getTPMapMatchingExistingAntiAffinity(pod *v1.Pod, nodeInfoMap map[string]*s
 			return
 		}
 		existingPods := nodeInfo.PodsWithAffinity()
-		// klog.V(2).Infof("Node %v, Pods with Affinity: %v", nodeInfo.Node().Name, len(existingPods))
+		klog.V(6).Infof("Node %v, Pods with Affinity: %v", nodeInfo.Node().Name, len(existingPods))
 		for _, existingPod := range existingPods {
 			existingPodTopologyMaps, err := getMatchingAntiAffinityTopologyPairsOfPod(pod, existingPod, node)
 			if err != nil {
@@ -411,7 +411,7 @@ func getTPMapMatchingExistingAntiAffinity(pod *v1.Pod, nodeInfoMap map[string]*s
 		}
 	}
 
-	klog.V(2).Infof("Number of nodes: %v", len(allNodeNames))
+	klog.V(6).Infof("Number of nodes: %v", len(allNodeNames))
 	// TODO: make this configurable so we can test perf impact when increasing or decreasing concurrency
 	//       on a 96 core machine, it can be much more than 16 concurrent threads to run the processNode function
 	workqueue.ParallelizeUntil(ctx, 16, len(allNodeNames), processNode)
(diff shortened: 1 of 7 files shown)


yb01:
in private testing

pkg/scheduler/algorithm/predicates/metadata.go
```diff
@@ -401,13 +401,17 @@ func getTPMapMatchingExistingAntiAffinity(pod *v1.Pod, nodeInfoMap map[string]*s
 		existingPods := nodeInfo.PodsWithAffinity()
 		klog.V(6).Infof("Node %v, Pods with Affinity: %v", nodeInfo.Node().Name, len(existingPods))
 		for _, existingPod := range existingPods {
+			// in current arktos perf runs, this should not be reached
+			klog.V(2).Infof("DEBUG: this should not be reached. print out the pod: %v-%v %v", existingPod.Namespace, existingPod.Name)
```
h-w-chen, line 405: extraneous %v?

pkg/scheduler/algorithm/predicates/metadata.go
```diff
@@ -401,13 +401,17 @@ func getTPMapMatchingExistingAntiAffinity(pod *v1.Pod, nodeInfoMap map[string]*s
...
 			existingPodTopologyMaps, err := getMatchingAntiAffinityTopologyPairsOfPod(pod, existingPod, node)
 			if err != nil {
 				catchError(err)
 				cancel()
 				return
 			}
-			appendTopologyPairsMaps(existingPodTopologyMaps)
+			if existingPodTopologyMaps!=nil {
```
h-w-chen, line 412: what if the initial map is nil/empty?

yb01: this should end up with empty topologyMap, sicne the appendTopologyPairsMap is not called.

pkg/scheduler/algorithm/predicates/metadata.go
yb01: thanks. fixed.

pkg/scheduler/internal/queue/scheduling_queue.go
```diff
@@ -402,12 +411,16 @@ func (p *PriorityQueue) Pop() (*v1.Pod, error) {
...
+
 	obj, err := p.activeQ.Pop()
 	if err != nil {
 		return nil, err
 	}
 	pInfo := obj.(*framework.PodInfo)
 	p.schedulingCycle++
+
+	end := time.Now()
+	klog.V(2).Infof("Pod pod %v-$v, duration %v ms", pInfo.Pod.Namespace, pInfo.Pod.Name, end.Sub(start).Milliseconds())
```
Sindica, line 423: Pop pod?

yb01: fixed. was fixed on my test box with the logging % instead of $, but not back add to the dev machine where the commit was made.

pkg/scheduler/algorithm/predicates/metadata.go
h-w-chen: my concern is, normally a map starts as the empty one (unless otherwise initialized), which will make the code never be called. not sure whether that is the right intention.

yb01: this is exactly the code change intended to do. i.e only call this function appendTopologyPairsMaps() when the existingPodTopologyMaps returned from the previous function is not nil, to avoid unnecessary calls, i.e. save a lock-unlock OP per existing POD in the cluster -- this could be huge considering there are 450K pods in the end for a 15k node run.

h-w-chen: ok got it. thanks for the explanation.

yb01:
closed. obsolete PR, not needed anymore.

PR #1092, opened 2021-05-10
Pick up recent perf test changes

Sindica:


h-w-chen (approved):
lgtm

cluster/gce/gci/configure.sh
```diff
@@ -367,6 +367,8 @@ function install-flannel-yml {
   local -r flannel_dir="${KUBE_HOME}/flannel"
   mkdir -p "${flannel_dir}"
   mv "${KUBE_HOME}/kube-flannel.yml" "${flannel_dir}"
+  echo "change docker registry to gcr.io"
+  sed -i 's+quay.io/coreos+gcr.io/workload-controller-manager+g' ${flannel_dir}/kube-flannel.yml
```
h-w-chen, line 371: if we can make this more robust against the varying flannel versions, that would be better.

yb01: this is already in the merge2mater PR, which is in master with the PR already

Sindica: This is a merge to 530 POC branch. We did not merge those changes there. We will not use this branch any more. Just trying to close issues and merge request.

PR #1108, opened 2021-06-21
set user agent for hollow nodes and perf test clients

yb01:
Explicitly set user agent for the hollow node and the perf-test client so that in log and in prometheus they can be differentiated for perf analysis.

### DIFF
--- a/cmd/kubemark/hollow-node.go
+++ b/cmd/kubemark/hollow-node.go
@@ -186,7 +186,7 @@ func run(config *hollowNodeConfig) {
 			klog.Fatalf("Failed to create a client config: %v. Exiting.", err)
 		}
 
-		clientFromConfig, err := clientset.NewForConfig(cfg)
+		clientFromConfig, err := clientset.NewForConfig(restclient.AddUserAgent(cfg, "hollow-node"))
 		if err != nil {
 			klog.Fatalf("Failed to create a ClientSet: %v. Exiting.", err)
 		}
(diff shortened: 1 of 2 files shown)


cmd/kubemark/hollow-node.go
```diff
@@ -186,7 +186,7 @@ func run(config *hollowNodeConfig) {
 			klog.Fatalf("Failed to create a client config: %v. Exiting.", err)
 		}
 
-		clientFromConfig, err := clientset.NewForConfig(cfg)
+		clientFromConfig, err := clientset.NewForConfig(restclient.AddUserAgent(cfg, "hollow-node"))
```
Sindica, line 189: I remember user-agent for hollow node is kubemark.

yb01: they are all with the kubecfg: for example:

/api/v1/tenants/monkey/namespaces/fy67ad-testns/pods/latency-deployment-4-6b5ccbc79d-qkxbt/status","verb":"patch","user":{"username":"kubecfg"

Sindica: Where is the user agent shown? Only in audit log? As I don't see it in Prometheus data

request-promethus

yb01: they are in the api server audit logs. you can try to grep one PODID to get all the audit-ed operations on the pod. username is tracked to see who ( which components ) makes the change or requests.

Sindica (approved):

PR #1111, opened 2021-06-28
CR only: remove unnecessary locks in List/ListPartition

yb01:
remove unneeded locks in listPartition(). also changed trace to get better ideas on the Get() and transform() delays
```
sonyali@sonyadev4:~/logs/perf-test/gce-15000/arktos/arktos-062221-15k/arktos/kubemark_tp_1_master/log$ grep "168993596" kube-apiserver.log 
I0622 20:35:04.364395       1 trace.go:81] Trace[168993596]: "listPartition etcd3: key=/registry/pods/" (started: 2021-06-22 20:34:47.210768417 +0000 UTC m=+13485.778327502) (total time: 17.153596435s):
Trace[168993596]: [8.704617084s] [8.704617084s] Acquiring ListAppendMux listPartition
Trace[168993596]: [17.153593818s] [8.448976144s] Took items from the response
```

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go
+++ b/staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go
@@ -730,6 +730,12 @@ func (s *store) List(ctx context.Context, key, resourceVersion string, pred stor
 		clients = s.getClientsFromKey(key)
 	}
 
+	// assert if the client is not 1
+	klog.V(2).Infof("length of clients: %v", len(clients))
+	if len(clients) != 1 {
+		panic("client is not 1")
+	}
+
 	listResults := make(map[uint8]*listPartitionResult, len(clients))
 	var wg sync.WaitGroup
 	wg.Add(len(clients))
@@ -765,9 +771,9 @@ func (s *store) List(ctx context.Context, key, resourceVersion string, pred stor
 			if err != nil {
 				result = &listPartitionResult{err: err}
 			}
-			listMux.Lock()
+		//	listMux.Lock()
 			listResults[i] = result
-			listMux.Unlock()
+		//	listMux.Unlock()
 			wg.Done()
 		}(c, keyToUse, returnedRV, continueKey, optionsToUse, listResults, i, &listAppendMux)
 	}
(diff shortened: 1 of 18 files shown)


staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go
```diff
@@ -867,14 +876,14 @@ func (s *store) listPartition(ctx context.Context, client *clientv3.Client, key
...
-		trace.Step("Acquired ListAppendMux listPartition")
+		// trace.Step("Acquiring ListAppendMux listPartition")
+		// listAppendMux.Lock()
+		// trace.Step("Acquired ListAppendMux listPartition")
 
 		if pred.Empty() {
 			growSlice(v, len(getResp.Kvs))
 		} else {
-			growSlice(v, 2048, len(getResp.Kvs))
+			growSlice(v, 4096, len(getResp.Kvs))
```
Sindica, line 886: Community is still using 2048 in 1.18.5. Not sure whether this is helpful. Suggest do not merge into POC branch.

yb01: this is experimental change trying to reduce the # of list reallocation.

Sindica (commented):
I am ok with the experiment. Question is what we plan to do if the performance is improved? Whether this is intended to be an ongoing changes in the poc branch. That's why I am not sure whether this PR shall be merged into 930 poc branch.

staging/src/k8s.io/apiserver/pkg/storage/etcd3/store.go
```diff
@@ -730,6 +730,12 @@ func (s *store) List(ctx context.Context, key, resourceVersion string, pred stor
 		clients = s.getClientsFromKey(key)
 	}
 
+	// assert if the client is not 1
+	klog.V(2).Infof("length of clients: %v", len(clients))
```
Sindica, line 734: Please comment out the log. It has a lot of occurrence in api server log.

yb01:
close the PR. local test didn't show issues around the lock and 15k run didn't hit the panic which means in arktos only one partition is used.

PR #1113, opened 2021-07-06
kubemark cluster starts partition servers in parallel

h-w-chen:
**What type of PR is this?**
/kind feature

This PR adopts parallel starting of arktos scale-out partitions when starting the kubemark cluster.

Current start-kunemark scripts starts the scale-out partitions (TP or RP) in sequence, which is quite time consuming if the number of TP servers and that of RP servers are not minimum. There are various ways to optimize the cluster setup time and reduce the overhead; one obvious way is making these partition servers started in parallel can significantly reduced the cluster setup time. This PR starts multiple TPs in parallel, after all TPs are started, then starts multiple RPs in parallel.

This PR does not try to start TPs and RPs at the same time.  

NONE


cluster/gce/util.sh
```diff
@@ -2502,12 +2511,12 @@ function kube-up() {
...
-  find-release-tars
-  registry-authentication
-  create-and-upload-etcd-image
-  create-and-upload-etcd-empty-dir-cleanup-image
-  upload-tars
+  if [[ -z ${UPLOAD_TAR_DONE:-} ]]; then
+    echo "DBG: uploading image + tar files..."
+    test_resource_upload
+    echo "DBG: (at kube-up) image + tar files uploaded"
+    UPLOAD_TAR_DONE="done in kube-up"
```
yb01, line 2518: will this flag, UPLOAD_TAR_DONE, visible to other process that calls kube-up() ?

test/kubemark/start-kubemark.sh
```diff
@@ -446,55 +450,92 @@ function setup_proxy {
...
+  echo "DBG: RP ${id} created"
+}
+
+detect-project &> /dev/null
 
 ### master_metadata is used in the cloud-int script to create the GCE VMs
 #
 MASTER_METADATA=""
 
+### to upload etcd image / binary tar once
```
yb01, line 477: the test_resource_upload() can be added to the very beginning of the test setup and ensure it is done. so that this check can be removed

test/kubemark/start-kubemark.sh
```diff
@@ -446,55 +450,92 @@ function setup_proxy {
...
-      MASTER_METADATA=${MASTER_METADATA},"tp-${tp_num}=${TP_KUBECONFIG}-${tp_num}"
-    fi
-
-    # export tp-1 token and share with other clusters
-    if [[ ${tp_num} == 1 ]]; then
-      tp1_token=$(grep token "${KUBEMARK_CLUSTER_KUBECONFIG}" | awk -F ": " '{print $2}')
-      export SHARED_APISERVER_TOKEN=${tp1_token}
-      echo "shares token: ${SHARED_APISERVER_TOKEN}"
-    fi
+    create_TP ${tp_num} &
```
yb01, line 532: logs for multiple TP creation will be mixed, should have a TODO to track and fix

test/kubemark/start-kubemark.sh
```diff
@@ -504,13 +545,19 @@ if [[ "${SCALEOUT_CLUSTER:-false}" == "true" ]]; then
 
   for (( rp_num=1; rp_num<=${SCALEOUT_RP_COUNT}; rp_num++ ))
   do
-    export KUBEMARK_CLUSTER_KUBECONFIG="${RP_KUBECONFIG}-${rp_num}"
-    export RESOURCE_PARTITION_SEQUENCE=${rp_num}
-    create-kubemark-master
+    create_RP ${rp_num} &
```
yb01, line 548: same here for the log mix-up in RP setup routines

yb01 (approved):
/lgtm
with a few minor comments

cluster/gce/util.sh
h-w-chen: yes each partition eventually calls into kube-up.sh; if it is set, partition starting won't have to repeat uploading.

test/kubemark/start-kubemark.sh
h-w-chen: admin cluster also needs this uploading, and it is directly set up by kube-up.sh script; starting partition clusters also eventually call into kube-up.sh; if we don't want to change this structure, we have to have some sort of condition check in place.

test/kubemark/start-kubemark.sh
h-w-chen: will add a TODO line as reminder

test/kubemark/start-kubemark.sh
h-w-chen: will add TODO line here too

PR #1125, opened 2021-07-24
Set user agent for clients when talking to api servers in the partition

Sindica:
fix scale out client identity

This fix fill in correct user agent for rest client when components need to talk to API server in another cluster.

### DIFF
--- a/cmd/kube-controller-manager/app/controllermanager.go
+++ b/cmd/kube-controller-manager/app/controllermanager.go
@@ -240,11 +240,19 @@ func Run(c *config.CompletedConfig, stopCh <-chan struct{}) error {
 		saTokenControllerInitFunc := serviceAccountTokenControllerStarter{rootClientBuilder: rootClientBuilder}.startServiceAccountTokenController
 
 		// start client to resource providers
-		if len(c.ResourceProviderClients) > 0 {
+		userAgent := getUserAgentFor2ndPartition(controllerContext)
+		if len(c.ResourceProviderKubeconfigs) > 0 {
 			controllerContext.ResourceProviderClients = make(map[string]clientset.Interface)
 			controllerContext.ResourceProviderNodeInformers = make(map[string]coreinformers.NodeInformer)
-			for i, rpClient := range c.ResourceProviderClients {
+			for i, rpKubeConfig := range c.ResourceProviderKubeconfigs {
 				rpId := "rp" + strconv.Itoa(i)
+
+				clientConfigs := restclient.CopyConfigs(rpKubeConfig)
+				for _, config := range clientConfigs.GetAllConfigs() {
+					config.UserAgent = userAgent
+				}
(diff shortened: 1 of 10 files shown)


yb01:
please rebase. the hollow-node and the clusterloader2 related userAgents were in 930Infra already.

cmd/kube-controller-manager/app/controllermanager.go
```diff
@@ -667,3 +676,11 @@ func shouldTurnOnDynamicClient(client clientset.Interface) bool {
 
 	return false
 }
+
+func getUserAgent(controllerContext ControllerContext) string {
+	isNodeControllerEnabled := controllerContext.IsControllerEnabled("nodelifecycle")
+	if isNodeControllerEnabled {
+		return "kcm-resource-provider"
+	}
+	return "kcm-tenant"
```
yb01, line 685: should be kcm-tenant-provider to align with the other agent name for RP

cmd/kube-controller-manager/app/controllermanager.go
```diff
@@ -240,11 +240,20 @@ func Run(c *config.CompletedConfig, stopCh <-chan struct{}) error {
 		saTokenControllerInitFunc := serviceAccountTokenControllerStarter{rootClientBuilder: rootClientBuilder}.startServiceAccountTokenController
 
 		// start client to resource providers
-		if len(c.ResourceProviderClients) > 0 {
+		userAgent := getUserAgent(controllerContext)
```
yb01, line 243: i can see a lot changes around the KCM config struct change is to set the RP and TP KCM with different agent name. probably it is not needed to differentiate them anyway if the user agent is set at the node and ipam controller ( i.e. node related ) already.

yb01 (approved):
/lgtm
with a few minor comments, clarifications.

Sindica:
> please rebase. the hollow-node and the clusterloader2 related userAgents were in 930Infra already.

This is test against master. To trigger CICD.

cmd/kube-controller-manager/app/controllermanager.go
Sindica: > i can see a lot changes around the KCM config struct change is to set the RP and TP KCM with different agent name. probably it is not needed to differentiate them anyway if the user agent is set at the node and ipam controller ( i.e. node related ) already.

The original way to create additional restclients does not allow user agent to be set.

Sindica:
Pod deletion has some issue. Scale up works. Scaleout kubelet has the following message:
I0724 19:18:51.370756   32346 kubelet_pods.go:951] Pod "nginx-deployment-a-7d8ffcd9fd-6zcbd_namespace-a_aaa(a354db7f-b646-4d41-85a8-5621f55ff17f):DeletionTimestamp=2021-07-24T19:10:59Z:ResourceVersion=853097220871290883" is terminated, but pod cgroup sandbox has not been cleaned up

yb01 (approved):
/lgtm

PR #1130, opened 2021-07-29
bug fix: unable to delete pods at hollow nodes

h-w-chen:
**What type of PR is this?**
/kind bug

It fixes the bug that pods cannot be deleted from hollow-node in perf tests


Sindica (approved):
/lgtm

PR #1135, opened 2021-08-09
add Immutable objects support from community

Sindica:
Try to reduce configmap watch from kubelet

### DIFF
--- a/pkg/kubelet/kubelet.go
+++ b/pkg/kubelet/kubelet.go
@@ -1237,6 +1237,9 @@ type Kubelet struct {
 
 	// Mutex to serialize new pod admission and existing pod resizing
 	podResizeMutex sync.Mutex
+
+	// lastestNode is the last node which kubelet updated or created
+	latestNode *v1.Node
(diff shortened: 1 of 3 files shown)


yb01:
is this the 1.18.5 change? i would suggest not to reduce this sort of load at this moment as we discussed sometime ago if 1.18.5 perf run doesn't have it.

Sindica:
> is this the 1.18.5 change? i would suggest not to reduce this sort of load at this moment as we discussed sometime ago if 1.18.5 perf run doesn't have it.

1.18.5 has the change but not enable the feature by default. I thought we were done only picking up 1.18.5 changes. Current focus is reducing API server load w/o too much new changes. But community changes should be fine.

yb01:
my concern is that this change could change the load pattern at the api server and distract the our focus in investigating the prelonge 10% slow pod creation cases. the watchers for secrets/config map is pretty consistent over the time during the test and it doesnt seem to be the issue we are chasing on currently. no?

PR #1149, opened 2021-08-11
Increase watch timeout period. Trying to reduce impact on frequently …

Sindica:
…re-establishing watch connections.

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note

```

### DIFF
--- a/staging/src/k8s.io/client-go/tools/cache/reflector.go
+++ b/staging/src/k8s.io/client-go/tools/cache/reflector.go
@@ -127,7 +127,7 @@ type Reflector struct {
 var (
 	// We try to spread the load on apiserver by setting timeouts for
 	// watch requests - it is random in [minWatchTimeout, 2*minWatchTimeout].
-	minWatchTimeout = 5 * time.Minute
+	minWatchTimeout = 30 * time.Minute
 )
 
 // NewNamespaceKeyedIndexerAndReflector creates an Indexer and a Reflector


staging/src/k8s.io/client-go/tools/cache/reflector.go
```diff
@@ -127,7 +127,7 @@ type Reflector struct {
 var (
 	// We try to spread the load on apiserver by setting timeouts for
 	// watch requests - it is random in [minWatchTimeout, 2*minWatchTimeout].
-	minWatchTimeout = 5 * time.Minute
+	minWatchTimeout = 30 * time.Minute
```
yb01, line 130: any reason using 30 minutes, i.e. not 10 minutes?
anyway, we should be careful when making such large min values, it could introduce side effort when reconnecting or concurrent socket sessions at the server.

Sindica: > any reason using 30 minutes, i.e. not 10 minutes?
> anyway, we should be careful when making such large min values, it could introduce side effort when reconnecting or concurrent socket sessions at the server.

Intentionally making big changes to see impact. What do you meant by concurrent socket sessions?

yb01: if this is experimental, it is OK for a temp change for testing. by "socket session" i refer to the low level connections to the server for the long-get call to the api server.

yb01 (approved):
/lgtm.
OK for some experiments.

staging/src/k8s.io/client-go/tools/cache/reflector.go
Sindica: > if this is experimental, it is OK for a temp change for testing. by "socket session" i refer to the low level connections to the server for the long-get call to the api server.

I understand that part. But what is the side effect?

PR #1150, opened 2021-08-11
Reduce kubelet getting node requests.

Sindica:
Same as PR 835 except test not ported.
Tested on 1TP/1RP 100 node cluster.
. Before change, get node request 17849, patch node request 1499. After change, get node request 303, patch node request 1100
. Depends on when log was pulled, total # of requests might variant. Using patch node request # as an example, get node still reduced a lot.
. There is no non 200 get/patch node requests in audit log.

25K run on 8/3 has 66M get and 4.5M patch. Total request in RP 142M.


yb01 (approved):
/lgtm.

let's keep it in the POC for a testing purpose on this change to see the benefit of reducing the GET calls.

PR #1151, opened 2021-08-13
porting changes from community issue 77733

yb01:
original changes are in PR https://github.com/kubernetes/kubernetes/pull/88962/files, which is porting changes for issue 77733 to 1.15 

+ quite a lot changes with the Arktos multiple tenant support.


pkg/controller/nodelifecycle/node_lifecycle_controller.go
```diff
@@ -166,13 +169,107 @@ type nodeHealthData struct {
...
+}
+
 // Controller is the controller that manages node's life cycle.
 type Controller struct {
 	taintManager            *scheduler.NoExecuteTaintManager
 	podInformersSynced      cache.InformerSynced
 	tenantPartitionManagers []*nodeutil.TenantPartitionManager
 	tenantPartitionClients  []clientset.Interface
 	resourcePartitionClient clientset.Interface
+	podLister         corelisters.PodLister
```
Sindica: Shouldn't this be an array of corelisters.PodLister or just use PodLister from tenantPartitionManagers?

pkg/controller/nodelifecycle/node_lifecycle_controller.go
```diff
@@ -204,6 +305,9 @@ type Controller struct {
 	nodeLister          corelisters.NodeLister
 	nodeInformerSynced  cache.InformerSynced
 
+	//TODO: remove this after UT fixes
+	getPodsAssignedToNode func(nodeName string) ([]*v1.Pod, error)
```
Sindica, line 309: Remove line 308-309?

Sindica (commented):
> original changes are in PR https://github.com/kubernetes/kubernetes/pull/88962/files, which is porting changes for issue 77733 to 1.15
> 
> * quite a lot changes with the Arktos multiple tenant support.

It is easier to just split issue 77733 to PR 79370 and 81167. Both PR merged to k8s master. Arktos already adopted 79370. This PR only needs to adopt 81167.

pkg/controller/nodelifecycle/node_lifecycle_controller.go
```diff
@@ -668,17 +789,30 @@ func (nc *Controller) doEvictionPass() {
 				klog.Warningf("Failed to get Node %v from the nodeLister: %v", value.Value, err)
 			}
 			nodeUID, _ := value.UID.(string)
+
+			// TODO: pods on nodes are for all TPs, need an efficient way to get subset of pods for the TP to the DeletePods function
```
Sindica, line 793: PodByNodeName is already indexed. Just a loop over all TPManagers should not affect performance much. We can definitely try to build index from Nodename to TPManager and avoid the loop. But I don't think it is worth just to save the extra loop.

pkg/controller/nodelifecycle/node_lifecycle_controller.go
```diff
@@ -668,17 +789,30 @@ func (nc *Controller) doEvictionPass() {
...
 			for _, tpManager := range nc.tenantPartitionManagers {
-				remaining, err := nodeutil.DeletePods(tpManager.Client, nc.recorder, value.Value, nodeUID, tpManager.DaemonSetStore)
+				pods, err := tpManager.PodByNodeNameLister(value.Value)
+
 				if err != nil {
+					utilruntime.HandleError(fmt.Errorf("unable to list pods from node %q: %v", value.Value, err))
```
Sindica, line 798: If tpManager is TP1, and node is from TP2, what will the output? Could it be some error?

pkg/controller/nodelifecycle/node_lifecycle_controller.go
```diff
@@ -728,6 +862,8 @@ func (nc *Controller) monitorNodeHealth() error {
 
 	zoneToNodeConditions := map[string][]*v1.NodeCondition{}
 	for i := range nodes {
+		continueToOuterLoop := false
```
Sindica, line 865: This is not what we generally do. You may use done or skip to break out.

pkg/controller/nodelifecycle/scheduler/taint_manager.go
```diff
@@ -78,6 +78,10 @@ type GetPodFunc func(name, namespace, tenant string) (*v1.Pod, error)
 // GetNodeFunc returns the node for the specified name, or a NotFound error if missing.
 type GetNodeFunc func(name string) (*v1.Node, error)
 
+//TODO: remove this after porting k8s 7777333 changes
```
Sindica, line 81: Shall we remove line 81-83?

pkg/controller/nodelifecycle/node_lifecycle_controller.go
Sindica: It looks like commented out later. Can this line be removed?

yb01: removed.

pkg/controller/nodelifecycle/node_lifecycle_controller.go
yb01: all UT has fixed to not reference to this function. so removed it as well.

pkg/controller/nodelifecycle/scheduler/taint_manager.go
yb01: removed.

pkg/controller/nodelifecycle/node_lifecycle_controller.go
yb01: this comment is a bit obsolete as it was added before the TP loop was added. i have updated it to likely for run the deletion in parallel.

pkg/controller/nodelifecycle/node_lifecycle_controller.go
yb01: the node is the same for all TPs. with multiple TPs, as long as one TP failed and the function retuns false to the TRY function which will replace the workitem in the queue and redo in the next iteration till ALL pods on all TPs were evicted for the nodes.

pkg/controller/nodelifecycle/node_lifecycle_controller.go
yb01: fixed for better readable code.

pkg/controller/nodelifecycle/node_lifecycle_controller.go
yb01: there is a potential case here where IF, TP1 keeps failing and it will affect the eviction pods from other TPs. but this can be fixed later on ( i guess use continue instead of break in the inner loop or run eviction in TP in parallel can solve this.

yb01: fixed

yb01:
> > original changes are in PR https://github.com/kubernetes/kubernetes/pull/88962/files, which is porting changes for issue 77733 to 1.15
> > 
> > * quite a lot changes with the Arktos multiple tenant support.
> 
> It is easier to just split issue 77733 to PR 79370 and 81167. Both PR merged to k8s master. Arktos already adopted 79370. This PR only needs to adopt 81167.

this is exactly this PR does. all the PRs picked up here is for 77733 and on top of PR 79370. all of them are for the final 81167 PR.

pkg/controller/nodelifecycle/node_lifecycle_controller.go
Sindica: Upon discussion, please make sure UT check whether error is possible if there is no pods existed for this TP and this node. Also can continue if len(pods) == 0.

yb01: fixed and added check for empty length of pods from one TP

Sindica (approved):
/lgtm

PR #1154, opened 2021-08-17
By randomize first update lease time, even out update lease requests.

Sindica:
Tested in local arktos-up


yb01 (approved):
/lgtm.

the same code can be used in the hollow-node main function to achieve the similar results and a clean way to do so,

Sindica:
At 25K test, this did not even out api server update lease requests.

PR #1163, opened 2021-08-24
Reduce perf test list all pods from API server cache

Sindica:
Trying avoid excessive list pods from API server during test. Reason: during density test, at the latency pod creation period, each latency pod is created by one deployment. Perf test create pod store for every single deployment and hence caused excessive list pods requests to API server. This may be the reason that perf test has pod start up latency spike every 10-15 seconds.

This change try to reuse pod store across multiple deployments. Pod store list/watch all latency pods, filter out pods that does not belong to current deployment at client side.

This is a preliminary change. Improvements might needed:
1. Scan over all pods at local store might added a lot of time. Need to compare with KCM pod start up latency to see whether additional local index for label selector might need to be added for pod.
2. Alternatively, pick up community change that use event received time as pod start up watch time.
3. Label selector comparison is very premitive. Depends on the usage, might need to support more cases.
4. Field selector comparison is not used in density test. Skipping for now.

5. TODO: Testing in 100 node, 1TP/1RP case.
    1. Run density test.
    2. Check audit log, compare list pods records with previous run.

### DIFF
--- a/perf-tests/clusterloader2/pkg/measurement/util/object_store.go
+++ b/perf-tests/clusterloader2/pkg/measurement/util/object_store.go
@@ -23,7 +23,10 @@ package util
 
 import (
 	"fmt"
+	"k8s.io/klog"
 	"reflect"
+	"strings"
+	"sync"
 	"time"
 
 	"k8s.io/api/core/v1"
@@ -47,7 +50,10 @@ type ObjectStore struct {
 func newObjectStore(obj runtime.Object, lw *cache.ListWatch, selector *ObjectSelector) (*ObjectStore, error) {
 	store := cache.NewStore(cache.MetaNamespaceKeyFunc)
 	stopCh := make(chan struct{})
-	name := fmt.Sprintf("%sStore: %s", reflect.TypeOf(obj).String(), selector.String())
+	name := fmt.Sprintf("%sStore", reflect.TypeOf(obj).String())
+	if selector != nil {
+		name = name + ": " + selector.String()
+	}
 	reflector := cache.NewNamedReflector(name, lw, obj, store, 0, false)
 	go reflector.Run(stopCh)
 	if err := wait.PollImmediate(50*time.Millisecond, 2*time.Minute, func() (bool, error) {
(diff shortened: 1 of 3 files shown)


Sindica:
100 node test run only has one list from latency pod
```
sindica2000@ying-dev1:~/logs/perf-test/gce-100/2021-08-24-reduce-perftest-list-pods/kubemark_tp_1_master/log$ grep "\"verb\":\"list\"" kube-apiserver-audit.log | grep pods | grep -v "\"userAgent\":\"kubemark" | grep latency{"kind":"Event","apiVersion":"audit.k8s.io/v1","level":"Request","auditID":"a090f307-da58-49dc-9e38-610392c21492","stage":"ResponseComplete","requestURI":"/api/v1/tenants/arktos/pods?labelSelector=group+%3D+latency\u0026limit=500\u0026resourceVersion=0","verb":"list","user":{"username":"admin","uid":"admin","groups":["system:masters","system:authenticated"]},"sourceIPs":["35.188.140.223"],"userAgent":"clusterloader/v0.0.0 (linux/amd64) kubernetes/$Format/cluster-loader","objectRef":{"resource":"pods","apiVersion":"v1"},"responseStatus":{"metadata":{},"code":200},"requestReceivedTimestamp":"2021-08-24T22:45:24.910270Z","stageTimestamp":"2021-08-24T22:45:24.912941Z","annotations":{"authorization.k8s.io/decision":"allow","authorization.k8s.io/reason":""}}
```

Pervious 100 node test only has 1001 list from latency pod
```
sindica2000@ying-dev1:~/logs/perf-test/gce-100/2021-08-11-reduce-kubelet-get/kubemark_tp_1_master/log$ grep "\"verb\":\"list\"" kube-apiserver-audit.log | grep pods | grep -v "\"userAgent\":\"kubemark" | grep latency | wc -l
1001
```

perf-tests/clusterloader2/pkg/measurement/util/object_store.go
```diff
@@ -74,27 +80,44 @@ func (s *ObjectStore) Stop() {
...
+var podStore *PodStore = nil
+var initPodStoreLock sync.Mutex
+
 // NewPodStore creates PodStore based on given object selector.
 func NewPodStore(c clientset.Interface, selector *ObjectSelector) (*PodStore, error) {
+	initPodStoreLock.Lock()
+	defer initPodStoreLock.Unlock()
+
+	if podStore != nil {
+		podStore.listener++
```
yb01, line 95: we can delete the podstore at the end of the test to avoid using this listener counter.

perf-tests/clusterloader2/pkg/measurement/util/object_store.go
```diff
@@ -74,27 +80,44 @@ func (s *ObjectStore) Stop() {
...
+
+func initPodStore(c clientset.Interface) (*PodStore, error) {
+	klog.Infof("0==== initPodStore")
+	//debug.PrintStack()
 	lw := &cache.ListWatch{
 		ListFunc: func(options metav1.ListOptions) (runtime.Object, error) {
-			options.LabelSelector = selector.LabelSelector
-			options.FieldSelector = selector.FieldSelector
-			return c.CoreV1().PodsWithMultiTenancy(selector.Namespace, util.GetTenant()).List(options)
+			return c.CoreV1().PodsWithMultiTenancy(metav1.NamespaceAll, util.GetTenant()).List(options)
```
yb01, line 110: do we still need the options for the list and watch function? or just use group=latency or group=saturation ?

yb01 (approved):
/lgtm
we can get in and refine the logic later ( mainly the local cache filters )

perf-tests/clusterloader2/pkg/measurement/util/object_store.go
Sindica: Currently we still have group=latency and group=saturation in the list/watch argument. It might come from options. Leave the option there until it is verified.

PR #1169, opened 2021-09-07
Add label name indexer to pod

Sindica:
Tested in 100 node. Test success.

```
I0907 02:35:43.592065    3452 phase_latency.go:127] PodStartupLatency: perc50: 2.73403405s, perc90: 4.304019186s, perc99: 6.818838053s; threshold 5m30s
I0907 02:39:22.978359    3452 phase_latency.go:127] PodStartupLatency: perc50: 1.751818304s, perc90: 2.489899127s, perc99: 2.871392377s; threshold 5s
```

Log: ying-dev1:/home/sindica2000/logs/perf-test/gce-100/2021-09-06-index-pod-in-perftest

. Added index for group, succeed in 500 nodes. Log ying-dev1:/home/sindica2000/logs/perf-test/gce-500/arktos/2021-09-07-index-pod-in-perftest
```
0907 19:31:25.278504     611 phase_latency.go:127] PodStartupLatency: perc50: 1.918785012s, perc90: 2.697729696s, perc99: 3.532998819s; threshold 5m30s
I0907 19:35:06.301422     611 phase_latency.go:127] PodStartupLatency: perc50: 1.847863708s, perc90: 2.48094182s, perc99: 2.848345077s; threshold 5s
```


perf-tests/clusterloader2/pkg/measurement/util/object_store.go
```diff
@@ -80,9 +85,17 @@ func (s *ObjectStore) Stop() {
 // PodStore is a convenient wrapper around cache.Store.
 type PodStore struct {
 	*ObjectStore
-	listener int
+	listener      int
+	podListerFunc func(labelKey string, labelValue string) ([]*v1.Pod, error)
 }
 
+const (
+	labelNameKeyIndex = "label.name"
```
yb01, line 93: do we need both indexers ? just group should be good for the latency pods which we only care about, right ?

Sindica: In local test, somehow there is case that FilterPods get selector with group=saturation only. It falls into checking each label case, which could still take a lot of time.

perf-tests/clusterloader2/pkg/measurement/util/object_store.go
```diff
@@ -146,13 +209,36 @@ func (s *PodStore) Stop() {
...
+		if name, isOK := selectorMap[labelNameKey]; isOK {
+			labelKeyName = labelNameKey
+			labelKeyValue = name
+		} else if group, isOK := selectorMap[labelGroupKey]; isOK {
+			labelKeyName = labelGroupKey
+			labelKeyValue = group
+		}
+		if labelKeyName != "" && labelKeyValue != "" {
+			pods, err := ps.podListerFunc(labelKeyName, labelKeyValue)
+			klog.Infof("==== FilterPods key [%v], value [%v], err [%v], len(pods)=[%v]", labelKeyName, labelKeyValue, err, len(pods))
```
yb01, line 232: ho many log entries this generates ?

perf-tests/clusterloader2/pkg/measurement/util/object_store.go
Sindica: It is used for the periodically print out "SchedulingThroughput: # pods scheduled"

yb01: if we can only have one index, the code change can be much simplified.

yb01 (approved):
/lgtm

perf-tests/clusterloader2/pkg/measurement/util/object_store.go
Sindica: about 2000 in 500 node test. Will delete before check in.

perf-tests/clusterloader2/pkg/measurement/util/object_store.go
Sindica: > if we can only have one index, the code change can be much simplified.

The code change is mainly for efficiency. W/o the new group index, the label comparision will need to be done for every single pod in saturation pod creation step. It can also be very CPU intensive.

perf-tests/clusterloader2/pkg/measurement/util/object_store.go
Sindica: Commented.

PR #1172, opened 2021-09-08
scale-out: kube-proxy connects to multiple TPs and arranges service/endpoint in tenant aware way 

### LINKED ISSUE #1141: [service support] kube-proxy: make it aware of multi-tenants
* kube-proxy needs to bear tenant/network in mind when manage mapping of service/EP;
* it needs to access multiple TPs;

**Why is this needed**:
upstream kube-proxy has not tenant concept at all; proper code change is needed to ensure service/EP(pod IP) mapping within the tenant and network scope.

**How to verify**:
1. to start up the cluster
* scale-up env
   simply run hack/arktos-up.sh
* scale-out 1x1 env
   kube-proxy at node of RP shall be able to connect to TP via simple kubeconfig file, tp1.kubeconfig
```yaml
apiVersion: v1
kind: Config
clusters:
- cluster:
    server: http://<tp-ip>:8080/
  name: tp
contexts:
- context:
    cluster: tp
    user: local-up-cluster
  name: dev-scale-out
current-context: dev-scale-out
users:
- name: local-up-cluster
  user: {}
```
also, the config file, kube-proxy.config, shall have content like
```yaml
apiVersion: kubeproxy.config.k8s.io/v1alpha1
kind: KubeProxyConfiguration
clientConnection:
  kubeconfig: <kubeconfig for RP>
tenantPartitionKubeConfig: /path-to/tp1.kubeconfig
hostnameOverride: z1
mode: 
featureGates:
  AllAlpha: false
  WorkloadInfoDefaulting: true
  QPSDoubleGCController: true
  QPSDoubleRSController: true
```
kube-proxy can be start with proper args
```
./_output/local/bin/linux/amd64/kube-proxy --v=4 --config=/root/kube-proxy.config
```
* scale-out 2x2 env
on top of what are for 1x1, the config file should include tp2.kubeconfig besides that of tp1.

2. on each tp server, run
```
kubectl create tenant <tp-specific-tenant>
kubectl run my-nginx --image=nginx --replicas=2 --tenant <tp-specific-tenant>
kubectl expose deploy my-nginx --port 8000 --target-port 80 --tenant <tp-specific-tenant>
```
after pods are running, swiicth to the node and run
```
curl <svc-ip-of-my-nginx>:8000
```
Inspecting the iptables chain and rules, the same svc from different tenant shall all exist without conflicts.

h-w-chen:
**What type of PR is this?**
/kind feature

This PR adapts kube-proxy in scale-out cluster, which is of multiple tenants + multiple TPs.

It composes 2 parts:
1. source code changes to kube-proxy making it work with multi-tenants (in one tenant server) and multi-tenant-partitions;
2. script code change to generate kubeconfig files and feed proper arg for kube-proxy to start in dev env of scale-out cluster;

Fixes #1141

no unit test change - existing unit tests are still using ns+name only w/o tenant, not broken by the additional tenant data.

yes
```release-note
kube-proxy accepts arg, --tenant-server-kubeconfig, specifying the connection config files to all TPs in scale-out cluster;
these conf files can also be specified in its app configuration file,  using the newly introduced field tenantPartitionKubeConfig.
```


cmd/kube-proxy/app/server.go
```diff
@@ -145,6 +145,7 @@ func (o *Options) AddFlags(fs *pflag.FlagSet) {
 	fs.StringVar(&o.ConfigFile, "config", o.ConfigFile, "The path to the configuration file.")
 	fs.StringVar(&o.WriteConfigTo, "write-config-to", o.WriteConfigTo, "If set, write the default configuration values to this file and exit.")
 	fs.StringVar(&o.config.ClientConnection.Kubeconfig, "kubeconfig", o.config.ClientConnection.Kubeconfig, "Path to kubeconfig files with authorization information (the master location is set by the master flag).")
+	fs.StringVar(&o.config.TenantPartitionKubeConfig, "tenant-server-kubeconfig", o.config.TenantPartitionKubeConfig, "Comma separated string representing kubeconfig files point to tenant api servers.")
```
yb01, line 148: with the new TenantPartitionKubeconfig parameter, the  ClientCOnnection.Kubeconfig parameter can be removed.

cmd/kube-proxy/app/server_others.go
```diff
@@ -110,6 +114,26 @@ func newProxyServer(
...
+		clients = []clientset.Interface{client}
+	} else {
+		klog.V(3).Infof("scale-out cluster; make KubeClients based on TenantPartitionKubeConfig arg: %v", config.TenantPartitionKubeConfig)
+		kubeConfigFiles, existed := genutils.ParseKubeConfigFiles(config.TenantPartitionKubeConfig)
+		if !existed {
+			klog.Fatalf("Kubeconfig file(s) [%s] for tenant server does not exist", config.TenantPartitionKubeConfig)
+		}
+		clients = make([]clientset.Interface, len(kubeConfigFiles))
+		for i, kubeConfigFile := range kubeConfigFiles {
+			clients[i], err = clientutil.CreateClientFromKubeconfigFile(kubeConfigFile, "kubelet")
```
yb01, line 130: this should be kube-proxy instead of kubelet

cmd/kube-proxy/app/server_others.go
```diff
@@ -110,6 +114,26 @@ func newProxyServer(
...
+			if err != nil {
+				klog.Fatalf("failed to initialize kubelet client from kubeconfig [%s]: %v", kubeConfigFile, err)
```
yb01, line 132: here too. should be kube-proxy instead of kubelet

cmd/kube-proxy/app/server_others.go
```diff
@@ -110,6 +114,26 @@ func newProxyServer(
 		return nil, err
 	}
 
+	// create client for each tenant partition, if applicable
+	var clients []clientset.Interface
+	if len(config.TenantPartitionKubeConfig) == 0 {
```
yb01, line 119: for your consideration: it probably will be clearer to move this loop to the caller, so to keep the proxyServerer type unchanged.

hack/lib/common.sh
```diff
@@ -781,11 +781,30 @@ function kube::common::start_kubeproxy {
...
+      serverCount=${#TENANT_SERVERS[@]}
+      kubeconfig_filename="tenant-server-kube-proxy"
+      for (( pos=0; pos<${serverCount}; pos++ ));
+      do
+        # here generate kubeconfig for remote API server. Only work in non secure mode for now
+        kube::util::write_client_kubeconfig "${CONTROLPLANE_SUDO}" "${CERT_DIR}" "" "${TENANT_SERVERS[${pos}]}" "${API_PORT}" tenant-server-kube-proxy "" "http"
+        ${CONTROLPLANE_SUDO} mv "${CERT_DIR}/${kubeconfig_filename}.kubeconfig" "${CERT_DIR}/${kubeconfig_filename}${pos}.kubeconfig"
+        ${CONTROLPLANE_SUDO} chown "$(whoami)" "${CERT_DIR}/${kubeconfig_filename}${pos}.kubeconfig"
+
+        TP_KUBEONFIGS="${TP_KUBEONFIGS}${CERT_DIR}/${kubeconfig_filename}${pos}.kubeconfig,"
```
yb01, line 797: nit. trailing ','

pkg/proxy/endpoints.go
```diff
@@ -130,7 +130,7 @@ func (ect *EndpointChangeTracker) Update(previous, current *v1.Endpoints) bool {
 		return false
 	}
 	metrics.EndpointChangesTotal.Inc()
-	namespacedName := types.NamespacedName{Namespace: endpoints.Namespace, Name: endpoints.Name}
+	namespacedName := types.NamespacedName{Tenant: endpoints.Tenant, Namespace: endpoints.Namespace, Name: endpoints.Name}
```
yb01, line 133: nit: variable name also needs to be updated.

yb01 (approved):
/lgtm. 
a few minor comments and suggestions to consider.

cmd/kube-proxy/app/server.go
h-w-chen: rp connection seems needed as events (from kube-proxy) would still go there. Besides, in scale-up env, kubeconfig is better than other form as it is sort of k8s convention.

cmd/kube-proxy/app/server_others.go
h-w-chen: thanks for pinting it out. yes it should be kube-proxy. will change.

cmd/kube-proxy/app/server_others.go
h-w-chen: will modify to kube-proxy

hack/lib/common.sh
h-w-chen: in line 799 TP_KUBEONFIGS=${TP_KUBEONFIGS::-1} it trims off the last ,.  as expected.

PR #1173, opened 2021-09-08
porting changes from community issue 77733 (#1151)

yb01:
* Porting code from k8s https://github.com/kubernetes/kubernetes/pull/88962/files

removing redundant code

moving podInformer to node controller scope

adding lock to node data map

eviction processing refactor

adding fakeGetPodsAssignedToNode

adding pods to DeletePods parameters

adding pods to MarkPodsNotReady parameters

using pod pointers in node lifecycle controller

MarkPodsNotReady retry fix

adding test for retrying MarkPodsNotReady

adding pods lister

* fix for multiple TP logic in new ported code

* minor fix and UT fix

* more node controller UT fixes

* more node controller UT fixes

* fix ut failures and minor updates

* optimize the case where the pods is empty on node for a TP

Co-authored-by: Krzysztof Siedlecki <ksiedlecki@google.com>

**What type of PR is this?**
>
> /kind api-change
> /kind bug
> /kind cleanup
> /kind design
> /kind documentation
> /kind failing-test
> /kind feature
> /kind flake

Fixes #

```release-note

```


pkg/controller/nodelifecycle/node_lifecycle_controller.go
```diff
@@ -753,97 +890,156 @@ func (nc *Controller) monitorNodeHealth() error {
...
-								node.Name,
-								decisionTimestamp,
-								nc.nodeHealthMap[node.Name].readyTransitionTimestamp,
-								nc.podEvictionTimeout,
-							)
-						}
-					}
+			for _, tpManager := range nc.tenantPartitionManagers {
+				pods, err := tpManager.PodByNodeNameLister(node.Name)
+				if len(pods) == 0 && err == nil {
```
Sindica, line 896: Is it possible that there is no pod deployed on the node, but node has problem and need to be evicted?

yb01:
cherry-picked an dependent PR from scale-930 branch as well. removed do-not-merge flag.

pkg/controller/nodelifecycle/node_lifecycle_controller.go
yb01: it is possible. e.g. all pods on the node has just been finished and before new pods were scheduled to it, the runtime or network failed on the node.

note that the eviction is for PODs, not the node itself.

pkg/controller/nodelifecycle/node_lifecycle_controller.go
```diff
@@ -753,97 +890,156 @@ func (nc *Controller) monitorNodeHealth() error {
...
+				if err != nil {
+					utilruntime.HandleError(fmt.Errorf("unable to list pods of node %v: %v", node.Name, err))
+					if currentReadyCondition.Status != v1.ConditionTrue && observedReadyCondition.Status == v1.ConditionTrue {
+						// If error happened during node status transition (Ready -> NotReady)
+						// we need to mark node for retry to force MarkPodsNotReady execution
+						// in the next iteration.
+						nc.nodesToRetry.Store(node.Name, struct{}{})
+						// mark the node for retry and continue for other TPs
+						donePodEvictionOnNode = false
+						continue
```
Sindica, line 909: Is "continue" inside or outside of is clause?

pkg/controller/util/node/controller_utils.go
```diff
@@ -123,50 +122,35 @@ func SetPodTerminationReason(kubeClient clientset.Interface, pod *v1.Pod, nodeNa
...
+		pod := pods[i].DeepCopy()
 		for _, cond := range pod.Status.Conditions {
 			if cond.Type == v1.PodReady {
 				cond.Status = v1.ConditionFalse
 				if !utilpod.UpdatePodCondition(&pod.Status, &cond) {
 					break
 				}
 				klog.V(2).Infof("Updating ready status of pod %v to false", pod.Name)
-				_, err := kubeClient.CoreV1().Pods(pod.Namespace).UpdateStatus(&pod)
+				_, err := kubeClient.CoreV1().Pods(pod.Namespace).UpdateStatus(pod)
```
Sindica, line 146: Shouldn't this use multitenancy API?

pkg/controller/nodelifecycle/node_lifecycle_controller.go
yb01: inside. the goal for this two-loops are ensure retryNode only deleted for pods on both TPs were evicted, and the flow runs for each nodes.

pkg/controller/util/node/controller_utils.go
yb01: only GET operation has the tenancy-based APIs. write OPs ( update,delete ) has the pod itself as parameter so it will be unique anyways. the change here is the pod pointer, function wise, it is not changed.

pkg/controller/nodelifecycle/node_lifecycle_controller.go
Sindica: Community change has continue outside of if clause.

yb01: logically they are no difference. but i will move out to align with the community changes.

yb01: fixed.

pkg/controller/nodelifecycle/node_lifecycle_controller.go
```diff
@@ -753,97 +890,156 @@ func (nc *Controller) monitorNodeHealth() error {
...
-						if nc.evictPods(node) {
-							klog.V(2).Infof("Node is NotReady. Adding Pods on Node %s to eviction queue: %v is later than %v + %v",
-								node.Name,
-								decisionTimestamp,
-								nc.nodeHealthMap[node.Name].readyTransitionTimestamp,
-								nc.podEvictionTimeout,
-							)
-						}
-					}
+			for _, tpManager := range nc.tenantPartitionManagers {
```
Sindica, line 894: Due to adding an extra loop inside, the continue has different meaning from community changes. Perhaps reconsider code logic.

yb01: refactored a bit to hide the logic for multiple tps in functions.

Sindica (approved):
/lgtm

PR #1174, opened 2021-09-08
adds deployment network admission controller to enforce network label in deployment spec

### LINKED ISSUE #1146: [service support] deployment/RS/pod: ensure network label is assigned properly
1. create a deployment in a certain network;
2. verify RS object created has same network property;
3. verify pods created have same network property

**Why is this needed**:
to ensure deploy/RS controller respect network property

This is UX related - when users define deployment, the network association (by means of network label) can be allowed to specify at deployment metadata, no need for repetitive definition of pod template. 

It ensures pods of  deployment will inherit its network as well. Without this enforcement, deployment could have pods of different network.

h-w-chen:
**What type of PR is this?**
/kind feature

This PR introduces the deployment network admission controller (default on), which validates deployment spec ensuring its network association is consistent with that of the pod template.

This introduces extra validation for arktos network, which enforces deployment has to specify consistent network label at 2 places - deployment metadata, and pot template metadata. 

Fixes #1146

yes
```release-note
deployment and its pod template have to have consistent network label; otherwise, deployment resource cannot be created. 
```


yb01:
i try to understand if we really need this. it seems it is needed here because the network object is created at the deployment metadata, not as a pod template. any reason we do so ?

plugin/pkg/admission/deployment/network/admission.go
```diff
@@ -0,0 +1,62 @@
...
+		return nil
+	}
+
+	d, ok := attributes.GetObject().(*apps.Deployment)
+	if !ok {
+		return apierrors.NewBadRequest("Resource was marked with kind Deployment but was unable to be converted")
+	}
+
+	// deployment and pod template shall have consistent network labels
+	networkOfDeploy := d.ObjectMeta.Labels[arktosv1.NetworkLabel]
```
yb01, line 40: this will require the network set on both deployment and the pod template. if the deployment created from portal UI, the UI logic can handle it, but if use directly create deployment via the API. user has to ensure setup on both.

yb01 (approved):
/lgtm.

PR #1175, opened 2021-09-09
Temporary increase pod cache size for density test in 25K cluster.

Sindica:
Note this change cannot be combined with api server target-ram-mb increment.

./cluster/gce/gci/configure-helper-common.sh:      params+=" --target-ram-mb=600000"


sonyafenge (approved):
/LGTM

yb01 (approved):
/lgtm

Issue #1140, opened 2021-08-10
[service support] kubelet: provision the pod sandbox with its network specific DNS ip


h-w-chen:
**What would you like to be added**:
1. add network resource manager to optimistically get pod's network resource (like that of config map and secret);
2. set sandbox's /etc/resolv content with proper kube-dns_{network} service IP

**Why is this needed**:
DNS service is network specific in multi-tenant Arktos

https://github.com/CentaurusInfra/arktos/pull/1176 was created on 2021-09-09.

q131172019:
Start one-node scale-up cluster and see whether the command is working? - Still not working

“./cluster/kubectl.sh exec -ti pod/<POD_NAME> -- ls /etc/resolv.conf”

```
ubuntu@ip-172-31-3-192:~/go/src/arktos$ ./cluster/kubectl.sh exec -ti pod/coredns-default-755c76cc67-d5tg9 -n kube-system --tenant mytenant -- ls /etc/resolv.conf
error: Internal error occurred: error executing command in container: failed to exec in container: failed to start exec "b5127dd94bc912a2db27bcbbb9f741124cc706d11a7d22e555b5f23382466461": OCI runtime exec failed: exec failed: container_linux.go:380: starting container process caused: exec: "ls": executable file not found in $PATH: unknown
ubuntu@ip-172-31-3-192:~/go/src/arktos$ ./cluster/kubectl.sh exec -ti pod/coredns-default-755c76cc67-d5tg9 -n kube-system --tenant mytenant -- /bin/ls /etc/resolv.conf
error: Internal error occurred: error executing command in container: failed to exec in container: failed to start exec "56210fa94ee3ac931a04e4adbc68967c35afd919358e62251535723647028409": OCI runtime exec failed: exec failed: container_linux.go:380: starting container process caused: exec: "/bin/ls": stat /bin/ls: no such file or directory: unknown
ubuntu@ip-172-31-3-192:~/go/src/arktos$ ./cluster/kubectl.sh exec -ti pod/coredns-default-755c76cc67-d5tg9 -n kube-system --tenant mytenant -- /bin/sh
error: Internal error occurred: error executing command in container: failed to exec in container: failed to start exec "dbdd509e431d4c0bb00880367f47cc2498dd24c91ddd28fa1f6a3ac7ebc01766": OCI runtime exec failed: exec failed: container_linux.go:380: starting container process caused: exec: "/bin/sh": stat /bin/sh: no such file or directory: unknown
[...]

PR #1176, opened 2021-09-10
The changes are to fix the issue #1140

q131172019:
**What type of PR is this?**

/kind bug
/kind feature

Use serviceLister method to get proper kube-dns_{network} service IP for sandbox
set sandbox's /etc/resolv content with proper kube-dns_{network} service IP

Fixes # 1140

To reduce complexity, use serviceLister method to get proper kube-dns_{network} service IP for sandbox instead of original plan - use network resource manager to optimistically get pod's network resource.

NONE.


hack/arktos-worker-up.sh
```diff
@@ -68,6 +68,7 @@ sudo ./_output/local/bin/linux/amd64/hyperkube kubelet \
 --container-runtime-endpoint="containerRuntime,container,/run/containerd/containerd.sock;vmRuntime,vm,/run/virtlet.sock" \
 --runtime-request-timeout=2m \
 --port=10250 \
+--resolv-conf=/run/systemd/resolve/resolv.conf \
```
h-w-chen, line 71: since RESOLV_CONF is already the env var for resolv file path, can we use that instead of hard coding at this place?

pkg/features/kube_features.go
```diff
@@ -622,7 +622,7 @@ var defaultKubernetesFeatureGates = map[featuregate.Feature]featuregate.FeatureS
 	PodOverhead:                                    {Default: true, PreRelease: featuregate.Beta},
 	InPlacePodVerticalScaling:                      {Default: false, PreRelease: featuregate.Alpha},
 	PerNetworkServiceIPAlloc:                       {Default: false, PreRelease: featuregate.Alpha},
-	MandatoryArktosNetwork:                         {Default: false, PreRelease: featuregate.Alpha},
+	MandatoryArktosNetwork:                         {Default: true, PreRelease: featuregate.Beta},
```
h-w-chen, line 625: will this change from false to true break kubemark testing?

pkg/kubelet/network/dns/dns.go
```diff
@@ -98,17 +104,35 @@ func (c *Configurer) getClusterDNS(pod *v1.Pod) ([]net.IP, error) {
...
-		ip := net.ParseIP(network.Status.DNSServiceIP)
-		if ip != nil {
-			return []net.IP{ip}, nil
+		// Get the cluster IP of service kube-dns-{network} as name server
+		// inside the file /etc/resolv.conf of pod
+		if service.Tenant == pod.Tenant && service.Name == kubeDNSServiceName {
+			ip := net.ParseIP(service.Spec.ClusterIP)
+			if ip != nil {
+				klog.V(4).Infof("Pod NAME: %q | ClusterDNS IP : %q | networkName : %q | Tenant : %q ", pod.Name, ip,  networkName, pod.Tenant)
+				return []net.IP{ip}, nil
```
h-w-chen, line 133: it seems that on success, it always returns one result. If that is the case, is there any other reason for the interface method to return collection instead of singular one?

pkg/kubelet/network/dns/dns.go
```diff
@@ -98,17 +104,35 @@ func (c *Configurer) getClusterDNS(pod *v1.Pod) ([]net.IP, error) {
...
+	services, err := c.serviceLister.List(labels.Everything())
+	if err != nil {
+		return nil, fmt.Errorf("failed to list services when getting service IP for network")
+	}
+
+	const kubeDNSPrefix = "kube-dns-"
+	var kubeDNSServiceName = kubeDNSPrefix + networkName
+
+	// Get all services and find out the cluster IP of service kube-dns-{network}
+	for i := range services {
```
h-w-chen, line 119: consider speed up the search, if it is possible.

pkg/kubelet/network/dns/dns.go
```diff
@@ -397,6 +421,9 @@ func (c *Configurer) GetPodDNS(pod *v1.Pod) (*runtimeapi.DNSConfig, error) {
 			dnsConfig.Servers = []string{}
 			for _, ip := range clusterDNS {
 				dnsConfig.Servers = append(dnsConfig.Servers, ip.String())
+				nodeInfoMsg := fmt.Sprintf("For 1 TP X 1 RP verification - ClusterDNS IP : %q ", ip.String())
```
h-w-chen, line 424: please rephrase this message to get rid of 1x1 specific info

pkg/kubelet/network/dns/dns.go
```diff
@@ -397,6 +421,9 @@ func (c *Configurer) GetPodDNS(pod *v1.Pod) (*runtimeapi.DNSConfig, error) {
...
+				c.recorder.Eventf(c.nodeRef, v1.EventTypeWarning, "GetingClusterDNS", nodeInfoMsg)
+				c.recorder.Eventf(pod, v1.EventTypeWarning, "GettingClusterDNS", "pod: %q. %s", format.Pod(pod), nodeInfoMsg)
```
h-w-chen, line 426: in scale-out env, single event recorder probably has difficulty to send events of node and pod to same place.

h-w-chen (changes requested):
this PR is about getting network specifc DNS service ip address for the pod, by leveraging the service lister which is already in kubelet code.
Generally it looks good; the major concern is ensuring kubemark is not broken.

hack/arktos-worker-up.sh
q131172019: The script hack/arktos-worker-up.sh does not use the two environment setting files - hack/lib/common-var-init.sh including  RESOLV_CONF and hack/lib/common.sh which are used by the script hack/arktos-up.sh. So --resolv-conf=/run/systemd/resolve/resolv.conf is added here.

pkg/features/kube_features.go
q131172019: Yes, will change back to false. Then instead to  start kubelet with --feature-gates=true

pkg/kubelet/network/dns/dns.go
q131172019: To do optimization: 
For 1 TP: consider use {TENANT}+{NAMESPACE}+{SERVICE NAME} as key to create serviceMAP, then use serviceMAP to find out corresponding service IP of kube-dns-{network}

For multi-TP:  have to consider how to identify serviceLister from different TPs.

pkg/kubelet/network/dns/dns.go
q131172019: The community code at https://github.com/CentaurusInfra/arktos/blob/8e5c308a2a1da21e18a8949db7224d16fba01f08/pkg/kubelet/network/dns/dns.go#L386-L387 call the method 'getClusterDNS' and return collection of IP other than singular one. Just follow-up community method and do not change too many codes.

		var clusterDNS []net.IP
		if clusterDNS, err = c.getClusterDNS(pod); err != nil {

pkg/kubelet/network/dns/dns.go
q131172019: Yes. Will remove "1 TP X 1 RP" specific info.

pkg/kubelet/network/dns/dns.go
q131172019: Yes, good to know.

pkg/features/kube_features.go
q131172019: The correct way is use environment setting to enable feature MandatoryArktosNetwork to be true in shell command   - export FEATURE_GATES="AllAlpha=false,MandatoryArktosNetwork=true"

q131172019:
The PR1180 covers this PR.

Issue #1168, opened 2021-08-31
scale-out analysis: core-DNS uses customized env var KUBERNETS_SERVICE_HOST, KUBERNETES_SERVICE_PORT to access relevant TP


h-w-chen:
**What would you like to be added**:
analysis task - for mxn (e.g. 2x2) scale-out cluster
1. to investigate using customized env vars to ensure core-DNS pods able to access their relevant TP servers;
2. to evaluate potential impacts of such customization upon existing Arktos components, particularly network controller;
3. to estimate the potential affect to scale-out env (e.g. deployment definitions);

**Why is this needed**:
KUBERNETS_SERVICE_HOST & KUBERNETES_SERVICE_PORT are two env vars kubelet injects into pod sandbox on pod provisioning. env var value could be overridden by as of pod spec.

h-w-chen:
1 - yes it is feasible - these env vars can be specified when the core-DNS deployment is created by network controller which accesses its single TP server. What we need is ensuring the pod template def of the deployment has proper values of the TP (e.g. kubernetes svc IP). As the network controller is responsible for creating core-DNS deployments, it is able to provide values in the template.

h-w-chen:
2. impacts upon other components in scale-out system:
* arktos network controller: when it starts, it needs to locate the proper values of the aforementioned env vars - a reasonable choice is to get the svc ip/port of kubernetes svc, after the TP has started and allocated such resource - this requires script code addition in arktos-up-scale-out-poc.sh;
* if network controller runs as pod, it needs to have its own KUBERNETES_SERVICE_HOST and PORT vars set properly, besides the that needs to change accordingly for proper core-DNS creation;
* TP API server: needs to have unique cidr range (--service-cluster-ip-range) of cluster IP addrs, so that  kubernetes svc IP of different TP shall be different;

h-w-chen:
3. for scale-up system, as there one API server, the values is trivial to identify and network controller picks it up.

h-w-chen:
DNS pod template of the deployment uses API server ip addr and port (6443) for such two env vars. closed

PR #1177, opened 2021-09-10
arktos-network-controller introduces arg to make names of resources it creates not fully hardcoded

h-w-chen:
**What type of PR is this?**
/kind feature

This PR is to address one point of issue #1168  
* arktos network controller may need to have extra setting to make the resources (like sa,secret) of system tenant (from multiple TPs) unique across whole cluster

This change is necessary for scale-out env, as some resources (like sa and configmap) used by DNS pods may clash if they are from different TPs, which would make DNS pod fail. 

part of issue 1168

unit test code also has minimum change

yes
```release-note
in scale-out env, core DNS names and coreDNS sa/configmap names could have TP-specific suffixes. 
```


cmd/arktos-network-controller/app/dns_deployment.go
```diff
@@ -31,39 +31,47 @@ import (
...
 	return nil
 }
 
-func ensureToCreateServiceAccount(net *v1.Network, client kubernetes.Interface) error {
-	if _, err := client.CoreV1().ServiceAccountsWithMultiTenancy(metav1.NamespaceSystem, net.Name).Get(dnsBaseName, metav1.GetOptions{}); err != nil {
+func getSuffixedName(base, suffix string) string {
+	if len(suffix) == 0 {
+		return base
+	}
+	return base + "-" + suffix
```
yb01, line 62: is there any size or char restriction for the string variables ?

yb01 (approved):
/lgtm

Issue #1180, opened 2021-09-14
[service support] kubelet: serviceLister only retrieves data from KubeTPClients[0]


q131172019:
**What happened**:
https://github.com/CentaurusInfra/arktos/blob/8e5c308a2a1da21e18a8949db7224d16fba01f08/pkg/kubelet/kubelet.go#L444
- serviceLW (ListWatch) only lists and watches the service changes from KubeTPClients[0] (one TP). In M TPs X N RPs env, this is a bug.

**What you expected to happen**:
serviceLW (ListWatch) can lists and watches the service changes from KubeTPClients[i] (i = 0,1 ... M-1) in M TPs X N RPs env

q131172019:
PR 1181 was sent out last Wednesday and approved last Thursday.

h-w-chen:
PR #1181 merged in. This bug is fixed.

PR #1181, opened 2021-09-14
This PR is to fix both issue #1180 and issue #1140 after consolidating with PR1176

q131172019:
**What type of PR is this?**
 /kind bug
 /kind feature

- Use serviceLister slice method to get proper kube-dns_{network} service IP for sandbox
set sandbox's /etc/resolv content with proper kube-dns_{network} service IP

- Use serviceLister slice mapping multiply KubeTPClients to fix the bug - ServiceLW (ListWatch) currently only lists service from KubeTPClients[0] in multi-tp scale-out environment, 

Fixes # 1140 and 1180

In pkg/kubelet/kubelet_pods.go, the known bugs line 507 and line 511 have been fixed for  multiply KubeTPClients, by the way.

N/A


pkg/kubelet/kubelet.go
```diff
@@ -439,12 +439,20 @@ func NewMainKubelet(kubeCfg *kubeletconfiginternal.KubeletConfiguration,
 	}
 
 	serviceIndexer := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc})
+	var serviceLister corelisters.ServiceLister
```
h-w-chen, line 442: not sure it can build or not.

h-w-chen (commented):
is this PR for internal discussion purpose, or ready fro peer review? From title & content, it seems internal discussion only.

pkg/kubelet/kubelet.go
q131172019: Can build successfully. I tested it last night.

pkg/kubelet/kubelet.go
```diff
@@ -439,12 +439,20 @@ func NewMainKubelet(kubeCfg *kubeletconfiginternal.KubeletConfiguration,
...
+
+		for i := range serviceLister {
+			serviceLW := cache.NewListWatchFromClient(kubeDeps.KubeTPClients[i].CoreV1(), "services", metav1.NamespaceAll, fields.Everything())
+			r := cache.NewReflector(serviceLW, &v1.Service{}, serviceIndexer, 0)
+			go r.Run(wait.NeverStop)
+
+			serviceLister[i] = corelisters.NewServiceLister(serviceIndexer)
+		}
+	} else {
+		serviceLister = corelisters.NewServiceLister(serviceIndexer)
```
yb01, line 454: this should be an error case.

pkg/kubelet/kubelet.go
```diff
@@ -439,12 +439,20 @@ func NewMainKubelet(kubeCfg *kubeletconfiginternal.KubeletConfiguration,
...
 
 	serviceIndexer := cache.NewIndexer(cache.MetaNamespaceKeyFunc, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc})
+	var serviceLister corelisters.ServiceLister
 	if hasValidTPClients(kubeDeps.KubeTPClients) {
-		serviceLW := cache.NewListWatchFromClient(kubeDeps.KubeTPClients[0].CoreV1(), "services", metav1.NamespaceAll, fields.Everything())
-		r := cache.NewReflector(serviceLW, &v1.Service{}, serviceIndexer, 0)
-		go r.Run(wait.NeverStop)
+		serviceLister := make([]corelisters.ServiceLister, len(kubeDeps.KubeTPClients))
+
+		for i := range serviceLister {
```
yb01, line 446: nit: better to use the source length which is the kubeClient's length

pkg/kubelet/kubelet.go
yb01: it probably can compile but this block of code can be made a bit clear as

if ! hasValidClients
    return error
declare the servicelist slice with length
create the servicelist and populate the slice

yb01:
Has the code to consume the serverList already been changed to treat it as a slice ?

pkg/kubelet/kubelet.go
q131172019: Yes, in line 444, use len(kubeDeps.KubeTPClients)

q131172019 (commented):
Resolve conversation.

pkg/kubelet/kubelet.go
q131172019: Add Error when ! hasValidTPClients(kubeDeps.KubeTPClients)

pkg/kubelet/kubelet.go
```diff
@@ -439,12 +439,22 @@ func NewMainKubelet(kubeCfg *kubeletconfiginternal.KubeletConfiguration,
...
+
+		for i := range serviceListers {
+			serviceLW := cache.NewListWatchFromClient(kubeDeps.KubeTPClients[i].CoreV1(), "services", metav1.NamespaceAll, fields.Everything())
+			r := cache.NewReflector(serviceLW, &v1.Service{}, serviceIndexer, 0)
+			go r.Run(wait.NeverStop)
+
+			serviceListers[i] = corelisters.NewServiceLister(serviceIndexer)
+		}
+	} else {
+		klog.Errorf("No Valid TP Clients: %v", err)
```
yb01, line 454: if the hasValidTPclient() returns false, then the kubeclient is nil or 0 length. this else block should return error.

pkg/kubelet/network/dns/dns.go
```diff
@@ -98,19 +108,49 @@ func (c *Configurer) getClusterDNS(pod *v1.Pod) ([]net.IP, error) {
...
+		klog.V(4).Infof("Need locate which serviceLister is correct one in multi-servceListers")
+		// To get correct index of serviceLister based on Pod's tenant information
+		index = c.clientManager.PickClient(pod.Tenant);
 	} else {
-		klog.Errorf("failed to get network %s/%s: %v", pod.Tenant, networkName, err)
+		// Only has one serviceLister
+		index = 0;
+	}
+	klog.V(4).Infof("Pod NAME: %q | Index : %q | networkName : %q | Tenant : %q ", pod.Name, index,  networkName, pod.Tenant)
+	services, err := c.serviceListers[index].List(labels.Everything())
```
yb01, line 125: the logic in DNS below ( in the for loop ) only needs the dns service name, so it probably better just use GET to get this service from the api server instead of a LIST

yb01 (approved):
/lgtm

pkg/kubelet/network/dns/dns.go
```diff
@@ -98,19 +108,49 @@ func (c *Configurer) getClusterDNS(pod *v1.Pod) ([]net.IP, error) {
...
+                return nil, fmt.Errorf("failed to get serviceListers when getting service IP for network")
+	}
 
-		ip := net.ParseIP(network.Status.DNSServiceIP)
-		if ip != nil {
-			return []net.IP{ip}, nil
-		}
+	var index int
+	if len(c.serviceListers) > 1 {
+		klog.V(4).Infof("Need locate which serviceLister is correct one in multi-servceListers")
```
h-w-chen, line 117: nit: consider rephrase the message to indicating which one is chosen

h-w-chen (approved):

PR #1183, opened 2021-09-15
Send fake bookmark event to client to reduce size of initEvents

Sindica:


staging/src/k8s.io/apiserver/pkg/storage/cacher/cacher.go
```diff
@@ -1361,6 +1361,18 @@ func (c *cacheWatcher) process(ctx context.Context, initEvents []*watchCacheEven
 		klog.V(2).Infof("processing %d initEvents of %s took %v", len(initEvents), objType, processingTime)
 	}
 
+	// send fake bookmark event
```
yb01, line 1364: this is not fake anymore. it is a valid to keep the client's high watermark for idle watch sessions.

yb01 (approved):

yb01 (approved):
/lgtm

staging/src/k8s.io/apiserver/pkg/storage/cacher/cacher.go
Sindica: I still think it is fake as the event body was not supposed to be send to client. Unless we clean up the event body, it is not a real bookmark event.

PR #1185, opened 2021-09-15
Update apiserver for large-scale cluster: target-ram-mb, max_request_inflight,max_mutating_request_inflight

sonyafenge:
**What type of PR is this?**

> /kind bug

update apiserver for large-scale cluster:
```
target-ram-mb = ${NUM_NODES} * 60
max_request_inflight=20000
max_mutating_request_inflight=20000
```

Fixes #

```release-note

```


cluster/gce/gci/configure-helper-common.sh
```diff
@@ -1808,16 +1808,21 @@ function start-kube-apiserver {
...
+       max_mutating_request_inflight=20000
     else
-      if [[ "${NUM_NODES}" -gt 3000 ]]; then
-        max_request_inflight=3000
-        max_mutating_request_inflight=1000
-      elif [[ "${NUM_NODES}" -gt 500 ]]; then
-        max_request_inflight=1500
-        max_mutating_request_inflight=500
+      if [[ "${NUM_NODES}" -gt 10000 ]]; then
+        max_request_inflight=5000
```
Sindica, line 1816: Where is 5000/2000 come from? I don't have data in hand. Not sure whether 20K cluster can have mutating request restricted to 2000. For 25K cluster, it spikes to 6K. I would not think 2K is enought for 20K cluster.

sonyafenge: Check the history and found 5000/2000 was added by Vinay when move scale-out from POC to master. For 40000/20000 is the data that we are used for current 1*1*25K and 2*2*25K

cluster/gce/gci/configure-helper-common.sh
```diff
@@ -1809,8 +1809,8 @@ function start-kube-apiserver {
 
     # If the cluster is large, increase max-requests-inflight limit in apiserver.
     if [[ "${NUM_NODES}" -gt 10000 ]]; then
-       max_request_inflight=5000
-       max_mutating_request_inflight=2000
+      max_request_inflight=5000
```
Sindica, line 1812: Since we are consistently testing in 25K and larger cluster, can you add 20000 as max_request_inflight and max_mutating_request_inflight for 10K+ cluster? Since this is an open source code, don't want people use default and find out it does not work. Also we are now having much more mutation than read, there is no need to set max_request_inflight higher than max_mutating_request_inflight.

sonyafenge: I will open a new PR for max_request_inflight and max_mutating_request_inflight. This PR is only for target-ram-mb for now

Sindica (approved):
/lgtm

PR #1190, opened 2021-09-17
scale-up.sh enables network service support by default

### LINKED ISSUE #1152: [service support] make arktos-up able to have network/service support
to provide a relatively easy approach (option) for scale-up/scale-out dev cluster with service support in place

plan to introduce an env var, say ENABLE_ARKTOS_NETWORK_SERVICE.

**Why is this needed**:
to make starting cluster with service support easy

h-w-chen:
**What type of PR is this?**
/kind feature

This PR automates the start of scale-up local dev cluster with the network service support enabled by default.
If such support is not desired, setting env var DISABLE_NETWORK_SERVICE_SUPPORT will disable this feature.

Fixes #1152 

this PR depends on following PRs in queue:
#1172
#1174 
#1176 (or #1181)
#1177
#1186 

YES
```release-note
the local scale-up dev env started by hack/scale-up.sh will be default have network service support enabled - each pod will use its network specific DNS service.
```


yb01:
this enables the local setup, do we also need this enabled in the perf runs in order to get perf-run clusters up? if so, the same change should make into the cluster/gce/gci/configure-helper-common.sh as well.

q131172019 (approved):

sonyafenge:
should we also need this for Kube-up and kubemark?

PR #1197, opened 2021-09-22
bug fix: provide resolv.conf if the default is unavailable

### LINKED ISSUE #1196: ubuntu 16.04 kubelet has error of "open /run/systemd/resolve/resolv.conf: no such file or directory"
**What happened**:
kubelet is emitting errors like:
```
E0922 21:08:51.291169   25968 kuberuntime_manager.go:1024] createPodSandbox for pod "nginx-deployment-latency-7d8ffcd9fd-5w7dm_default_system(c454b88d-05cd-4f48-be63-6994e4e4899d)" failed: open /run/systemd/resolve/resolv.conf: no such file or directory
```
and pods stuck in Creating state

pods in Running state, and kubelet has not such error

in Ubuntu 16.04 LTS, run ```./hack/arktos-up.sh```

Ubuntu 18.04 or 20.04 LTS seem to have no such issue

**Environment**:
- OS (e.g: `cat /etc/os-release`): Ubunttu 16.04
- Kernel (e.g. `uname -a`):
- Install tools:
- Others:

h-w-chen:
**What type of PR is this?**
/kind bug

service support requires kubelet to have a resolv.conf that points to a valid external DNS root w/o forming a self-loop chain. current arktos-up script uses /run/systemd/resolve/resolv.conf file which is available in Ubuntu 18.04 and 20.04. However, on Ubuntu 16.04, there is no such file, hence kubelet fails.

This PR fixes the problem by providing a minimalist resolv.conf when the default one is not there, and kubelet requires an external anyway ehn service support is enabled.

Fixes #1196

NONE

### DIFF
--- a/hack/lib/common-var-init.sh
+++ b/hack/lib/common-var-init.sh
@@ -215,7 +215,19 @@ REUSE_CERTS=${REUSE_CERTS:-false}
 #
 # Avoid to create loop error at https://coredns.io/plugins/loop/#troubleshooting
 # when create coredns pod
-RESOLV_CONF=${RESOLV_CONF:-"/run/systemd/resolve/resolv.conf"}
+RESOLV_CONF=${RESOLV_CONF:-}
+if [ -z ${RESOLV_CONF} ] && [ -z ${DISABLE_NETWORK_SERVICE_SUPPORT} ]
+then
+  if [ -f "/run/systemd/resolve/resolv.conf" ]; then
+    RESOLV_CONF=${RESOLV_CONF:-"/run/systemd/resolve/resolv.conf"}
+  else
+    RESOLV_CONF=$(mktemp -p /tmp)
+    cat << 'EOF' > ${RESOLV_CONF}
+nameserver 8.8.8.8
+EOF
+  fi
+fi
+
 # --------------------------------------------------------------------------------------------
 # End of 2nd Common environment variables used in arktos-up.sh& arktos-apiserver-partition.sh
 # --------------------------------------------------------------------------------------------


hack/lib/common-var-init.sh
```diff
@@ -215,7 +215,19 @@ REUSE_CERTS=${REUSE_CERTS:-false}
...
-RESOLV_CONF=${RESOLV_CONF:-"/run/systemd/resolve/resolv.conf"}
+RESOLV_CONF=${RESOLV_CONF:-}
+if [ -z ${RESOLV_CONF} ] && [ -z ${DISABLE_NETWORK_SERVICE_SUPPORT} ]
+then
+  if [ -f "/run/systemd/resolve/resolv.conf" ]; then
+    RESOLV_CONF=${RESOLV_CONF:-"/run/systemd/resolve/resolv.conf"}
+  else
+    RESOLV_CONF=$(mktemp -p /tmp)
+    cat << 'EOF' > ${RESOLV_CONF}
+nameserver 8.8.8.8
```
Sindica, line 226: What does "nameserver 8.8.8.8" do?

BTW, I have tested this change on my host. W or w/o DISABLE_NETWORK_SERVICE_SUPPORT, all works fine. Thank you for prompt fix.

h-w-chen: 8.8.8.8 is a well-known google public DNS service.

Sindica: How does it work in our cluster?

h-w-chen: hard-code 8.8.8.8 is indeed not a good idea. will change it to env var to ask user to provide a suitable DNS ip addr instead.

hack/lib/common-var-init.sh
```diff
@@ -215,7 +215,24 @@ REUSE_CERTS=${REUSE_CERTS:-false}
...
+then
+  if [ -f "/run/systemd/resolve/resolv.conf" ]; then
+    RESOLV_CONF=${RESOLV_CONF:-"/run/systemd/resolve/resolv.conf"}
+  else
+    if [ -z ${UPSTREAM_DNS_IP:-} ]
+    then
+      echo "please specify the upstream DNS ip address in env var UPSTREAM_DNS_IP" >&2
+      exit 1
+    fi
+    RESOLV_CONF=$(mktemp -p /tmp)
```
yb01, line 229: this is better under /run or /etc instead of /tmp/.

yb01 (approved):
/lgtm

hack/lib/common-var-init.sh
h-w-chen: for now let's keep in /tmp, not to mess up with host

hack/lib/common-var-init.sh
```diff
@@ -215,7 +215,24 @@ REUSE_CERTS=${REUSE_CERTS:-false}
...
+    cat << EOF > ${RESOLV_CONF}
+nameserver ${UPSTREAM_DNS_IP}
```
Sindica, line 231: Try to run arktos-up, got message "please specify the upstream DNS ip address in env var UPSTREAM_DNS_IP". There need to be a default or suggested setting. This is not user friendly.

q131172019: on AWS Ubuntu 16.04, can change nameserver from 127.0.0.53 to 8.8.8.8 in /etc/resolv.conf then create new temporary file /tmp/resolv.conf when the file /run/systemd/resolve/resolv.conf is not available.
```
sed -e "s/127.0.0.53/8.8.8.8/g" /etc/resolv.conf > /tmp/resolve.conf
```

Then codes can be changed to the following:

```
RESOLV_CONF=${RESOLV_CONF:-}
if [ -z ${RESOLV_CONF} ] && [ -z ${DISABLE_NETWORK_SERVICE_SUPPORT} ]
then
  if [ -f "/run/systemd/resolve/resolv.conf" ]; then
    RESOLV_CONF=${RESOLV_CONF:-"/run/systemd/resolve/resolv.conf"}
  else
     sed -e "s/127.0.0.53/8.8.8.8/g" /etc/resolv.conf > ./tmp/resolve.conf
     RESOLV_CONF=${RESOLV_CONF:-"/tmp/resolv.conf"}
  fi
```

PR #1199, opened 2021-09-25
arktos-up-scale-out-poc.sh to start 2x2 scale-out cluster with service support enabled

### LINKED ISSUE #1195: [service support] change hack/arktos-up-scale-out-poc.sh to start 2x2 scale-out cluster with service support
Following the [setup document (to be revised)](https://github.com/CentaurusInfra/arktos/blob/master/docs/setup-guide/scale-out-local-dev-setup.md), and run hack/arktos-up-scale-out-poc.sh on 2 TP servers and 2 RP servers will result in a scale-out cluster with the service support.

**Why is this needed**:
this task is on top of #1194 1x1 scale-out scripting. This task is more focused upon the changes required for multiple TPs and multiple RPs.

h-w-chen:
**What type of PR is this?**
/kind documentation
/kind feature

This is the last piece of service support in the scale-out architecture. It makes necessary changes to coordinate scale-out local dev cluster which has multiple TP and RP servers.

Besides scripting change, it has companion setup document change too.

Fixes #1195

YES
```release-note
hack/arktos-up-scale-out-poc.sh starts scale-out cluster having 2 TPs and 2 RPs, with the service support emnabled, and flannel cni installed.
```


docs/setup-guide/scale-out-local-dev-setup.md
```diff
@@ -59,12 +67,33 @@ Note:
 ```
 export IS_RESOURCE_PARTITION=true
 export TENANT_SERVER=[TP1_IP]<,[TP2_IP]>
+export RESOURCE_PARTITION_POD_CIDR=pod-cidr
 ```
 
+an examplative allocation of pod cidr for 2 RPs could be
```
yb01, line 73: the setup should also describe the kube-proxy's  --cluster-cidr parmaeter should be the super set of the RP cluster IPs, in this case, should be 10.0.0.0/8 if assuming all RPs will be 10.x.0.0/16 based.

h-w-chen: thanks for the directions. Haproxy seems have no place for cluster-cidr (of service IP ranges); it only cares about TP/RP IP addresses. Seems non issue.

h-w-chen: --cluster-cidr is not mandatory in our cluster; neither scale-up nor scale-out script provide value of cluster-cidr for kube-proxy.  it should be fine not specifying it.

yb01: this is used to determine the internal and external traffics, even it is not required, but from this code we should have it specified and supported.

```
if len(clusterCIDR) == 0 {
		klog.Warning("clusterCIDR not specified, unable to distinguish between internal and external traffic")
	} else if utilnet.IsIPv6CIDRString(clusterCIDR) != ipt.IsIpv6() {
		return nil, fmt.Errorf("clusterCIDR %s has incorrect IP version: expect isIPv6=%t", clusterCIDR, ipt.IsIpv6())
	}
```

h-w-chen: my point is we did not specify it in the first place with arktos-up.sh. This is not 2x2 scale-out specific issue; better to cope it out side of scale-out service support PRs. I'll create an issue to track it.

yb01: that might be OK for now. my original thought this can be just a superset of RP cluster-cidrs. a issue to track this is good for now.

h-w-chen: issue @1201 was created to track this discussion/concern

hack/arktos-up-scale-out-poc.sh
```diff
@@ -44,6 +52,19 @@ echo "TENANT_SERVER: |${TENANT_SERVER}|"
 echo "RESOURCE_SERVER: |${RESOURCE_SERVER}|"
 echo "IS_SCALE_OUT: |${IS_SCALE_OUT}|"
 
+# scale-out specific env vars
```
yb01, line 55: with multiple TPs and RPs in the setup, will the call need to call this hack/arktos-up-scale-out-poc.sh multiple times and the desired cidr for this TP and RP were set in their ? i assume there exists a sort of loop to create the TPs and RPs ( just like start-kubemark.sh ) does, so that logic should be also updated to set the 2Tp 2RP env, right ?

h-w-chen: this doc is only using 2x2, following the original style. However, as you can see, it is pretty obvious that more TP/RP would has their own srv/pod cidr.

For now, these script is expected to run manually on each TP/RP server. In the future, if it is needed so, more automation (like one controlling script to loop through TP & RP) definitely is feasible - but for now it is not in scope of this PR.

yb01 (approved):
/lgtm

a couple comments/future enhancement for this.

q131172019 (approved):
/lgtm

PR #1210, opened 2021-10-08
revisted daemonset design, added kube-proxy related change

### LINKED ISSUE #1171: scale-out revisit - various resources from different TPs should not collide
to investigate a generic way to ensure resources not conflicting each other, even if they are in same tenant/ns, having same name - as long as they are from different TPs

**Why is this needed**:
in scale-out env, some resources -in theory- of "system" tenant, could reside in multiple TPs. If a component is working on the resources from multiple TPs (like kubelet, kube-proxy), it is possible the tenant-ns-name triple, when tenant is system tenant, insufficient to differ them, which results in resource collision. 

in 930 we put constraints that not having resources in system tenants (other than the kube-dns related) to work around such conflict; however, it is still valuable to have a more genral way to resolve suvh conflict.

This task is to track the ideas/discussion/progress of this problem.

h-w-chen:
**What type of PR is this?**
/kind design
/kind documentation

This PR imporves the daemonset deign:
1. kubelet service lister related change
2. kube-proxy is covered
3. keeping track of TP origin for resources on all tenant is decided

Fixes #1171

NONE


docs/design-proposals/arch/daemonset.md
```diff
@@ -48,7 +50,21 @@ This is a variant of the above-mentioned DS in RP: cluster admin manages DaemonS
...
+Only resources of system tenant need to keep track of TP origin; user tenant has not change.
+
+Its pro is smaller memory footprint; smaller impact on existing system design.
+
+**All tenants**
+All reources used in kubelet (and kube-proxy), regardless of tenat they are in, keep track of theit TP origin.
+
+Its pro is consistency, no assumption that user tenant being on one single TP.
+
+We'll implement by keeping track for all tenants.
```
yb01, line 65: +1

docs/design-proposals/arch/daemonset.md
```diff
@@ -48,7 +50,21 @@ This is a variant of the above-mentioned DS in RP: cluster admin manages DaemonS
 
 This architecture mitigates the DS sync problem, but it brings more incompatibility of scale-out and scale-up, as scale-up has no such sync in the first place.
 
-### Reequired Changes
+Regarding the scope of changes to keep track of TP origin, there are options:
```
Sindica, line 53: This statement starts a new section of discussion. Please add proper section title to make it obvious.

docs/design-proposals/arch/daemonset.md
Sindica: IIRC, service support also keeps all resource origin. It is a universal design, you can add it as one of the pros.

docs/design-proposals/arch/daemonset.md
```diff
@@ -72,3 +88,10 @@ There are two significant required functionalities for kubelet, at high level:
 2. Retrieving supporting resources by TP origin.
    <p/>Secret and configmap resources each is managed via a local cached store being replenished by external API servers (TP's). Current store implementation simply derives TP by tenant name of the resource. To accomodate the DaemonSet design, the cache store should accept requests to get resource with specified TP origin; it should be able to keep system-tenanted secret or configmap without losing their TP origins; also it should use the resource-specific TP origin to replenish the staled resource instead of the simple tenant-name derived one.
    <p/>The specific detail the changes of store should be carefully considered before the kubelet module implementation.
+3. Keeping track of TP origin of service in Service lister, using services from proper TP when identifying services of the Pod.
```
Sindica, line 91: There are <b>three</b> significant required functionalities for kubelet?

How is this done in service support for scale out? I assume it is very similar. Can there a design doc that can be referred here?

Sindica: Line 87, segguested - > suggested

Sindica: Also, I don't think it is best to derive TP origin from tenant name. There is no such fixed configuration. It is better track the mapping in kubelet internal memory when the info was first received. Isn't this how it was implemented in service support for scale out?

docs/design-proposals/arch/daemonset.md
```diff
@@ -2,6 +2,8 @@
 
 Due to the nature Arktos being a multiple-tenanted system, DaemonSet is only allowed in system tenant; only cluster admin is permitted to create/delete/update DaemonSet resources.
 
+Multiple-TP also causes similar issue to Kube-proxy; changes to kube-proxy is covered in this design.
+
 ### What is not covered yet
 At this stage of Arktos, PVC/PV for storage volume is not fully supported - out of scope of this DaemonSet design.
```
yb01, line 8: any reason PV/PVC is related with this design?

docs/design-proposals/arch/daemonset.md
```diff
@@ -48,7 +50,21 @@ This is a variant of the above-mentioned DS in RP: cluster admin manages DaemonS
```
yb01, line 50: this is in the previous doc, so it is not commentable. but some clarification is neede大：
line 18:
>>Different Daemonsets can reside on different Tenant Partitions. At one TP, DaemonSet should be uniquely >>named; however, a DaemonSet of TP could have the same name as one managed by aother TP - though this >>practice is not recommended.

this is only for system tenant what is causing problems in the multiple TP env. i assumed we are on the same context but it is better to clarify here.

docs/design-proposals/arch/daemonset.md
yb01: what is the mapping used for ? any other resources it needs to explicitly get the desired TP to get ?

below is a mail we have long time ago:
if just for secrets or configmaps etc. i thought we have looked into this and the kubelet should be able get those info per POD which is under the pod ID.

Secrets, configmaps were stored as paths as below where the POD UID is unique per the tenant-ns-podName. E.g. below is the flannel token secret path on the node.  In a worst case where the system pods were from different TP, it should be OK at the node in terms of loading secrets.
 
./var/lib/kubelet/pods/0a376aa2-9ec3-4fa4-bb48-d85e76991dcc/volumes/kubernetes.io~secret/flannel-token-k65mf

docs/design-proposals/arch/daemonset.md
```diff
@@ -2,6 +2,8 @@
 
 Due to the nature Arktos being a multiple-tenanted system, DaemonSet is only allowed in system tenant; only cluster admin is permitted to create/delete/update DaemonSet resources.
 
+Multiple-TP also causes similar issue to Kube-proxy; changes to kube-proxy is covered in this design.
+
 ### What is not covered yet
```
yb01, line 7: can we add feature mentioned in line 28 ( recover DS from failed TPs ) to the not covered in this doc ?

docs/design-proposals/arch/daemonset.md
h-w-chen: 1-  will add a sentence stating the discussion applies to resources of both kubelet and kube-proxy (which includes service).
2-  defining one set of the secrets (and possible other supporting resources) that applies to all TPs is interesting; however, it is not the concern of current design.
3-  will rewrite section of requried changes, to cites resources of kubelet to be pod, configmap, secret, service, resourcces of kube-proxy to be service and endpoint.

docs/design-proposals/arch/daemonset.md
h-w-chen: will make subsection

docs/design-proposals/arch/daemonset.md
h-w-chen: 1- 3 is out-of-date. will reword it; will correct typos; 
2- Service resource will be mentioned in section of kube-proxy change;
3- yes the current implementation is indeed tracking the tenant's origin on first-time receipt. will re-phase to reduce the confusion.

docs/design-proposals/arch/daemonset.md
h-w-chen: current arktos does npt support PV/PVC in multi-tenant yet.

docs/design-proposals/arch/daemonset.md
h-w-chen: will add of system tenant only to re-state it.

docs/design-proposals/arch/daemonset.md
h-w-chen: will add it to no scope.

Sindica (approved):
/lgtm

yb01 (approved):
/lgtm

Sindica (approved):

PR #1215, opened 2021-10-21
Bug Fix: Issue 1211 - --kube-apiserver-ip is set to a set of IP addresses by parameter "${API_HOST_IP_EXTERNAL}"

q131172019:
**What type of PR is this?**
> /kind bug

When ./hack/arktos.up script calls the script hack/lib/common.sh to start arktos network controller, the parameter kube-apiserver-ip is set to a set of IP address by another parameter "${API_HOST_IP_EXTERNAL}" so that arktos network controller can not be started. The reason for parameter "${API_HOST_IP_EXTERNAL}" to have a set of IP address is to use command 'hostname -i' at line 181 in hack/lib/common-var-init.sh.

API_HOST_IP_EXTERNAL=${API_HOST_IP_EXTERNAL:-"${hostip}"}

This needs to be fixed so that a single IP address is used as parameter by getting the interface from default route

Fixes # 1211

The bug was found when PR1114 of Mizar was tested.

NONE


hack/lib/common-var-init.sh
```diff
@@ -178,7 +178,25 @@ API_SECURE_PORT=${API_SECURE_PORT:-6443}
 # WARNING: For DNS to work on most setups you should export API_HOST as the docker0 ip address,
 API_HOST=${API_HOST:-"${lohostname}"}
 API_HOST_IP=${API_HOST_IP:-"0.0.0.0"}
-API_HOST_IP_EXTERNAL=${API_HOST_IP_EXTERNAL:-"${hostip}"}
+
+# Fix the issue 1211 at https://github.com/CentaurusInfra/arktos/issues/1211
+API_HOST_IP_EXTERNAL=${API_HOST_IP_EXTERNAL:-}
+if [ -z ${API_HOST_IP_EXTERNAL} ]
+then
+  if [ -f "/etc/netplan/50-cloud-init.yaml" ]; then
```
yb01, line 186: this is not a generic approach and it will fail on some machines. e.g. on my GCE VM, this file does not exist.

q131172019: Currently the bug is found in AWS Ubuntu 18.05 for Mizar. so we can fix the issue for this case on AWS Ubuntu 18.05. If your machine is GCE VM and the file does not exist, the script gives warning and exit. You have to manually set the parameter API_HOST_IP_EXTERNAL.

yb01: we cannot fix one env but break another one that worked before.

q131172019: I make the more generic changes based on your suggestion to support Ubuntu 16.04 and GCE VM.

q131172019:
If issue is still found, the PR will be reopened to address the issue.

PR #1235, opened 2021-11-25
write config to stop running of mizar kopf when cluster is arktos

Hong-Chang:
**What type of PR is this?**
/kind bug

Write config to stop running of mizar kopf when cluster is arktos.
Mizar depends on such config to decide whether the cluster is arktos or k8s. So the config needs to be written to let mizar know it's arktos.

Current mizar controllers and kopf are both running which is unexpected. 

NONE

### DIFF
--- a/hack/arktos-up.sh
+++ b/hack/arktos-up.sh
@@ -561,6 +561,7 @@ fi
 
 # Applying mizar cni
 if [[ "${CNIPLUGIN}" == "mizar" ]]; then
+  ${KUBECTL} create configmap system-source --namespace=kube-system --from-literal=name=arktos --from-literal=company=futurewei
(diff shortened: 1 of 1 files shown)


hack/arktos-up.sh
```diff
@@ -561,6 +561,7 @@ fi
 
 # Applying mizar cni
 if [[ "${CNIPLUGIN}" == "mizar" ]]; then
+  ${KUBECTL} create configmap system-source --namespace=kube-system --from-literal=name=arktos --from-literal=company=futurewei
```
Sindica, line 564: Do you need pass in kubeconfig?

Hong-Chang: Not needed.

Sindica (approved):
/lgtm

hack/arktos-up.sh
Hong-Chang: Added.

PR #1239, opened 2021-12-06
Simplify worker up

Sindica:
Run worker up for arktos up with Mizar (aws):
1. Set up dev host with https://github.com/q131172019/arktos/blob/CarlXie_singleNodeArktosCluster/docs/setup-guide/single-node-dev-scale-up-cluster-with-Mizar.md
2. Start master node with Mizar
3. Copy /var/run/kubernetes/client-ca.crt from master to worker
4. Start worker with Mizar: 
```
API_HOST=<master node name:-ip-172-30-0-14> CNIPLUGIN=mizar ./hack/arktos-worker-up.sh
```

Tested with 1 master and 3 workers

### DIFF
--- a/hack/arktos-worker-up.sh
+++ b/hack/arktos-worker-up.sh
@@ -31,7 +31,7 @@ function write_client_kubeconfig {
     local api_port=$5
     local client_id=$6
     local token=${7:-}
-    local protocol=${8:-"https"}
+    local protocol=${8:-"http"}
 
     cat <<EOF | ${sudo} tee "${dest_dir}"/"${client_id}".kubeconfig > /dev/null
 apiVersion: v1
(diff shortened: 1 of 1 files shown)


hack/arktos-worker-up.sh
```diff
@@ -17,29 +17,56 @@
...
+# creates a self-contained kubeconfig: args are sudo, dest-dir, ca file, host, port, client id, token(optional)
+function write_client_kubeconfig {
+    local sudo=$1
+    local dest_dir=$2
+    local ca_file=$3
+    local api_host=$4
+    local api_port=$5
+    local client_id=$6
+    local token=${7:-}
+    local protocol=${8:-"https"}
```
h-w-chen, line 34: is the prtocol default https or http?

h-w-chen (approved):
lgtm

hack/arktos-worker-up.sh
Sindica: You are right, we don't support https now. It should default to http.

> is the prtocol default https or http?

PR #1245, opened 2021-12-09
Remove APIServerConfigManager usage from KCM - not currently used - does not work well with scale out

Sindica:


h-w-chen (approved):
lgtm

PR #1250, opened 2021-12-22
Kube proxy multi tenant scaleout

Sindica:
Tested with PR 1244 with 2TP1RP.
Deployed a whoami service in both TP, deployed another pod in both TP for curling purpose. All pods are in system tenant. Was able to curl service endpoint and get back correct whoami container id.

```
---
apiVersion: v1
kind: Service
metadata:
  name: whoami-service
  namespace: default
spec:
  selector:
    app: whoami
  ports:
  - protocol: TCP
    port: 8000
    targetPort: 80
---
apiVersion: apps/v1 
kind: Deployment
metadata:
  name: whoami-deployment
spec:
  replicas: 2
  selector:
    matchLabels:
      app: whoami
  template:
    metadata:
      labels:
        app: whoami
    spec:
      containers:
      - image: ctmagazin/whoami 
        name: whoami-container
        ports:
        - containerPort: 8000
```

### DIFF
--- a/pkg/proxy/config/config.go
+++ b/pkg/proxy/config/config.go
@@ -1,5 +1,6 @@
 /*
 Copyright 2014 The Kubernetes Authors.
+Copyright 2020 Authors of Arktos - file modified.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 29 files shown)


staging/src/k8s.io/apimachinery/pkg/types/namespacedname.go
```diff
@@ -43,3 +43,16 @@ const (
 func (n NamespacedName) String() string {
 	return fmt.Sprintf("%s%c%s%c%s", n.Tenant, Separator, n.Namespace, Separator, n.Name)
 }
+
+// NamespacednameWithTenantSource is used to identify a service and endpoint in scale out architecture
```
h-w-chen, line 47: is this also representing a service in scale-up arch as well?

pkg/proxy/config/config.go
```diff
@@ -149,20 +152,22 @@ func (c *EndpointsConfig) handleDeleteEndpoints(obj interface{}) {
...
 	}
 }
 
 // ServiceConfig tracks a set of service configurations.
 type ServiceConfig struct {
-	listerSynced  cache.InformerSynced
-	eventHandlers []ServiceHandler
+	listerSynced      cache.InformerSynced
+	eventHandlers     []ServiceHandler
+	tenantPartitionId int
```
h-w-chen, line 163: nit - in scale-up there is no tenant partition. suggest to consider a name suited in both scale-out and scale-up.

h-w-chen (approved):
lgtm. Left a few comments but they are minor issues; non blocking.

staging/src/k8s.io/apimachinery/pkg/types/namespacedname.go
Sindica: In scale-up arch, there is single source of tenant. It does not hurt to have an extra const field. For better understanding in scale out arch, I prefer to specifically call it out. It should not cause confusing in scale up.

pkg/proxy/config/config.go
Sindica: Keep the original as scale up can have single tenantPartitionId.

PR #1261, opened 2022-01-04
Bug fix for scale out local dev env: set FIRST_SERVICE_CLUSTER_IP from TENANT_PARTITION_SERVICE_SUBNET

Sindica:
This change adds kubernetes service cluster ip to serving certkey


hack/arktos-up-scale-out-poc.sh
```diff
@@ -57,6 +57,12 @@ echo "IS_SCALE_OUT: |${IS_SCALE_OUT}|"
 export TENANT_SERVER_NAME=${TENANT_SERVER_NAME:-${API_HOST}}
 export TENANT_PARTITION_SERVICE_SUBNET=${TENANT_PARTITION_SERVICE_SUBNET:-10.0.0.0/16}
 SERVICE_CLUSTER_IP_RANGE=${TENANT_PARTITION_SERVICE_SUBNET}
+
+# get FIRST_SERVICE_CLUSTER_IP
```
h-w-chen, line 61: better not to override in case explicitly specified;
simple assign 1 to last segment is right for almost all practical uses; fine with that.

q131172019: agree.

Sindica: > better not to override in case explicitly specified; simple assign 1 to last segment is right for almost all practical uses; fine with that.

Not sure what you meant by override? Which variable is overriden?

h-w-chen: FIRST_SERVICE_CLUSTER_IP

Sindica: FIRST_SERVICE_CLUSTER_IP was initialized in hack/lib/common-var-init.sh with default value 10.0.0.1. It was not set in anywhere else. In our scale out case, it should be set to based on TENANT_PARTITION_SERVICE_SUBNET value passed in from commandline arg.

h-w-chen: for setting FIRST_SERVICE_CLUSTER_IP, here is always setting based on tp-service-range, the other place (in hack/lib/common-var-init.sh) is falling back in case not set. Better to make them consistent.

h-w-chen (approved):
lgtm. left minor comment for author to consider. not blocking.

PR #1263, opened 2022-01-04
in arktos-up-scale-out-poc.sh, write config to stop running of mizar kopf when cluster is arktos

Hong-Chang:
What type of PR is this?
/kind bug

Write config to stop running of mizar kopf when cluster is arktos and it's in roll-out mode. The same thing has been done for arktos-up.sh via https://github.com/CentaurusInfra/arktos/pull/1235. It needs to be done for file hack/arktos-up-scale-out-poc.sh.
Mizar depends on such config to decide whether the cluster is arktos or k8s. So the config needs to be written to let mizar know it's arktos.

Which issue(s) this PR fixes:
In Arktos roll-out, current mizar controllers and kopf are both running which is unexpected.

NONE


Sindica (approved):
/lgtm

PR #1264, opened 2022-01-05
add Ubuntu support for kube-up.sh

sonyafenge:
**What type of PR is this?**

> /kind feature

Add python3 installation to support ubuntu for kube-up.sh. 

**Existing issue for this PR**:
Cluster can be started with PR, but if cni plugin using flannel, several pods failed to start: metrics-server, kubernetes-dashboard, event-exporter, corednns. There will be a follow-up PR to fix this issue.

Fixes #

```release-note

```

### DIFF
--- a/cluster/addons/dns/coredns/coredns.yaml.base
+++ b/cluster/addons/dns/coredns/coredns.yaml.base
@@ -129,6 +129,8 @@ spec:
         - name: config-volume
           mountPath: /etc/coredns
           readOnly: true
+        - name: tmp
+          mountPath: /tmp
         ports:
         - containerPort: 53
           name: dns
@@ -163,6 +165,8 @@ spec:
           readOnlyRootFilesystem: true
       dnsPolicy: Default
       volumes:
+        - name: tmp
+          emptyDir: {}
         - name: config-volume
           configMap:
             name: coredns
--- a/cluster/gce/gci/configure.sh
+++ b/cluster/gce/gci/configure.sh
@@ -56,13 +56,15 @@ EOF
 }
 
 function validate-python {
-  local ver=$(python -c"import sys; print(sys.version_info.major)")
-  echo "python version: $ver"
-  if [[ $ver -ne 2 && $ver -ne 3 ]]; then
+  local ver=$(python3 -c"import sys; print(sys.version_info.major)")
+  echo "python3 version: $ver"
+  if [[ $ver -ne 3 ]]; then
     apt -y update
-    apt install -y python
-    apt install -y python-pip
+    apt install -y python3
+    apt install -y python3-pip
     pip install pyyaml
+  else
+    echo "python3: $ver is running.."
   fi
 }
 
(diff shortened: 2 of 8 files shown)


sonyafenge:
investigation for flannel plugin is in progress. This PR can be used for Mizar integration.

cluster/addons/dns/coredns/coredns.yaml.base
```diff
@@ -129,6 +129,8 @@ spec:
         - name: config-volume
           mountPath: /etc/coredns
           readOnly: true
+        - name: tmp
```
h-w-chen, line 132: is there a reason we need to explicitly define /tmp here?

cluster/gce/gci/configure.sh
```diff
@@ -56,13 +56,15 @@ EOF
 }
 
 function validate-python {
-  local ver=$(python -c"import sys; print(sys.version_info.major)")
-  echo "python version: $ver"
-  if [[ $ver -ne 2 && $ver -ne 3 ]]; then
+  local ver=$(python3 -c"import sys; print(sys.version_info.major)")
```
h-w-chen, line 59: 1. isn't python to be installed for ubuntu 20.04+ version 3?
2. besides this imperative installation, is it possible to leverage cloud-init way?

cluster/gce/gci/scaleoutserver-helper.sh
```diff
@@ -85,7 +85,7 @@ function create-proxy-instance-internal() {
       --zone "${ZONE}" \
       --machine-type "${MASTER_SIZE}" \
       --image-project="${PROXY_IMAGE_PROJECT:-ubuntu-os-cloud}" \
-      --image "${PROXY_IMAGE:-ubuntu-1804-bionic-v20201014}" \
+      --image "${PROXY_IMAGE:-ubuntu-2004-focal-v20211202}" \
```
h-w-chen, line 88: if PROXY_IMAGE is safely assumed, no need to have the default fallback, which could simplify code a bit.

h-w-chen (commented):
not blocking others' approval.

cluster/addons/dns/coredns/coredns.yaml.base
sonyafenge: This is a cherry-pick from Kubernetes: https://github.com/kubernetes/kubernetes/pull/82128.
Arktos Coredns did have same issue before cherry-pick.

cluster/gce/gci/scaleoutserver-helper.sh
sonyafenge: thanks and updated.

cluster/gce/gci/configure.sh
sonyafenge: Looks like new version of ubuntu build already add python installation. but anyway, we have conditions and if python already exist, will skip installation.

h-w-chen (approved):
lgtm

PR #1266, opened 2022-01-06
openstack essential flavors and images support for VM creations -- part 1

yb01:
This is the first part of the VM creation to use the openstack flavors and images.

### DIFF
--- a/staging/src/k8s.io/apiserver/pkg/util/openstack/flavors.go
+++ b/staging/src/k8s.io/apiserver/pkg/util/openstack/flavors.go
@@ -1,0 +1,58 @@
+/*
+Copyright 2021 The Arktos Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package openstack
+
+import "fmt"
+
+var flavors = map[string]FlavorType{}
+
+// slim down version of the openstack flavor data structure
+type FlavorType struct {
+	Id          int
+	Name        string
+	MemoryMb    int
+	Vcpus       int
+	RootGb      int
+	EphemeralGb int
+}
+
+func initFlavorsCache() {
+	flavors = make(map[string]FlavorType)
+	flavors["m1.tiny"] = FlavorType{1, "m1.tiny", 512, 1, 0, 0}
+	flavors["m1.small"] = FlavorType{2, "m1.small", 2048, 1, 0, 0}
+	flavors["m1.medium"] = FlavorType{3, "m1.medium", 4096, 2, 0, 0}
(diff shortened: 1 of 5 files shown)


staging/src/k8s.io/apiserver/pkg/util/openstack/flavors.go
```diff
@@ -0,0 +1,58 @@
...
+
+func GetFalvor(name string) (*FlavorType, error) {
+	if flavor, found := flavors[name]; found {
+		return &flavor, nil
+	}
+
+	return nil, fmt.Errorf("flavor %s, not found", name)
+}
+
+func ListFalvors() []*FlavorType {
```
Sindica, line 50: This function returns array of pointers. No need to construct array every time.

staging/src/k8s.io/apiserver/pkg/util/openstack/images.go
```diff
@@ -0,0 +1,54 @@
...
+func GetImage(name string) (*ImageType, error) {
+	if image, found := images[name]; found {
+		return &image, nil
+	}
+
+	return nil, fmt.Errorf("image %s, not found", name)
+}
+
+func ListImages() []*ImageType {
+	imageList := make([]*ImageType, len(flavors))
```
Sindica, line 47: Same here

staging/src/k8s.io/apiserver/pkg/util/openstack/openstackutils.go
```diff
@@ -38,12 +38,21 @@ func init() {
 	}
 
 	POD_JSON_STRING_TEMPLATE = string(t)
+
+	initFlavorsCache()
+	initImagesCache()
+
+	fl := ListFalvors()
+	klog.(V6).Infof("built-in flavors: %v", fl)
```
Sindica, line 46: I don't know klog.(V6) work

yb01: typo, fixed. thanks.

staging/src/k8s.io/apiserver/pkg/util/openstack/flavors.go
yb01: This is intentional to rebuild when the function is called. for couple of reasons:  flavors, images will be added in future; and also the map is used for the structure for quick GET() support hence the conventing to the slices from map values.

Sindica: In this code, all flavors are added at the init cache stage. There is no add function to support dynamic adding/delete flavors. I don't see why the list cannot be initialized and reused.

yb01: fixed and add comment for the READ only support for now.

staging/src/k8s.io/apiserver/pkg/util/openstack/flavors.go
```diff
@@ -0,0 +1,61 @@
...
+	Id          int
+	Name        string
+	MemoryMb    int
+	Vcpus       int
+	RootGb      int
+	EphemeralGb int
+}
+
+// for 130 release, only READ operation with static list of flavors
+func initFlavorsCache() {
```
Sindica, line 35: Question: I did not find the usage of initFlavorsCache and initImagesCache. When will they be called and initialize the cache?

Sindica (approved):
/lgtm

staging/src/k8s.io/apiserver/pkg/util/openstack/images.go
yb01: fixed as well.

PR #1304, opened 2022-01-25
Start proxy in worker

Sindica:
When pod is scheduled to worker, it cannot be started correctly - missing proxy in worker. Starting proxy in worker is aligned with production minion deployment.


yb01:
Any reason kube-proxy is still needed when Mizar is in place ? is it a workaround for mizar issues ?

q131172019 (approved):
/LGTM

PR #1349, opened 2022-02-09
Annotation service with default vpc/subnet for mizar integration

Sindica:
Tested service support with arktos-up:

```
ubuntu@ip-172-30-0-14:~/go/src/k8s.io/arktos$ ./cluster/kubectl.sh get pods -AT -o wide
TENANT   NAMESPACE     NAME                               HASHKEY               READY   STATUS    RESTARTS   AGE   IP            NODE             NOMINATED NODE   READINESS GATES
aaa      kube-system   coredns-default-65df55d94b-bh7wx   4337140920005754809   1/1     Running   0          15m   196.47.0.2    ip-172-30-0-14   <none>           <none>
system   default       mizar-daemon-sbk8g                 8004198104059958592   1/1     Running   0          18m   172.30.0.14   ip-172-30-0-14   <none>           <none>
system   default       mizar-operator-76dc9bb758-dz5sq    3443165855150812074   1/1     Running   0          18m   172.30.0.14   ip-172-30-0-14   <none>           <none>
system   default       netpod1                            3720414259593175506   1/1     Running   0          15m   42.181.0.4    ip-172-30-0-14   <none>           <none>
system   default       netpod2                            8562793631757122090   1/1     Running   0          15m   42.181.0.7    ip-172-30-0-14   <none>           <none>
system   kube-system   coredns-default-65df55d94b-gtfq9   176973603745552176    1/1     Running   0          18m   42.181.0.2    ip-172-30-0-14   <none>           <none>
system   kube-system   kube-dns-554c5866fc-p7vlz          5954901709882455418   3/3     Running   0          18m   42.181.0.3    ip-172-30-0-14   <none>           <none>
system   kube-system   virtlet-cw7qg                      4183437341678711375   3/3     Running   0          18m   172.30.0.14   ip-172-30-0-14   <none>           <none>
```

Currently, for repeated test, need to clean up with following script (will be added into mizar later):
```
for name in $(ip l |  grep -Po '(vehost-\w+)(?=@)'); do sudo ip l delete $name; done
```


pkg/controller/mizar/mizar-service-controller.go
```diff
@@ -213,115 +214,143 @@ func (c *MizarServiceController) syncService(eventKeyWithType KeyWithEventType)
...
-	networkName, hasDNSServiceLabel := service.Labels[arktosapisv1.NetworkLabel]
+	tenant, namespace, name, err := cache.SplitMetaTenantNamespaceKey(key)
+	if err != nil {
+		return err
+	}
+
+	// Get tenant default network
+	tenantDefaultNetwork, err := c.networkLister.NetworksWithMultiTenancy(tenant).Get(defaultNetworkName)
+	if tenantDefaultNetwork.Spec.Type != mizarNetworkType {
+		return nil
```
yb01, line 225: should this be an error case?

pkg/controller/mizar/mizar-service-controller.go
```diff
@@ -213,115 +214,143 @@ func (c *MizarServiceController) syncService(eventKeyWithType KeyWithEventType)
...
+	if err != nil {
+		return err
+	}
+	_, vpcNameOk := obj.Annotations[mizarAnnotationsVpcKey]
+	_, subnetNameOk := obj.Annotations[mizarAnnotationsSubnetKey]
 
-	if hasDNSServiceLabel && len(netName) != 0 {
-		klog.Infof("Starting ProcessArktosNetworkCreation network: %v", netName)
-		net, err := c.netClient.ArktosV1().NetworksWithMultiTenancy(service.Tenant).Get(netName, metav1.GetOptions{})
+	if !vpcNameOk && !subnetNameOk {
```
yb01, line 240: probably should re-update if either one of them does not exist.

pkg/controller/mizar/mizar-service-controller.go
```diff
@@ -213,115 +214,143 @@ func (c *MizarServiceController) syncService(eventKeyWithType KeyWithEventType)
...
+	klog.V(4).Infof("Assigned ip by mizar is %v", ip)
 
 	switch code {
 	case CodeType_OK:
-		klog.Infof("Mizar handled service creation successfully: %s", key)
-		if beginsWithKubernetes(service.Name) {
-			endpoint, err := c.kubeClientset.CoreV1().EndpointsWithMultiTenancy("default", "system").Get(kubernetesSvcDefaultName, metav1.GetOptions{})
+		klog.V(4).Infof("Mizar handled service creation successfully: %s", key)
+		if beginsWithKubernetes(service.Name) {	// Is this logic expected?
+			kubernetesEndpoint, err := c.kubeClientset.CoreV1().EndpointsWithMultiTenancy(metav1.NamespaceDefault, metav1.TenantSystem).Get(kubernetesSvcDefaultName, metav1.GetOptions{})
```
yb01, line 276: a service can have multiple endpoints. so this should be xxxEndpoints.

also, let's use arktos instead of kubernetes

pkg/controller/mizar/mizar-service-controller.go
```diff
@@ -213,115 +214,143 @@ func (c *MizarServiceController) syncService(eventKeyWithType KeyWithEventType)
...
 
+	// update service cluster ip with ip from mizar
 	if len(service.Spec.ClusterIP) == 0 {
 		svcToUpdate := service.DeepCopy()
 		svcToUpdate.Spec.ClusterIP = ip
-		svcUpdated, err := c.kubeClientset.CoreV1().ServicesWithMultiTenancy(service.Namespace, service.Tenant).Update(svcToUpdate)
+		_, err := c.kubeClientset.CoreV1().ServicesWithMultiTenancy(service.Namespace, service.Tenant).Update(svcToUpdate)
 		if err != nil {
-			klog.Errorf("The following service failed to update: %v", svcUpdated)
+			klog.Errorf("Failed to update service %v/%v/%v: %v", service.Tenant, service.Namespace, service.Name, err)
```
yb01, line 309: any cleanup from mizar is needed with the failure case here ?

pkg/controller/mizar/mizar-service-controller.go
```diff
@@ -213,115 +214,143 @@ func (c *MizarServiceController) syncService(eventKeyWithType KeyWithEventType)
...
 		Tenant:        service.Tenant,
 		Ip:            service.Spec.ClusterIP,
+		Vpc:           service.Annotations[mizarAnnotationsVpcKey],
+		Subnet:        service.Annotations[mizarAnnotationsSubnetKey],
 	}
 	response := c.grpcAdaptor.UpdateService(c.grpcHost, msg)
 	code := response.Code
 
 	switch code {
 	case CodeType_OK:
```
yb01, line 350: so the switch default to the function return, which is nil ?

yb01 (approved):
/lgtm to the POC.

pkg/controller/mizar/mizar-service-controller.go
Sindica: > should this be an error case?

The design of tenant network object allows arktos to support multiple network solution. It is not an error case.

pkg/controller/mizar/mizar-service-controller.go
Sindica: If one of them does not exist, this is customer error. For 130, we will not deal with it.

pkg/controller/mizar/mizar-service-controller.go
Sindica: This part is old that I did not fully understand. Will need further debugging to understand better. The service name started with kubernetes, I don't think it is the branding issue that needs to be changed now. If we need to change, all service name should be updated to arktos.

pkg/controller/mizar/mizar-service-controller.go
Sindica: Actually Mizar does not change service up so this part of code I don't think it will have chance to run. Keep it there until we have full understanding of how all services works together.

pkg/controller/mizar/mizar-service-controller.go
```diff
@@ -213,115 +214,143 @@ func (c *MizarServiceController) syncService(eventKeyWithType KeyWithEventType)
...
-		klog.Infof("Updated service: %v", svcUpdated)
 	}
 
-	if hasDNSServiceLabel && len(netName) != 0 {
-		klog.Info("[Mizar network controller] Arktos Network update starts ...")
-		net, err := c.netClient.ArktosV1().NetworksWithMultiTenancy(service.Tenant).Get(netName, metav1.GetOptions{})
+	// If this is the tenant dns service, update arktos network object with cluster ip from mizar
+	_, hasDNSServiceLabel := service.Labels[arktosapisv1.NetworkLabel]
+	if hasDNSServiceLabel {
+		networkObj, err := c.networkLister.NetworksWithMultiTenancy(service.Tenant).Get(defaultNetworkName)
```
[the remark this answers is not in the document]
Sindica: Only one default network per tenant

pkg/controller/mizar/mizar-service-controller.go
Sindica: You are right. We never touch the logic of service update as we think it is not necessary. I will check with Phu and Hong further to see how we want to handle service update.

PR #1352, opened 2022-02-11
Mizar service controller code simplify and logic fix

Sindica:
1. Fix create mizar service logic
. Remove invalid DNS Service checking logic.
. Break infinite loop for invalid VPC/subnet
2. Consolidate create and update service logic


q131172019 (approved):
/LGTM

PR #1362, opened 2022-02-15
Make local scale out have same architecture as kube up

Sindica:
Features:
. Deploy mizar operator on tp master
. Start kube-proxy and kubelet on TP master
. Use mizar templates saved in third-party folder
. Do not start daemonset from non primary TP

Test done:
1. Verified kube proxy and kubelet running on tp master only connects to tp master
2. Verified previous start up setting can start 1x1 scale out cluster

Instruction to start non primary TP:
```
export IS_SECONDARY_TP=true
```

### DIFF
--- a/hack/arktos-up-scale-out-poc.sh
+++ b/hack/arktos-up-scale-out-poc.sh
@@ -638,12 +638,8 @@ if [[ "${START_MODE}" != "nokubelet" ]]; then
         KUBELET_LOG=""
         ;;
       Linux)
-    	if [ "${IS_RESOURCE_PARTITION}" == "true" ]; then
-	   KUBELET_LOG=/tmp/kubelet.log
-           kube::common::start_kubelet
-	else
-           KUBELET_LOG=""
-    	fi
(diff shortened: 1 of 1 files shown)


hack/arktos-up-scale-out-poc.sh
```diff
@@ -640,12 +640,8 @@ if [[ "${START_MODE}" != "nokubelet" ]]; then
         KUBELET_LOG=""
         ;;
       Linux)
-    	if [ "${IS_RESOURCE_PARTITION}" == "true" ]; then
-	   KUBELET_LOG=/tmp/kubelet.log
-           kube::common::start_kubelet
-	else
-           KUBELET_LOG=""
-    	fi
+	       KUBELET_LOG=/tmp/kubelet.log
```
q131172019, line 643: If the indent is adjusted, that will be better.

q131172019 (approved):
/LGTM

hack/arktos-up-scale-out-poc.sh
Sindica: done

PR #1365, opened 2022-02-16
[Arktos-Mizar-Integration][ScaleOut]For starting mizar-operator pod on scale-out TP servers successfully

q131172019:
**What type of PR is this?**
/kind bug

This PR is to ensure the directory /etc/kubernetes will be automatically created on scale-out TP servers if the directory is not existing as default. This will ensure to start mizar-operator pod on scale-out TP servers successfully.

Fixes # N/A

This PR depends on PR #1362.

Test is done in local scale-out 1 x 1 running on Ubuntu20.04.

### DIFF
--- a/hack/lib/common-var-init.sh
+++ b/hack/lib/common-var-init.sh
@@ -236,10 +236,6 @@ nameserver ${UPSTREAM_DNS_IP}
 EOF
   fi
 fi
-#
-# For starting mizar-operator pods on scale-out TP servers successfully on Ubuntu 20.04
-#
-MIZAR_OPERATOR_HOST_PATH=${MIZAR_OPERATOR_HOST_PATH:-"/etc/kubernetes"}
 
 # --------------------------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


hack/lib/common-var-init.sh
```diff
@@ -236,6 +236,10 @@ nameserver ${UPSTREAM_DNS_IP}
 EOF
   fi
 fi
+#
+# For starting mizar-operator pods on scale-out TP servers successfully on Ubuntu 20.04
+#
+MIZAR_OPERATOR_HOST_PATH=${MIZAR_OPERATOR_HOST_PATH:-"/etc/kubernetes"}
```
Sindica, line 242: This is only used in arktos-up-scale-out-poc.sh, let's just create the path there. Not suitable for common variables.

q131172019: yes.

Sindica (approved):
/lgtm

Issue #1357, opened 2022-02-13
[Kube-up][scale-out]2x2 has 9 pods stuck in ContainerCreating


Sindica:
Start up env variables:
```
export KUBEMARK_NUM_NODES=100
export NUM_NODES=2
export SCALEOUT_TP_COUNT=2
export SCALEOUT_RP_COUNT=2
export RUN_PREFIX=ying-scaleout

export MASTER_DISK_SIZE=200GB
export MASTER_ROOT_DISK_SIZE=200GB
export KUBE_GCE_ZONE=us-central1-b 
export MASTER_SIZE=n1-highmem-32 
export NODE_SIZE=n1-highmem-16 
export NODE_DISK_SIZE=200GB 
export GOPATH=$HOME/go 
export KUBE_GCE_ENABLE_IP_ALIASES=true 
export KUBE_GCE_PRIVATE_CLUSTER=true 
export CREATE_CUSTOM_NETWORK=true KUBE_GCE_INSTANCE_PREFIX=${RUN_PREFIX} KUBE_GCE_NETWORK=${RUN_PREFIX} ENABLE_KCM_LEADER_ELECT=false ENABLE_SCHEDULER_LEADER_ELECT=false ETCD_QUOTA_BACKEND_BYTES=8589934592 SHARE_PARTITIONSERVER=false LOGROTATE_FILES_MAX_COUNT=50 LOGROTATE_MAX_SIZE=200M KUBE_ENABLE_APISERVER_INSECURE_PORT=true KUBE_ENABLE_PROMETHEUS_DEBUG=true KUBE_ENABLE_PPROF_DEBUG=true
export TEST_CLUSTER_LOG_LEVEL=--v=4 HOLLOW_KUBELET_TEST_LOG_LEVEL=--v=4 
#export SCALEOUT_CLUSTER=false # current need to set false in admin cluster, true in kubemark cluster
export SCALEOUT_CLUSTER=true # set to true for scaleout in admin cluster
export KUBE_FEATURE_GATES=ExperimentalCriticalPodAnnotation=true,QPSDoubleGCController=true

export NETWORK_PROVIDER=mizar
```

9 pods on TP1 and TP2 stuck in ContainerCreating. Endpoints were not filled in:
```
sindica2000@ying-dev1:~/go/src/sindica-arktos$ ./cluster/kubectl.sh --kubeconfig cluster/kubeconfig.tp-1 get pods -AT -o wide | grep -v Running
[...]

PR #1367, opened 2022-02-16
Add TP master into mizar droplets

Sindica:
1. Deploy daemon to TP master.
2. Add TP master node(s) into mizar droplets

Currently test passed in 1x1 local dev env. But 2x2 test is missing mizar eps for hosts. Will need mizar team to take a look.

In 1x1 scaleout, if kube-dns and coredns pod can be created successfully (Issue https://github.com/CentaurusInfra/arktos/issues/1357 tracks the pod creation issue), mizar host endpoints can be created, kube-dns&coredns pods runs w/o crashing, service endpoints filled in correctly)


q131172019 (approved):
/LGTM

PR #1371, opened 2022-02-18
Mizar VPC support for service, add TP master to mizar droplets

Sindica:
Plus some code refactoring


pkg/controller/mizar/mizar-node-controller.go
```diff
@@ -40,12 +40,16 @@ const (
...
 type MizarNodeController struct {
-	// A store of objects, populated by the shared informer passed to MizarNodeController
+	// A store of node objects, populated from Resource partition node informers
 	nodeListers map[string]corelisters.NodeLister
 	// listerSynced returns true if the store has been synced at least once.
 	// Added as a member to the struct to allow injection for testing.
 	nodeListersSynced map[string]cache.InformerSynced
 
+	// A store of node objects, populated from Tenant partition node informer
+	tpNodeListers      corelisters.NodeLister
```
yb01, line 50: nit, this is from single TP, so it should not be plural.

yb01: a design clarification: does mizar control even care about the tp node?  only nodes under RP are the worker nodes in Arktos so Mizar should only care about the RP nodes.

Sindica: > nit, this is from single TP, so it should not be plural.

When HA is in place, there will be multiple TP master. Even though most of them will be standby, all of them need to have mizar daemon running and registered with mizar.

Sindica: > a design clarification: does mizar control even care about the tp node? only nodes under RP are the worker nodes in Arktos so Mizar should only care about the RP nodes.

Apparently, TP master needs to register as a mizar endpoint. Otherwise, pod allocated to RP won't be able to talk to API server located on TP master.

yb01: for HA scenario, the multiple master node will be behind a LB, and LB is exposed to the users: in that case, the controller will go to the master node directly? consider a case where a customer pod needs to access the api server, i needs to go to the LB, not the master node directly.

pkg/controller/mizar/mizar-pod-controller.go
```diff
@@ -302,3 +301,16 @@ func processPodGrpcReturnCode(c *MizarPodController, returnCode *ReturnCode, key
...
 }
+
+func getSubnetNameFromVPC(vpc string) string {
+	return fmt.Sprintf("%s%s", vpc, subnetSuffix)
+}
+
+func getVPC(network *arktosextv1.Network) string {
+	vpc := network.Spec.VPCID
+	if len(vpc) == 0 {
+		//vpc = tenant + vpcSuffix
```
yb01, line 312: nit. unused code.

pkg/controller/mizar/mizar-service-controller.go
```diff
@@ -211,208 +212,160 @@ func (c *MizarServiceController) syncService(eventKeyWithType KeyWithEventType)
...
+	}
+
+	// Get tenant default network
+	tenantDefaultNetwork, err := c.networkLister.NetworksWithMultiTenancy(tenant).Get(defaultNetworkName)
+	if tenantDefaultNetwork.Spec.Type != mizarNetworkType {
+		return nil
+	}
+
+	if tenantDefaultNetwork.Status.Phase != arktosapisv1.NetworkReady {
+		return errors.New(fmt.Sprintf("The arktos network %s is not Ready.", tenantDefaultNetwork.Name))
```
yb01, line 229: error msgs, please use lower case for first char.

pkg/controller/mizar/mizar-node-controller.go
Sindica: Renamed. Leave HA discussion post 130

pkg/controller/mizar/mizar-pod-controller.go
Sindica: Removed.

pkg/controller/mizar/mizar-service-controller.go
Sindica: Fixed.

PR #1389, opened 2022-03-02
Remove kube-dns from local env.

Sindica:
Tested for service creation and curl

### DIFF
--- a/hack/arktos-up-scale-out-poc.sh
+++ b/hack/arktos-up-scale-out-poc.sh
@@ -623,7 +623,7 @@ if [[ "${START_MODE}" != "kubeletonly" ]]; then
   fi
   if [ "${IS_RESOURCE_PARTITION}" != "true" ]; then
      kube::common::start_kubescheduler
-     start_kubedns
+     # start_kubedns
   fi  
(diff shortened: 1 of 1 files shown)


hack/arktos-up-scale-out-poc.sh
```diff
@@ -623,7 +623,7 @@ if [[ "${START_MODE}" != "kubeletonly" ]]; then
   fi
   if [ "${IS_RESOURCE_PARTITION}" != "true" ]; then
      kube::common::start_kubescheduler
-     start_kubedns
```
yb01, line 626: suggest to remove the kube-dns as a whole since Arktos will not use it any more. the corresponding function should be removed too in this case.

i suggest this go to master directly.

yb01 (approved):
/lgtm.

hack/arktos-up-scale-out-poc.sh
Sindica: > suggest to remove the kube-dns as a whole since Arktos will not use it any more. the corresponding function should be removed too in this case.
> 
> i suggest this go to master directly.

Keep there for further testing in kube-up and arktos-up. Prefer remove kube-dns from scale out and scale up together.

PR #1391, opened 2022-03-03
Add more restrictions on VPC ip range

Sindica:
Fix ContainerCreating issue.

### DIFF
--- a/pkg/controller/mizar/mizar-arktos-network-controller.go
+++ b/pkg/controller/mizar/mizar-arktos-network-controller.go
@@ -255,8 +255,15 @@ func createVpcAndSubnet(vpc, subnet string, dynamicClient dynamic.Interface) err
 func generateVPCSpec(vpcName string) (int, *MizarVPC) {
 	// randomize ip start segment:
 	ipStart := ran.Intn(255) + 1 // IpStart range [1, 255]
+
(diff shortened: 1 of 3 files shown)


pkg/controller/mizar/mizar-arktos-network-controller.go
```diff
@@ -255,8 +255,15 @@ func createVpcAndSubnet(vpc, subnet string, dynamicClient dynamic.Interface) err
 func generateVPCSpec(vpcName string) (int, *MizarVPC) {
 	// randomize ip start segment:
 	ipStart := ran.Intn(255) + 1 // IpStart range [1, 255]
+
+	// 224.x.x.x - 239.x.x.x is reserved for IPv4 multicast
+	// https://www.iana.org/assignments/multicast-addresses/multicast-addresses.xhtml
+	if ipStart >= 224 && ipStart <= 239 {
+		ipStart %= 224
```
Hong-Chang, line 262: Minor: later you check isStart == 0. If the check is because of here, you can do %= 200 instead of 224 then the result will fall [24,39].

Hong-Chang (approved):
Approved with a small suggestion

pkg/controller/mizar/mizar-arktos-network-controller.go
Sindica: Thank you for the suggestion.

PR #1395, opened 2022-03-04
VPC start ip range fix: only generate ip start from range [11, 19]U[21,99]

Sindica:
Check in this change for Hong's further test for ContainerCreating issue


h-w-chen (approved):
lgtm for POC purpose

PR #1397, opened 2022-03-08
kube-up support vpc range

sonyafenge:
**What type of PR is this?**

> /kind feature

kube-up.sh support  VPC range for each TP

Fixes #

No additional env added when start  cluster, kube-up will automatically assign vac range for each TP between 11-99.

Tested on scenarios blow:

1. Kubeup (scale up) with mizar: successfully and vpc_range bound to kube-controller-manger
2. Kubeup (scale up with default network): successfully, no vpc_range bound to kube-controller-manger
3. Kubeup (scale up with default network) + kubemark ( scale up with default network: 100 nodes): successfully
4. Kubeup (scale out 1x1, 2x2) with mizar: successfully and TP servers have vpc_range bound to kube-controller-manger

```release-note

```

### DIFF
--- a/cluster/common.sh
+++ b/cluster/common.sh
@@ -942,6 +942,16 @@ EOF
 DISABLE_NETWORK_SERVICE_SUPPORT: $(yaml-quote ${DISABLE_NETWORK_SERVICE_SUPPORT})
 EOF
     fi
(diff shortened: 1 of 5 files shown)


q131172019 (approved):
/LGTM

Issue #1399, opened 2022-03-09
[Local dev env] Allow user to switch among different CNI plugin solutions


Sindica:
**What would you like to be added**:
If I start arktos-up using mizar, and then shut down the cluster, use default (bridge) to start cluster, I got coredns pod stuck in container creating state. Manually delete /etc/cni/net.d/10-mizarcni.conf and restart cluster will fix the issue.

We would like to allow dev to use same machine to test different CNI plugin, currently supported: bridge, flannel, and mizar.

**Why is this needed**:
Better dev experience.

q131172019:
PR1400 fixed this issue.

Issue #1348, opened 2022-02-09
[Arktos][ScaleUp]The pods of coredns-default and kube-dns are in "ContainerCreating" states when test "CNIPLUGIN=flannel ./hack/arktos-up.sh" on AWS EC2 instance


q131172019:
**What happened**:
The pods of coredns-default and kube-dns are in "ContainerCreating" states when running "CNIPLUGIN=flannel ./hack/arktos-up.sh" test flannel on simple-node cluster after this same cluster is used to test default bridge when running "./hack/arktos-up.sh" without rebooting to do cleaning on AWS EC2 instance.

On Ubuntu18.04:
```
ubuntu@ip-172-31-13-51:~/go/src/arktos$ ./cluster/kubectl.sh get pods -AT
TENANT   NAMESPACE     NAME                               HASHKEY               READY   STATUS              RESTARTS   AGE
system   kube-system   coredns-default-7b74ff4d5d-45s65   4905848824627034617   0/1     ContainerCreating   0          8m47s
system   kube-system   kube-dns-554c5866fc-dh8wr          3701911268299850684   0/3     ContainerCreating   0          8m47s
system   kube-system   virtlet-d8gjd                      5776219229357027768   3/3     Running             0          8m22s
```

On Ubuntu20.04:
```
ubuntu@ip-172-31-13-51:~/go/src/arktos$ ./cluster/kubectl.sh get pods -AT
TENANT   NAMESPACE     NAME                               HASHKEY               READY   STATUS              RESTARTS   AGE
system   kube-system   coredns-default-7b74ff4d5d-45s65   4905848824627034617   0/1     ContainerCreating   0          8m47s
system   kube-system   kube-dns-554c5866fc-dh8wr          3701911268299850684   0/3     ContainerCreating   0          8m47s
system   kube-system   virtlet-d8gjd                      5776219229357027768   3/3     Running             0          8m22s
```

**What you expected to happen**:
The pods of coredns-default and kube-dns should be in "ContainerCreating" states.

**How to reproduce it (as minimally and precisely as possible)**:
1.  Create a clean AWS EC2 t2.2xlarge instance running Ubuntu 18.04 or Ubuntu20.04.
2. Follow up https://github.com/q131172019/arktos/blob/CarlXie_singleNodeArktosCluster/docs/setup-guide/setup-dev-env.md to set up local dev environment;
[...]

PR #1400, opened 2022-03-09
[ScaleUP and ScaleOUT]Clean old CNI configuration file before install new CNI for bridge, flannel and mizar and clean interface cni0 when shutdown scaleup cluster with bridge and flannel

q131172019:
**What type of PR is this?**
/kind cleanup

This PR is to give better development experience when switch different cniplugin.
1. Clean old cni configuration files in the directories /etc/cni/net.d and /opt/cni/bin before new cni is installed and started with bridge, flannel or mizar 
2. Cleaning interface and ip link for cni0 to fix the conflict of cni0 associating with different ip addresses when switch different cniplugin between bridge and flannel to start scale-up cluster

By the way, this PR makes improvement to keep the consistency of local variables and global variables in hack/arktos-cni.rc

Tests have been done in the following 3 AWS environments:

Scale-up environment Ubuntu 18.04 (master + 3 workers)
- default (cniplugin=bridge)
- cniplugin=flannel
- cniplugin=mizar

Scale-up environment Ubuntu 20.04 (master + 3 workers)
- default (cniplugin=bridge)
- cniplugin=flannel
- cniplugin=mizar

Scale-out 2x2 environment Ubuntu 20.04 (2 TPs X 2 RPs and each RP with 1 worker)
- default (cniplugin=flannel)
- cniplugin=mizar

[Issue 1399](https://github.com/CentaurusInfra/arktos/issues/1399)
[Issue 1348](https://github.com/CentaurusInfra/arktos/issues/1348)

N/A

N/A


hack/arktos-cni.rc
```diff
@@ -67,11 +67,13 @@ install_cni_bridge() {
     echo "done with bridge cni plugin installation"
 }
 
-cleanup_cni_mizar() {
+cleanup_cni() {
     sudo rm -rf "${cni_bin_dir}"
     [ ! -d "${cni_conf_dir}" ] && sudo mkdir -p "${cni_conf_dir}"
     [ ! -d "${cni_bin_dir}" ] && sudo mkdir -p "${cni_bin_dir}"
     sudo find ${cni_conf_dir}/ -type f -not -name '10-mizarcni.conf' -delete
+    sudo find ${cni_conf_dir}/ -type f -not -name '10-flannel.conflist' -delete
```
Sindica, line 75: Can this be completed by remove all files in the directory? If so, we don't need to keep updating the file list when adding new cni plugin. Plus, delete by exact filename could be error prone.

q131172019: Because the permission of two directory /etc/cni and /etc/cni/net.d is "drwx------", which means only root can read/write/executable, all files in the directory /etc/cni/net.d can not deleted in one time whereas all files in the directory /opt/cni/bin can be removed in one time because the permission of /opt/cni/ is "drwxr-xr-x" and permission of /opt/cni/bin is "drwxrwxr-x".

Let me try to change the permission of /etc/cni into 'drwxr-xr-x' and the permission of /etc/cni/net.d into 'drwxrwxr-x', respectively first, then attempt to remove all files in the directory /etc/cni/net.d, then change the permissions of two directories back to original ones.

hack/arktos-cni.rc
```diff
@@ -230,14 +245,17 @@ EOF"
 
 CNIPLUGIN=${CNIPLUGIN:-"bridge"}
 cni_conf_dir=${CNI_CONF_DIR:-"/etc/cni/net.d"}
+cni_conf_parent_dir=${CNI_CONF_PARENT_DIR:-"/etc/cni"}
 cni_bin_dir=${CNI_BIN_DIR:-"/opt/cni/bin"}
 
 if [[ -n "${ARKTOS_NO_CNI_PREINSTALLED}" ]]; then
     echo "CNI plugin will be installed after cluster is started"
     echo "Right now the minimum common cni binary package is being installed only"
+    cleanup_cni
```
Sindica, line 254: cleanup_cni only need to be called once before the first if, right?

hack/arktos-cni.rc
```diff
@@ -230,14 +245,17 @@ EOF"
 
 CNIPLUGIN=${CNIPLUGIN:-"bridge"}
 cni_conf_dir=${CNI_CONF_DIR:-"/etc/cni/net.d"}
+cni_conf_parent_dir=${CNI_CONF_PARENT_DIR:-"/etc/cni"}
 cni_bin_dir=${CNI_BIN_DIR:-"/opt/cni/bin"}
```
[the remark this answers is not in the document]
q131172019: these vars are not global, only use in the file hack/arktos-cni.rc. The GLOBAL vars are defined in the file hack/lib/common-var-init.sh.

hack/arktos-cni.rc
q131172019: Move function cleanup_cni to be called once prior to line 251.

Sindica (approved):
/lgtm

hack/arktos-cni.rc
q131172019: New codes do work so far.

hack/arktos-cni.rc
q131172019: Works for all tests.

hack/arktos-cni.rc
q131172019: New codes do work for all tests.

hack/arktos-cni.rc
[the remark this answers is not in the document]
q131172019: The variables with lower case should be used in the file hack/arktos-cni.rc ONLY whereas the variables with upper case should be GLOBAL variables, which may be used by other scripts because thhe hack/arktos-cni.rc are called by other scripts.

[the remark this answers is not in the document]
q131172019: The new codes achieve this goal - the local variables are lower-case and the global variables are upper-case in hack/arktos-cni.rc

sonyafenge (approved):
/lgtm

PR #1398, opened 2022-03-08
Distinct VPC range, passing VPC start/end from cmd arg for scale out

Sindica:
This is a workaround for Mizar cniplugin that cannot handle VPC overlap within & across TPs

Tested this PR on top of master with mizar option - for arktos up and scale out

PR #1404, opened 2022-03-10
kube-up support vpc range (#1397)

sonyafenge:
**What type of PR is this?**

> /kind feature

This is a followup for https://github.com/CentaurusInfra/arktos/pull/1398
kube-up.sh support VPC range for each TP

Fixes #

No additional env added when start cluster, kube-up will automatically assign vac range for each TP between 11-99.

Tested on scenarios blow:

Kubeup (scale up) with mizar: successfully and vpc_range bound to kube-controller-manger
Kubeup (scale up with default network): successfully, no vpc_range bound to kube-controller-manger
Kubeup (scale out 2x2) with mizar: successfully and TP servers have vpc_range bound to kube-controller-manger

```release-note

```

### DIFF
--- a/cluster/gce/util.sh
+++ b/cluster/gce/util.sh
@@ -1695,6 +1695,16 @@ EOF
 TENANT_SERVER_KUBECONFIGS: $(yaml-quote ${TENANT_SERVER_KUBECONFIGS})
 EOF
   fi
+  if [ -n "${MIZAR_VPC_RANGE_START:-}" ]; then
+      cat >>$file <<EOF
+MIZAR_VPC_RANGE_START: $(yaml-quote ${MIZAR_VPC_RANGE_START})
+EOF
+    fi
+    if [ -n "${MIZAR_VPC_RANGE_END:-}" ]; then
+      cat >>$file <<EOF
+MIZAR_VPC_RANGE_END: $(yaml-quote ${MIZAR_VPC_RANGE_END})
+EOF
+    fi
 }
 
 function build-proxy-kube-env {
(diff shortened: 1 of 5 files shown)


cluster/gce/util.sh
```diff
@@ -3342,6 +3364,8 @@ function create-rpmaster {
   MASTER_ADVERTISE_ADDRESS="${RPSERVER_RESERVED_INTERNAL_IP[$rp_sequence]}"
   KUBERNETES_MASTER_INTERNAL_IP="${RPSERVER_RESERVED_INTERNAL_IP[$rp_sequence]}"
   KUBECONFIG="${RESOURCE_DIRECTORY}/kubeconfig${KUBEMARK_PREFIX}.rp-$rp_sequence"
+  MIZAR_VPC_RANGE_START=""
```
h-w-chen, line 3367: do these setting empty really needed?

h-w-chen (approved):
lgtm. left with a minor feedback

cluster/gce/util.sh
sonyafenge: Not sure what impact if we give RP KCM vpc_range. so leave it empty, then will not pass to RP KCM. otherwise, these will be always the last TP's value.

Sindica: > Not sure what impact if we give RP KCM vpc_range. so leave it empty, then will not pass to RP KCM. otherwise, these will be always the last TP's value.

Mizar controller won't be started in RP KCM. Argument will be taken but useless in RP.

Sindica (approved):
/lgtm

